PR #4, opened 2016-12-23
Feature/split scalar type

hdevalence:
The Scalar type is stored in memory as an array of bytes.  This allows easy
access to the bits of a scalar for Scalar x Point operations, at the cost of
forcing a pack/unpack for Scalar x Scalar.  This branch splits the Scalar type
into Scalar (packed) and UnpackedScalar (limbs).

### DIFF
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -24,9 +24,11 @@
 //!
 //! Arithmetic operations on `Scalar`s are done using 12 21-bit limbs.
 //! However, in contrast to `FieldElement`s, `Scalar`s are stored in
-//! memory as bytes, allowing easy access to the bits of the `Scalar`.
+//! memory as bytes, allowing easy access to the bits of the `Scalar`
+//! when multiplying a point by a scalar.  For efficient arithmetic
+//! between two scalars, the `UnpackedScalar` struct is stored as
+//! limbs.
 
-use std::clone::Clone;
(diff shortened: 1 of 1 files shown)


src/scalar.rs
```diff
@@ -31,6 +31,7 @@ use std::ops::{Index, IndexMut};
 
 use rand::Rng;
 
+// XXX should these be in a utility module ?
```
isislovecruft: We could move these to a `utils` module, but if we do that, should we also move `util` to something more appropriately named, like `subtle`?

src/scalar.rs
```diff
@@ -347,89 +457,11 @@ impl Scalar {
...
-        result[19] =  al[8]*bl[11] +  al[9]*bl[10] + al[10]*bl[9] + al[11]*bl[8];
-        result[20] =  al[9]*bl[11] + al[10]*bl[10] + al[11]*bl[9];
-        result[21] = al[10]*bl[11] + al[11]*bl[10];
-        result[22] = al[11]*bl[11];
-        result[23] =          0i64;
-
-        // reduce limbs and pack into output
-        Scalar::reduce_limbs(&mut result);
-        Scalar::pack_limbs(array_ref!(result, 0, 12))
+        // XXX better way to get [i64;12] from [i64;24] ?
```
isislovecruft: This is probably the cleanest way to do it.

PR #13, opened 2017-01-06
Move CTAssignable trait and implement it for more types

hdevalence:
These commits move the `CTAssignable` trait into the `util` module, and provide implementations of the type for `Scalar`s and `FieldElement`s.


isislovecruft:
I'm thinking of renaming `util` to make it more clear that they are utilities for doing constant time operations.

PR #14, opened 2017-01-08
Move constants to a separate file

hdevalence:


src/constants.rs
```diff
@@ -30,8 +30,7 @@ pub const SQRT_M1: FieldElement = FieldElement([
     -32595792,  -7943725,   9377950,   3500415,  12389472,
     -272473,   -25146209,  -2005654,    326686,  11406482, ]);
 
-/// When we define the curve according to the form y² = x³+Ax²+Bx,
-/// we get A=486662 and B=1.
+/// In Montgomery form y² = x³+Ax²+x, Curve25519 has A=486662.
```
isislovecruft: s/,/;/ technically, but it's in a docstring so I don't care.

src/constants.rs
```diff
@@ -35,6 +35,9 @@ pub const A: FieldElement       = FieldElement([
     486662, 0, 0, 0, 0, 0, 0, 0, 0, 0, ]);
 
 /// SQRT_MINUS_A is sqrt(-486662)
+// XXX I think that this was used in Adam's code for his elligator
+// implementation, but that should maybe be using sqrt(-486664)
+// instead...?  - hdevalence
```
isislovecruft: This is the part of the forward mapping where one checks that `(s² - 2)(s² + 2) ≠ 0` where `s ∈ 𝐆𝐅(q)`? I think I'm not seeing where the `-A - 2` comes in.

isislovecruft: @hdevalence Maybe we should try asking Adam how he derived this? And maybe in a more polite manner this time than ambushing him on the metro.

isislovecruft:
LGTM

PR #18, opened 2017-01-23
Link needs http://www.

burdges:
Link is broken here https://docs.rs/curve25519-dalek/0.1.3/curve25519_dalek/curve/index.html


isislovecruft:
@burdges Thanks!

PR #20, opened 2017-02-14
Add "cryptography" and "no-std" categories

tarcieri:
These will get picked up the next time you publish the crate (provided you have a recent version of cargo):

https://crates.io/categories/cryptography


isislovecruft:
Cool! Thanks, @tarcieri!

PR #21, opened 2017-02-20
Derive Eq, PartialEq for CompressedEdwardsY

hdevalence:
Cut out boilerplate using a derive directive.


isislovecruft:
It went into `develop`

PR #22, opened 2017-02-20
Add a conditional negation trait with generic implementation

hdevalence:
The name `CTNegateable` is kind of awkward, but I think gets the point across, and it's in any case only used when importing the trait.

### DIFF
--- a/src/curve.rs
+++ b/src/curve.rs
@@ -87,6 +87,7 @@ use field::FieldElement;
 use scalar::Scalar;
 use util::bytes_equal_ct;
 use util::CTAssignable;
+use util::CTNegateable;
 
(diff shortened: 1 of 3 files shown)


isislovecruft:
"for better ergonomics" :smile:

isislovecruft:
Okay, it all looks good, except that I _think_ "negateable" is spelled _negatable_: https://en.wiktionary.org/wiki/negatable

hdevalence:
Good catch -- thanks for `Referer`ing me to the correct spelling :+1:

isislovecruft:
No problem, I wouldn't want to `Closure` this issue when we could just free the variable spelling

Issue #9, opened 2016-12-29
Implement the remaining functionality needed for (V)XEdDSA


isislovecruft:
Specifically, the functionality here which we don't already have: https://github.com/WhisperSystems/curve25519-java/tree/master/android/jni/ed25519/additions

I'd prefer to do this all at once (as in, all in one minor version), show it to Trevor, and then (if he approves) merge it in, rather than implement incrementally over time.

isislovecruft:
What we have and don't have:

* `compare.c`
   - ✓ `crypto_verify_32_ref()` → `curve25519_dalek::util::arrays_equal_ct()`

* `curve_sigs.c`
   - ✓ `curve25519_sign()` → `ed25519_dalek::SecretKey::sign()`
   - ✓ `curve25519_verify()` → `ed25519_dalek::PublicKey::verify()`
   - However, both curve25519-java functions operate on curve25519 keys, so **we still need Mongomery x-coordinate → Edwards y-coordinate conversion.**

* `elligator.c`
   - `legendre_is_nonsquare()`
   - `elligator()`
   - `hash_to_point()`
   - `calculate_Bv()`
   - `calculate_Bv_and_V()`

* `fe_isequal.c`
   - ✓* `fe_isequal(f, g)` → `impl PartialEq for FieldElement`… *but* our implementation converts both `f` and `g` to a "canonical"/unique compressed representation first, and then compares byte-for-byte. I personally feel that a function which, in constant time, tests for equality of field elements (which may have different representations due to internal shifting in radix/carries) is *less* strict and *differently* expensive than one which does a representational conversion and tests byte-for-byte equality. However, I'm not sure how to best represent these two concepts clearly and usably in the API. Suggestions/ideas welcome. (* What I mean is "meh, ✓, this is kind of done but should have some further evaluation/discussion")

* `fe_mont_rhs.c`
   - `fe_mont_rhs()`

* `fe_montx_to_edy.c`
   - `fe_montx_to_edy()`

* `fe_sqrt.c`
   - `fe_sqrt()`

* `ge_isneutral.c`
   - `ge_isneutral()` **Note:** we do *not* want to test that a group element is equal to the identity in the way this function is doing so. This function is doing `(fe_isequal(p->X, zero) & fe_isequal(p->Y, p->Z))`… the latter part of that test allows `Y` and `Z` coordinates to be different to those of the identity element.

* `ge_is_small_order.c`
   - `ge_is_small_order()`

[...]

PR #23, opened 2017-02-21
Some VXEdDSA functionality

isislovecruft:
These commits fix part of #9. They add:

* determining if two scalars are equal in constant time
* negating a scalar
* determining if a point is of small order
* determining if an element is the identity


src/curve.rs
```diff
@@ -240,6 +243,16 @@ impl Identity for ExtendedPoint {
                        Z: FieldElement::one(),
                        T: FieldElement::zero() }
     }
+
+    fn is_identity(&self) -> bool {
+        if self.X == FieldElement::zero() &&
+           self.Y == FieldElement::one()  &&
+           self.Z == FieldElement::one()  &&
+           self.T == FieldElement::zero() {
+               return true;
```
hdevalence: This code rejects, e.g., (0:2:2:0), which is the identity point.  IMO a clean way to implement this instead is to compress the point to bytes and compare with the bytes of the identity point.

src/curve.rs
```diff
@@ -248,6 +261,15 @@ impl Identity for ProjectivePoint {
...
                          Z: FieldElement::one() }
     }
+
+    fn is_identity(&self) -> bool {
+        if self.X == FieldElement::zero() &&
+           self.Y == FieldElement::one()  &&
+           self.Z == FieldElement::one() {
+               return true;
+            }
+        false
```
hdevalence: same comment as above

src/curve.rs
```diff
@@ -257,6 +279,16 @@ impl Identity for CachedPoint {
                      Z:         FieldElement::one(),
                      T2d:       FieldElement::zero() }
     }
+
+    fn is_identity(&self) -> bool {
```
hdevalence: same comment as above, but I'm not sure why anyone would ever need to do this on a `CachedPoint`

src/curve.rs
```diff
@@ -231,6 +231,9 @@ pub trait Identity {
     /// Returns the identity element of the curve.
     /// Can be used as a constructor.
     fn identity() -> Self;
+
```
hdevalence: I'm not sure whether this should be in the same trait, because I don't see why everything that implements identity() in order to have a constructor now has to have equality testing...

src/scalar.rs
```diff
@@ -48,6 +50,25 @@ use util::CTAssignable;
 #[derive(Copy,Clone)]
 pub struct Scalar(pub [u8; 32]);
 
+// XXX make CTEq traits
+impl Eq for Scalar{}
+impl PartialEq for Scalar {
+    /// Test equality between two `Scalar`s in constant time.
```
hdevalence: I don't think this should be advertised as being "in constant time", since it returns a bool...

isislovecruft: Yeah, that's why there's the XXX there, I'll just fix that now.

src/scalar.rs
```diff
@@ -62,6 +85,15 @@ impl IndexMut<usize> for Scalar {
     }
 }
 
+impl Neg for Scalar {
```
hdevalence: Is there a good reason to do this instead of just negating the limbs?

src/curve.rs
```diff
@@ -654,6 +654,24 @@ impl ExtendedPoint {
...
+    ///
+    /// True if it is of small order; false otherwise.
+    pub fn is_small_order(&self) -> bool {
+        let p8: ExtendedPoint = self.mult_by_pow_2(3);
+
+        if p8.compress() == ExtendedPoint::identity().compress() {
+            return true;
+        }
+        false
+    }
```
hdevalence: LGTM, maybe the `is_identity()` method should use `compress()` and this should call that?

src/curve.rs
hdevalence: Is there a good reason not to just have `is_identity()` part of `impl ExtendedPoint`?

src/curve.rs
isislovecruft: Ack, good plan. I'll fix that.

isislovecruft: Fixed in 911cce093

src/curve.rs
isislovecruft: Also fixed in 911cce093

src/curve.rs
isislovecruft: Also fixed (i.e., removed) in 911cce093

src/curve.rs
isislovecruft: I made a new `IsIdentity` trait in 911cce093, and then made a generic implementation which will exist for all point representations which have `Identity` and `CTEq` traits implemented.

src/scalar.rs
isislovecruft: That's probably faster. Poking at the limbs sounds kind of messy/ugly, but I can't really think of any other reason.

src/curve.rs
isislovecruft: Yep, done in 911cce093

isislovecruft:
I've added `CTEq` for `Scalar`, `FieldElement`, and `ExtendedPoint`.

isislovecruft:
See also #24 which adds more sign/negation functionality for `Scalar`.

src/scalar.rs
hdevalence: We can just leave this as-is and change it later

src/curve.rs
isislovecruft: Oh, actually, not in that commit, but I've fixed it now and autosquashed

isislovecruft:
Re-squashed and now 03589df uses `is_identity()`

PR #24, opened 2017-02-21
Add methods for testing scalar sign and getting absolute value

isislovecruft:


src/scalar.rs
```diff
@@ -496,6 +520,14 @@ impl UnpackedScalar {
         UnpackedScalar(*array_ref!(limbs,0,12))
     }
 
+    /// Determine if this `UnpackedScalar` is negative in constant time.
```
hdevalence: What definition of negative does this use?

src/scalar.rs
```diff
@@ -496,6 +520,14 @@ impl UnpackedScalar {
...
+    ///
+    /// # Returns
+    ///
+    /// If negative, return `-1i32`.  Otherwise, return `0i32`.
+    pub fn is_negative(&self) -> i32 {
```
hdevalence: Should this use `u8`s like the rest of the CT code?

src/scalar.rs
```diff
@@ -496,6 +520,14 @@ impl UnpackedScalar {
...
+        (*self[0] as i32) >> 255  // 255 = (sizeof(i32) * CHAR_BIT) - 1
```
hdevalence: I think this should be 31, not 255

isislovecruft:
Closed because there's no way any of this would be mergeable at this point.

PR #30, opened 2017-02-27
Some cleanup / API tweaks

hdevalence:


src/curve.rs
```diff
@@ -140,7 +137,7 @@ impl CompressedEdwardsY {
         if is_nonzero_square != 1u8 { return None; }
 
         // Flip the sign of X if it's not correct
-        let compressed_sign_bit = self[31] >> 7;
+        let compressed_sign_bit = self.as_bytes()[31] >> 7;
```
isislovecruft: Did you bench this? I'm wondering this is slower with the added function call.

isislovecruft:
There's a `#[deny(missing_debug)]` thing (re: 950519b97b) which I could turn on.

PR #31, opened 2017-02-27
Hash to scalar

hdevalence:
Add a generic hash-to-scalar function, parameterized by a `Digest` instance with 512 bits of output.


isislovecruft:
LGTM

PR #32, opened 2017-02-28
Montgomery-U format

hdevalence:


src/field.rs
```diff
@@ -798,6 +798,8 @@ impl FieldElement {
     /// Given a nonzero field element, compute its inverse.
     /// The inverse is computed as self^(p-2), since
     /// x^(p-2)x = x^(p-1) = 1 (mod p).
+    ///
+    /// XXX should we add a debug_assert that self is nonzero?
```
isislovecruft: yes, debug asserts are friendly to other devs (i.e. all our users)

hdevalence: Only concern is that it may be useful to have 0.inverse() return zero, for similar reasons as it's useful to have the invsqrt function return 0 on 0 in Decaf...

src/curve.rs
```diff
@@ -189,6 +189,18 @@ impl CompressedMontgomeryU {
         // y = (u - 1) / (u + 1)
 
         let u = FieldElement::from_bytes(&self.0);
+
+        // If u = -1, then v^2 = u*(u^2+486662*u+1) = 486660.
+        // But 486660 is nonsquare mod p, so this is not a curve point.
+        //
+        // XXX what does Signal do here?
```
isislovecruft: Signal uses Montgomery `u` to derive Edwards `y` (by `y = (u-1)/(u+1)`, `u` to derive `v` (by `v = sqrt(u(u² + Au + 1))`), `(u, v)` to derive Edwards `x` (by `x = (u/v) * sqrt(-(A+2))`), It also expects to have `ge_montx_to_p3()` be passed a mask indicating the sign, and it conditionally negates such that the resulting Edwards `x` is always positive.

Maybe we should just copy that behaviour? If for no other reason than it has one less inversion.

PR #33, opened 2017-03-02
Refactor Scalar::hash_from_bytes to allow streaming input to the hash.

hdevalence:


burdges:
I so hope the generic array crate gets replaced by something cleaner soon. :(

isislovecruft:
This will be available in 0.7.0

Issue #8, opened 2016-12-24
Implement Elligator 2 (hashing to a curve point)


hdevalence:
Stub issue for tracking Elligator 2 implementation.

hdevalence:
Closing this for now, we can reopen it later if need be.

isislovecruft:
Do we want to potentially add an invertible Elligator2 variant? It's rare that this is needed in practice, but there are systems like https://eprint.iacr.org/2019/1416 (cf. §4.1) which require not only _hash_-to-group but also _encode_-to-group (and the corresponding _decode_-to-group).

hdevalence:
Hmm, I think that that's pretty different functionality from what this issue was about, so I'd prefer to discuss it separately from this issue.

hdevalence:
For reference, it's possible to implement encode-to-group using the existing API without any changes to the library.  I wrote an example of how to do this last year: https://github.com/hdevalence/ristretto255-data-encoding/blob/master/src/main.rs

isislovecruft:
Thanks! I'll open a separate issue for discussion.

PR #34, opened 2017-03-04
Montgomery-U format, revised

isislovecruft:
A few commits on top of hdevalence's PR #32 to cleanup the decompression and match Signal's behaviour.

### DIFF
--- a/src/curve.rs
+++ b/src/curve.rs
@@ -121,7 +121,7 @@ impl Index<usize> for CompressedEdwardsY {
 
 impl CompressedEdwardsY {
     /// View this `CompressedEdwardsY` as an array of bytes.
-    pub fn to_bytes(&self) -> [u8;32] {
+    pub fn to_bytes(&self) -> [u8; 32] {
         self.0
     }
 
@@ -165,7 +165,7 @@ pub struct CompressedMontgomeryU(pub [u8; 32]);
 
 impl CompressedMontgomeryU {
     /// View this `CompressedMontgomeryU` as an array of bytes.
-    pub fn to_bytes(&self) -> [u8;32] {
+    pub fn to_bytes(&self) -> [u8; 32] {
         self.0
     }
 
--- a/src/curve.rs
+++ b/src/curve.rs
@@ -171,29 +171,30 @@ impl CompressedMontgomeryU {
 
     /// Attempt to decompress to an `ExtendedPoint`.
     ///
-    /// Note that since there are two curve points with the same
+    /// # Note
+    ///
+    /// Since there are two curve points with the same
     /// `u`-coordinate, the `u`-coordinate does not fully specify a
-    /// point.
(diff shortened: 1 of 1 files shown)


isislovecruft:
FWIW, I need this for #8, otherwise our `hash_to_point()` will be different to Signal's and I would need to make a breaking change later.

src/curve.rs
```diff
@@ -165,49 +165,120 @@ pub struct CompressedMontgomeryU(pub [u8; 32]);
...
-    /// and rewrite this note
+    /// # Warning
     ///
-    /// XXX check for div by zero: when is u = -1 ?
-    /// XXX exceptional points for the birational map 
+    /// This function is *not* constant time.
+    ///
+    /// # Return
+    ///
+    /// An `Option<ExtendedPoint>`, which will be `None` if either condition holds:
```
hdevalence: I think it's more meaningful to say that the function returns `None` if the input is not a valid Montgomery `u`-coordinate.

src/curve.rs
```diff
@@ -165,49 +165,120 @@ pub struct CompressedMontgomeryU(pub [u8; 32]);
...
-        //
         // Note: currently, without this check, u = -1 will accidentally
         // decode to a valid (but incorrect) point, since 0.invert() = 0.
         if u == FieldElement::minus_one() {
             return None;
         }
 
-        let u_plus_1_inv = (&u + &FieldElement::one()).invert();
-        let y = &(&u - &FieldElement::one()) * &u_plus_1_inv;
+        let (ok, v): (u8, FieldElement) = CompressedMontgomeryU::to_montgomery_v(&u); // v = sqrt(u(u²+Au+1))
```
hdevalence: Computation of `v` isn't necessary and just costs an extra inversion for a value that's never used.

src/curve.rs
```diff
@@ -165,49 +165,120 @@ pub struct CompressedMontgomeryU(pub [u8; 32]);
...
-        // is it possible to do one?
         CompressedEdwardsY(y.to_bytes()).decompress()
     }
+
+    /// Given a Montgomery `u` coordinate, compute an Edwards `y` via
+    /// `y = (u-1)/(u+1)`.
+    ///
+    /// # Note
+    ///
+    /// Since `u = (1+y)/(1-y)` and `v = √(u(u²+Au+1))`, we can see that
```
hdevalence: Where does the `v` appear?

isislovecruft: Demoted to inline code comment in 2c83d48

src/curve.rs
isislovecruft: Yep, that was a mistake since it was a remnant of doing it a different way where we can optionally recover the sign if we had it earlier, but it's not necessary to do it that way everytime (which is why I removed the other coordinate recovery calls, but forgot to remove this one). It's removed in 915f41c.

hdevalence: Right, that makes sense, sorry.

PR #35, opened 2017-03-04
Rename ExtendedPoint.compress() to ExtendedPoint.compress_edwards()

isislovecruft:
Or maybe perhaps we should name the functions `compress_to_edwards()` and `compress_to_montgomery()`? Not sure. Anyway, we shouldn't have a `compress()` and a compressed_montgomery()`, the asymmetry is killing me.


isislovecruft:
Merged to develop

PR #36, opened 2017-03-08
Reorganize test code

hdevalence:


src/curve.rs
```diff
@@ -1119,29 +1119,27 @@ mod test {
...
-        let bp = BASE_CMPRSSD.decompress().unwrap();
-        assert_eq!(m_bp.X, -(&bp.X));
-        assert_eq!(m_bp.Y,  bp.Y);
-        assert_eq!(m_bp.Z,  bp.Z);
-        assert_eq!(m_bp.T, -(&bp.T));
+        // Manually set the high bit of the last byte to flip the sign
+        let mut minus_basepoint_bytes = BASE_CMPRSSD.as_bytes().clone();
+        minus_basepoint_bytes[31] |= 1 << 7;
+        let minus_basepoint = CompressedEdwardsY(minus_basepoint_bytes)
+                              .decompress().unwrap();
```
isislovecruft: IIRC the Rust style guide says lines over 80 chars and up to 100 are fine, FWIW.

Also, I personally don't care if a line goes up to 100, because this isn't 1992 and my screen is now wider than it is tall.

hdevalence:
Yes, I think those two tests are covered by other tests.

PR #37, opened 2017-03-08
no_std import cleanups

tarcieri:
- Gate no_std on the "std" feature
- Import core when std is present
- Import collections when std is absent
- Add a placeholder gated "use" directive for Box in curve.rs

### DIFF
--- a/src/curve.rs
+++ b/src/curve.rs
@@ -91,6 +91,9 @@ use subtle::CTAssignable;
 use subtle::CTEq;
 use subtle::CTNegatable;
 
+#[cfg(not(feature = "std"))]
+use collections::boxed::Box;
+
 // ------------------------------------------------------------------------
 // Compressed points
 // ------------------------------------------------------------------------
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,7 +9,8 @@
 //  - Isis Agora Lovecruft <isis@patternsinthevoid.net>
 //  - Henry de Valence <hdevalence@hdevalence.ca>
 
-#![no_std]
+#![cfg_attr(not(feature = "std"), no_std)]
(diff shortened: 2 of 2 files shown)


src/curve.rs
```diff
@@ -91,6 +91,9 @@ use subtle::CTAssignable;
 use subtle::CTEq;
 use subtle::CTNegatable;
 
+#[cfg(not(feature = "std"))]
+use collections::boxed::Box;
```
tarcieri: Put this in here as an example of how to use `Box` in a `no_std`-friendly way

src/lib.rs
```diff
@@ -9,7 +9,8 @@
 //  - Isis Agora Lovecruft <isis@patternsinthevoid.net>
 //  - Henry de Valence <hdevalence@hdevalence.ca>
 
-#![no_std]
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(not(feature = "std"), feature(collections))]
```
tarcieri: Needed to use `Box`, `Vec`, and friends with `no_std`

src/lib.rs
```diff
@@ -33,18 +34,20 @@
...
-extern crate std;
-
-#[cfg(test)]
 extern crate test;
 
 #[macro_use]
 extern crate arrayref;
 
 #[cfg(feature = "std")]
+extern crate core;
```
tarcieri: Absent `no_std` this needs to be included for `core` to work

tarcieri:
I should probably give this a whirl and make sure things still work even in environments where allocators aren't available. I think I should be good (using `--gc-sections` during linking) so long as I avoid features that use allocators.

Might be worth properly feature gating.

PR #38, opened 2017-03-13
Added conditional compilation for 'test' feature

tqoitc:
The crate won't currently compile on stable unless this is added.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,7 +11,7 @@
 
 #![no_std]
 #![allow(unused_features)]
-#![feature(test)]
+#![cfg_attr(test, feature(test))]
 #![deny(missing_docs)] // refuse to compile if documentation is missing
 
 //! # curve25519-dalek


hdevalence:
Doesn't this change disable tests entirely on stable?  I think we really don't want to do that, since the tests are sort of an integral part of the library.  

As I recall, the reason for using the `test` feature was to allow using `cargo bench`.  The changes merged into `develop` by #36 split the benchmarks into their own `bench` module, so I think it would be better to just disable benchmarks on stable rather than all tests.

tqoitc:
Yes, it would disable tests on stable but as of right now it doesn't compile at all on stable.

isislovecruft:
Hi @tqoitc! Thanks for pointing this out. It makes sense now that you've mentioned it, but I didn't fully realise that `#[feature(test)]` was only on nightly. (I compile my Rust compiler, and had never actually thought about what "channel" that meant I was on.)

In #42, I've made it so that both compilation and tests should work for you on either stable or beta. (If you want to run the benchmarks, you'll still have to run nightly, however.) Also, as of #40, all of the tests are run on all rustc release channels, with benchmarks expected to pass only on nightly.

Thanks again for bringing this to our attention!

PR #48, opened 2017-04-25
Clippy fixes

hdevalence:
Various fixes to deal with issues flagged by rust-clippy.

### DIFF
--- a/src/curve.rs
+++ b/src/curve.rs
@@ -116,7 +116,7 @@ impl Debug for CompressedEdwardsY {
 
 impl CompressedEdwardsY {
     /// View this `CompressedEdwardsY` as an array of bytes.
-    pub fn as_bytes<'a>(&'a self) -> &'a [u8; 32] {
+    pub fn as_bytes(&self) -> &[u8; 32] {
         &self.0
     }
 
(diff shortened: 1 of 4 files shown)


src/curve.rs
```diff
@@ -116,7 +116,7 @@ impl Debug for CompressedEdwardsY {
 
 impl CompressedEdwardsY {
     /// View this `CompressedEdwardsY` as an array of bytes.
-    pub fn as_bytes<'a>(&'a self) -> &'a [u8; 32] {
+    pub fn as_bytes(&self) -> &[u8; 32] {
```
isislovecruft: Wow, Rust lifetime elision is getting fancier and fancier.

src/field.rs
```diff
@@ -455,9 +455,8 @@ impl FieldElement {
         FieldElement(limbs)
     }
     #[cfg(not(feature="radix_51"))]
-    fn reduce(input: &[i64;10]) -> FieldElement { //FeCombine
+    fn reduce(mut h: [i64; 10]) -> FieldElement { //FeCombine
```
isislovecruft: Clippy recommended this change?

isislovecruft:
These are going to be in 0.7.0

PR #55, opened 2017-05-19
Add cargo-fuzz

Manishearth:


Manishearth:
This can be run via `cargo fuzz run decaf`. Sometimes you may need to delete the lockfile under `fuzz`

So far it hasn't found anything for me, but it _immediately_ catches the overflow bug in #54 if run on 121818800397ef9c619470d696183867aad8f48c

Issue #57, opened 2017-05-22
Operators only implemented on references


UnlawfulMonad:
I've noticed that operators on `ExtendedPoint` and the like are only implemented on `&'a ExtendedPoint`. Is there any reason for a PR that also implemented them for the concrete types wouldn't be accepted? The types are already copy and as far as I know the compile is better at reasoning about the code when there isn't explicit memory (i.e. a reference) involved.

burdges:
If I [understand](https://github.com/rust-lang/rfcs/issues/1936), the benchmarks suffer if you pass these structures by value, but maybe one can tweak that with optimizations modes, etc.  I'd also think making the operators `#[inline(always)]` wrappers around call by reference functions should fix this, but maybe not.

hdevalence:
When we implemented the arithmetic ops for `T` and not for `&T`, we had a lot of nested redundant copies: first, copying the LHS and RHS points into the function (each point is 160 bytes), plus copying the LHS and RHS of every field element operation inside (each field element is 40 bytes).  It did make a difference in cutting out needless copies.  However it's obviously worse ergonomically, as I'm sure you've noticed.

I didn't think that it would be possible to work around this, but it seems like @burdges' suggestion [might work](https://godbolt.org/g/qBbfCE), at least on that toy example.

So, to answer your original question, @UnlawfulMonad, if there were implementations of the arithmetic ops that didn't require borrows but were just as fast, we would be happy to take that PR.

isislovecruft:
Hi @UnlawfulMonad!

As @hdevalence said, we'd be stoked to take a patch for this, provided it doesn't slow anything down by introducing needless copies everywhere. If you want to implement this, my suggestion for the steps would be:

1. Implement the operators for non-borrow types.
[...]

PR #58, opened 2017-05-27
Replace call by reference by call by value

UnlawfulMonad:
Tracking issue: #57

Implements `Add`, `Sub` and `Mul` on value variants of `FieldElement`, `ExtendedElement` and `Scalar`.
At the same time gets rid of "Eye of Sauron" (arithmetic that looks like `&(&x + &y) * &z.square()`) expressions. Difference in performance can be compared to master and seems to be in the low _ns_ for operatations on `FieldElement`s however is higher on more complex operations.

This PR isn't ready to be merged until the performance issues are resolved.

### DIFF
--- a/src/curve.rs
+++ b/src/curve.rs
@@ -867,6 +867,9 @@ impl<'b> MulAssign<&'b Scalar> for ExtendedPoint {
     }
 }
 
+mul_impl!(Scalar, ExtendedPoint, ExtendedPoint);
+mul_impl!(ExtendedPoint, Scalar, ExtendedPoint);
+
 impl<'a, 'b> Mul<&'b Scalar> for &'a ExtendedPoint {
     type Output = ExtendedPoint;
     /// Scalar multiplication: compute `scalar * self`.
@@ -907,6 +910,8 @@ impl<'a, 'b> Mul<&'b ExtendedPoint> for &'a Scalar {
 /// Precomputation
 #[derive(Clone)]
 pub struct EdwardsBasepointTable(pub [[AffineNielsPoint; 8]; 32]);
+
+mul_impl!(Scalar, EdwardsBasepointTable, ExtendedPoint);
 
 impl<'a, 'b> Mul<&'b Scalar> for &'a EdwardsBasepointTable {
     type Output = ExtendedPoint;
@@ -956,6 +961,8 @@ impl<'a, 'b> Mul<&'b Scalar> for &'a EdwardsBasepointTable {
         h
     }
 }
+
+mul_impl!(EdwardsBasepointTable, Scalar, ExtendedPoint);
 
 impl<'a, 'b> Mul<&'a EdwardsBasepointTable> for &'b Scalar {
     type Output = ExtendedPoint;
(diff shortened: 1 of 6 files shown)


src/curve.rs
```diff
@@ -716,94 +716,108 @@ impl ExtendedPoint {
...
-            T: &ZZ2 - &TT2d
+            X: PP - MM,
+            Y: PP + MM,
+            Z: ZZ2 + TT2d,
+            T: ZZ2 - TT2d
         }
     }
 }
 
+impl Add<ProjectiveNielsPoint> for ExtendedPoint {
```
hdevalence: Should this impl be using one of the macros you defined?

src/field.rs
```diff
@@ -109,27 +109,36 @@ impl Debug for FieldElement {
...
         for i in 0..self.0.len() {
             self[i] += _rhs[i];
         }
     }
 }
 
+add_impl!(FieldElement, FieldElement, FieldElement);
+sub_impl!(FieldElement, FieldElement, FieldElement);
+mul_impl!(FieldElement, FieldElement, FieldElement);
+neg_impl!(FieldElement, FieldElement);
```
hdevalence: This syntax is a bit confusing -- maybe it would be better if the macro invocation looked like
```
mul_impl!( LHS = FieldElement, RHS = FieldElement, Output = FieldElement )
neg_impl!( Self = FieldElement, Output = FieldElement )
```
or something similar, so that the roles of the parameters were clear just by looking?

src/macros.rs
```diff
@@ -0,0 +1,53 @@
+// Various macros for implmenting repetative pass-by-value semantics
+
+#[macro_export]
+macro_rules! add_impl {
```
hdevalence: Maybe we should extend these macros to also generate implementations of "mixed" operators like `&FieldElement * FieldElement` and `FieldElement * &FieldElement`, so that all cases of borrow / non-borrow are covered?

UnlawfulMonad: Yeah I was considering adding those afterwards if this worked out.

UnlawfulMonad:
@hdevalence On the issue (#57) I posted my benchmarks. That seems in the range of the difference I saw. Keep in mind that's including removing most of the Eye of Saurons from the library code too. I wouldn't be surprised if the performance penalty is simply because I messed up and forgot to add inline tags everywhere I was supposed to.

src/macros.rs
UnlawfulMonad: Yeah, I was considering adding those later if this PR goes through.

src/curve.rs
```diff
@@ -887,7 +889,7 @@ impl<'a> Neg for &'a AffineNielsPoint {
 // ------------------------------------------------------------------------
 
 impl<'b> MulAssign<&'b Scalar> for ExtendedPoint {
-    #[inline]
+    #[inline(always)]
     fn mul_assign(&mut self, scalar: &'b Scalar) {
         let result = (self as &ExtendedPoint) * scalar;
```
burdges: Wouldn't `&*self` be simpler here?

UnlawfulMonad:
Closing due to many conflicts and #101 fixing this for the most part.

PR #61, opened 2017-06-27
Only pull in subtle's std feature if "std" is enabled

tonychain:
Otherwise subtle implicitly pulls in std because it's a default feature.

Tested and confirmed working in our `no_std` environment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ version = "0.6"
 
 [dependencies.subtle]
 version = "^0.1"
+default-features = false
 
 [dependencies.generic-array]
 # same version that digest depends on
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -48,7 +49,7 @@ version = "0.6"
 [features]
 nightly = ["radix_51"]
 default = ["std"]
-std = ["rand"]
+std = ["rand", "subtle/std"]
```
tonychain: I don't think you're actually using any of the std-dependent features of `subtle`? I enabled them just in case you want to, though.

isislovecruft:
Merged for 0.9.2

Thanks @tonychain!

PR #64, opened 2017-07-30
Better FieldElement impl handling

hdevalence:
Split the 32 and 64-bit implementations across different files, using type aliases and re-exports to hide the implementation from consumers.


hdevalence:
Yeah, I'm not sure why it's still there -- it's included because I branched this off of my `develop` branch, but I'd expect it to be gone now that the other one is merged.

PR #65, opened 2017-07-31
Add constant time k-fold scalar mult

hdevalence:
This provides a constant-time version of the `k_fold_scalar_mult` function already in the `vartime` module.

Question: are we happy with the name? Does the `vartime::` submodule effectively separate two functions that do the same thing, but one is constant-time?


hdevalence:
Renamed `k_fold_scalar_mult` to `multiscalar_mult`.

hdevalence:
Looks like `720da34` breaks things, because chaining two iterators of known size results in an iterator of unknown size: https://github.com/rust-lang/rust/issues/34433

Otherwise I think this branch is good to be merged; I can revert that commit explicitly and push an update if that's a better workflow.

PR #66, opened 2017-07-31
Bump subtle dependency to ^0.2 and use new API.

isislovecruft:
This should bring some decent (~30%) speedups wherever `slices_equal` is used, which for us is all over the place, but usually we are comparing relatively small slices (~32 bytes) so the speedup is not as drastic as it is for larger slice comparisons.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ version = "0.3"
 version = "0.6"
 
 [dependencies.subtle]
-version = "^0.1"
+version = "^0.2"
 default-features = false
 
 [dependencies.generic-array]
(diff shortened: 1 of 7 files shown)


isislovecruft:
Yeah, I'll change it to either `ConstantTimeEquality` or `ConstantTimeEqual` or `ConstantTimeEq` but I'll have to cut a subtle release for that and I'm not doing it just for one name change.

isislovecruft:
This no longer applies, at all, in any salvageable way due to merging #64. It's faster to redo the patch because it's just a few sed invocations.

isislovecruft:
The new branch is `feature/use-subtle-0.2_v2`

PR #71, opened 2017-08-22
Optimize scalar inversion by implementing square()

khyperia:
This shows an 11% speedup for invert(). From `cargo bench`:

* Old `invert`: `61,945 ns/iter (+/- 19,909)`
* New `invert`: `55,051 ns/iter (+/- 6,177)`
* scalar_unpacked_multiply_add: `181 ns/iter (+/- 7)`
* square: `159 ns/iter (+/- 21)`

This addresses bullet point `b` of https://github.com/isislovecruft/curve25519-dalek/issues/68 (and possibly `c` too?)

Questions I don't know the answer to: should `square` be part of the public API? Should I add a wrapper to `Scalar` instead of only having it on `UnpackedScalar`?

This is my first PR to this repo, I would welcome advice on how to improve 😊

### DIFF
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -568,12 +568,46 @@ impl UnpackedScalar {
         let mut y = UnpackedScalar::one();
         // Run through bits of l-2 from highest to least
         for bit in constants::l_minus_2.bits().iter().rev() {
-            y = UnpackedScalar::multiply_add(&y, &y, &UnpackedScalar::zero());
+            y = y.square();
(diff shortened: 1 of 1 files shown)


hdevalence:
This looks good to me.

Re: the API question, I'm not super opinionated, but I think it makes sense for people who care about scalar arithmetic performance (the rare case) to just use the UnpackedScalar type directly.

hdevalence:
Merged into the `develop` branch instead of `master`.

Issue #47, opened 2017-03-21
X25519 functions


mcginty:
It would be great to have this library (or possibly a `x25519-dalek` similar to `ed25519-dalek`, to keep this one low-level focused) provide the higher-level DH functions `curve25519()` and `curve25519_base()` functions found in donna, so people like myself don't find unique and beautiful ways to royally fudge up the scalar multiplication on their own.

burdges:
If I [understand](https://docs.rs/curve25519-dalek/0.6.0/curve25519_dalek/curve/struct.CompressedMontgomeryU.html), there is no Montgomery ladder implementation here yet.  You can implement x25519 with this library, but the Edwards multiplication here will take at least twice as long, and the conversion to/from Montgomery form will take time too, and not everything is constant time, so they do not plan to do it that way.

isislovecruft:
Hi @mcginty!

Sorry this took so long to get to, but I hacked on it a bit last weekend and wrote a montgomery ladder in my `feature/montgomery-arithmetic` branch.

I've also made [x25519-dalek](https://github.com/isislovecruft/x25519-dalek) which uses the branch (so not published as a crate yet). x25519-dalek passes the testvectors in RFC7748. However, I haven't done anything to optimise it yet; in particular, I estimate we could save about 30-40% if the internal steps to the Montgomery ladder, `differential_add()` and `differential_double()` are combined into one another.

All of this is a clean room implementation from [RFC7748](https://tools.ietf.org/html/rfc7748), Costello and Smith's ["Montgomery curves and their arithmetic"](http://eprint.iacr.org/2017/212), and Montgomery's original ["Speeding the Pollard and elliptic curve methods of factorization"](http://www.ams.org/mcom/1987-48-177/S0025-5718-1987-0866113-7/).

isislovecruft:
The Montgomery arithmetic in #74 was merged, and I'll be pushing the x25519-dalek crate to crates.io later this evening. I still haven't optimised it, but it works. Anyone should feel free to open a ticket for combining `differential_add` and `differetial_double` and start working on it.

mcginty:
@isislovecruft sehr fantastisch 💃 Sorry right back for taking so long to respond to this - just moved back to the bay temporarily and took a convenient coding break in the process.

I'm in the middle of reading the papers you referenced, and will create a dalek branch in snow!

PR #74, opened 2017-09-14
Add Montgomery arithmetic

isislovecruft:
This PR adds arithmetic for projective Montgomery points (as requested in #47), required for the [x25519-dalek](https://github.com/isislovecruft/x25519-dalek) crate.

Please review, then I'll merge the changes to subtle, and then we can remove the temporary cargo git branch dependency on subtle. <3


Cargo.toml
```diff
@@ -35,6 +35,8 @@ version = "0.6"
 [dependencies.subtle]
 version = "^0.2"
 default-features = false
+git = "https://github.com/isislovecruft/subtle"
+branch = "feature/conditional-swap_r1"
```
hdevalence: Assuming this is going to be swapped out when you tag a new subtle release

src/constants_32bit.rs
```diff
@@ -75,6 +75,9 @@ pub const HALF: FieldElement32 = FieldElement32([
 pub const A: FieldElement32       = FieldElement32([
     486662, 0, 0, 0, 0, 0, 0, 0, 0, 0, ]);
 
+/// `APLUS2_OVER_FOUR` is (A+2)/4. (This is used internally within Montgomery laddering.)
```
hdevalence: nitpick: "within the Montgomery ladder"?

src/decaf.rs
```diff
@@ -194,6 +194,11 @@ impl<'de> Deserialize<'de> for DecafPoint {
 pub struct DecafPoint(pub ExtendedPoint);
 
 impl DecafPoint {
+    /// Convert this `DecafPoint` to its underlying `ExtendedPoint`.
```
hdevalence: I don't think this conversion should exist, since the underlying point is an implementation detail.

src/decaf.rs
```diff
@@ -752,7 +757,7 @@ mod test {
     fn decaf_decompress_id() {
         let compressed_id = CompressedDecaf::identity();
         let id = compressed_id.decompress().unwrap();
-        assert_eq!(id.0.compress_edwards(), CompressedEdwardsY::identity());
+        assert_eq!(id.to_edwards().compress(), CompressedEdwardsY::identity());
```
hdevalence: I think this should just be `0.compress()` since it's OK for the tests to use the internal structure.

src/edwards.rs
```diff
@@ -123,7 +123,6 @@ impl CompressedEdwardsY {
     }
 
     /// Copy this `CompressedEdwardsY` to an array of bytes.
-    /// XXX is this useful?
```
hdevalence: Did we figure out whether it makes sense to have both `as_bytes` and `to_bytes`?

src/edwards.rs
```diff
@@ -447,32 +446,67 @@ impl ProjectivePoint {
...
+    ///      v → V/W
+    ///
+    /// thus the Montgomery curve equation (1) becomes
+    ///
+    ///      E_(A,B) : BV²W = U³ + AU²W + UW² ⊆ 𝗣^2  <span style="float: right">(5)</span>
+    ///
+    /// Here, again, to differentiate from points in the twisted Edwards model, we
+    /// call the point `(x,y)` in affine coordinates `(u,v)` and similarly in projective
+    /// space we use `(U:V:W)`.  However, since (as per Montgomery's original work) the
+    /// v-coordinate is superfluous to the definition of the group law, we merely
```
hdevalence: nit: rather than "superfluous", should be something like "scalar multiplication can be computed using only the u-coordinate" since the Montgomery line doesn't inherit the group law from the curve

src/field.rs
```diff
@@ -197,11 +197,15 @@ impl FieldElement {
...
 
     /// Given a nonzero field element, compute its inverse.
+    ///
     /// The inverse is computed as self^(p-2), since
     /// x^(p-2)x = x^(p-1) = 1 (mod p).
-    ///
-    /// XXX should we add a debug_assert that self is nonzero?
+    //
+    // XXX do we want the debug assertion to check for zero? it breaks behaviour
+    // such as that such as in curve25519_dalek::montgomery::test::identity_to_monty.
```
hdevalence: not relevant for this PR, but I guess we should probably specify in all of our code (inv, invsqrt, sqrt_ratio) how we handle zeros

src/montgomery.rs
```diff
@@ -65,7 +91,7 @@ impl CompressedMontgomeryU {
     /// * `v` is not square.
     //
     // XXX any other exceptional points for the birational map?
-    pub fn decompress(&self) -> Option<ExtendedPoint> {
+    pub fn decompress_edwards(&self) -> Option<ExtendedPoint> {
```
hdevalence: I don't remember where we'd actually want to use this, instead of just lifting a MontgomeryPoint to an ExtendedPoint... Should we keep it around?

src/montgomery.rs
```diff
@@ -84,6 +110,23 @@ impl CompressedMontgomeryU {
...
+    /// Going from affine to projective coordinates, we have:
+    ///
+    ///      u → U/W
+    ///
+    /// # Returns
+    ///
+    /// A projective `MontgomeryPoint` corresponding to this compressed point.
+    pub fn decompress(&self) -> MontgomeryPoint {
+        MontgomeryPoint{
+            // XXX is it a problem here if we're not using a canonical encoding? —isis
```
hdevalence: I think the only room for encoding malleability here is from the field elements, and I'm not sure what the problem would be with decoding `2^255 - 18` as `1`... if people need a canonical encoding I guess I think they should use Ristretto?

src/edwards.rs
hdevalence: oops, I see that you already did this below, just here the same change is missing

src/montgomery.rs
```diff
@@ -150,34 +193,256 @@ impl CompressedMontgomeryU {
...
+///
+/// Because a compressed point on the Montgomery form of the curve doesn't
+/// include the sign bit, there's two points here (if translated from the
+/// Edwards form) which will equate.
+///
+/// # Returns
+///
+/// `1` if the points are equal, and `0` otherwise.
+impl Equal for MontgomeryPoint {
+    fn ct_eq(&self, that: &MontgomeryPoint) -> u8 {
```
hdevalence: We can save an inversion here since `(U1:W1) = (U2:W2)` iff `U1*W2 == U2*W1` (since `U1/W1 == U2/W2`)

src/montgomery.rs
```diff
@@ -150,34 +193,256 @@ impl CompressedMontgomeryU {
...
+        slices_equal(self.compress().as_bytes(),
+                     that.compress().as_bytes())
+    }
+}
+
+/// Determine if this `MontgomeryPoint` is valid.
+///
+/// # Note
+///
+/// All points, except for `(X:W) = (0:0)`, are valid, since the projective
```
hdevalence: nit: I think this should just say that all projective points are valid, and `(0:0)` is the only point of `F_p^2` which does not represent a projective point.

src/montgomery.rs
```diff
@@ -150,34 +193,256 @@ impl CompressedMontgomeryU {
...
+    /// define *pseudo-addition* for a singular coordinate.
+    ///
+    /// # Warning
+    ///
+    /// If the `difference` is the identity point, or a two torsion point, the
+    /// results of this method are not correct, but instead result in `(0:0)`
+    /// (an invalid projective point in the Montgomery model).
+    ///
+    /// The doubling case is degenerate, in that using this method to accomplish
+    /// point doubling is less efficient than using `differential_double()`.
```
hdevalence: this can't be used for doubling at all, since then the difference would be the identity...

src/montgomery.rs
```diff
@@ -150,34 +193,256 @@ impl CompressedMontgomeryU {
...
+    type Output = MontgomeryPoint;
+
+    fn mul(self, scalar: &'b Scalar) -> MontgomeryPoint {
+        let mut x0: MontgomeryPoint = MontgomeryPoint::identity();
+        let mut x1: MontgomeryPoint = *self;
+
+        let bits: [i8; 256] = scalar.bits();
+
+        for i in (0..255).rev() {
+            let mask: u8 = (bits[i+1] ^ bits[i]) as u8;
```
hdevalence: Does this handle the case `bits[255] = 1` correctly? Can we add a test with a scalar with the highest bit set?

src/constants_32bit.rs
isislovecruft: What's weird about saying that?

src/decaf.rs
isislovecruft: ack, yeah, removed in d39cb275

src/decaf.rs
isislovecruft: yep, also in d39cb275

src/edwards.rs
isislovecruft: I think it's nice syntactic sugar, and I'm using both in other crates? Also it is nice because it lets users know it's okay to do something, because anytime you're doing a `.0` somewhere you should feel like you're touching something you shouldn't.

src/edwards.rs
isislovecruft: Oh, github is showing you an old diff? This was fixed in c08591a7

Cargo.toml
isislovecruft: I just tagged it like 10 minutes ago, so this is reverted in 49f87816

src/field.rs
isislovecruft: yeah we should make an issue to note that

src/montgomery.rs
isislovecruft: I kept it around because I wasn't yet fully convinced that the conversion equations we worked out on the board were correct and non-exceptional. I have a git stash where I was working on it, but since the only benefit of that would be speed, I didn't see a reason to remove this until the less kludgy thing is finished properly. Do you feel like Montgomery P² ↦ Edwards P³ conversion would be a blocker for 1.0.0?

src/montgomery.rs
isislovecruft: Oh, good call. It's now down to ~300ns (from ~27000ns) in commit 5e6e6c3fa8

src/montgomery.rs
isislovecruft: Okay I reworded it a bit in 7b378ada6

src/montgomery.rs
isislovecruft: ca5b58c2b75

src/montgomery.rs
isislovecruft: Agreed. Do you think I should remove the comment?

src/montgomery.rs
isislovecruft: Added a test in 9da24d8af but I had to mark it with

```rust
#[should_panic(expected = "assertion failed: self[31] <= 127")]
```
since `Scalar::radix_16` has a debug assertion in there to catch it.

src/constants_32bit.rs
isislovecruft: Oh, I see, I thought my comment was saying "within the Montgomery ladder" and you were saying that was weird, but instead I'm *not* saying that and it *is* in fact weird.

src/montgomery.rs
isislovecruft: Removed the comment in d39e47ff11

isislovecruft:
Okay, I think I've addressed all the review issues. Merging for 0.13.0.

Whooooooo, now I can go show off my adorable x25519-dalek documentation which has kittens!!!

hdevalence:
LGTM

PR #78, opened 2017-09-25
optimized scalar implementations for 32/64 bit

floodyberry:
I (unfortunately) stomped on the scalar multiplication/squaring patches with this, so I apologize for that! 

These are 32/64 bit scalar implementations with add/sub/mul/square + montgomery reduction. Plain mul/square are done with two montgomery reductions since they aren't used very often and it makes the code much cleaner. 32 bit mul uses a karatsuba step, but it's hard to tell a difference either way with how little it's used.

Tests/documentation should be pretty decent. I tried to make the structure of the multiplications visible with the spacing, that makes it more obvious to me what is going on versus collapsing everything. Otherwise I tried to match the formatting in the rest of the project.

### DIFF
--- a/src/constants_32bit.rs
+++ b/src/constants_32bit.rs
@@ -19,6 +19,7 @@
 #![allow(non_snake_case)]
 
 use field_32bit::FieldElement32;
+use scalar_32bit::Scalar32;
 use edwards::ExtendedPoint;
 use edwards::AffineNielsPoint;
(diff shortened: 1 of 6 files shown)


isislovecruft:
@floodyberry! Hi!

Cool! Thanks a bunch for this patch! @hdevalence and I will review it together one night this week after work.

PR #81, opened 2017-10-23
Ristretto encoding

hdevalence:
This renames `decaf` to `ristretto`, changes the encoding procedure to match Mike's tweaks, and removes `ristretto` from the `yolocrypto` feature. (The `yolocrypto` feature is kept, even though it doesn't do anything at the moment, so that we can use it for more experimental features later).

There are some notes on the encoding, but they're not finished and shouldn't be regarded as a real write-up. Otherwise I think the code is functionally complete.


src/ristretto.rs
```diff
@@ -64,63 +65,65 @@ impl CompressedRistretto {
...
+        // 1.c) s is nonnegative
+        //
+        // Our decoding routine ignores the high bit, so the only
+        // possible failure for 1.b) is if someone encodes s in 0..18
+        // as s+p in 2^255-19..2^255-1.  We can check this by
+        // converting back to bytes, and checking that we get the
+        // original input, since our encoding routine is canonical.
+
         let s = FieldElement::from_bytes(self.as_bytes());
+        let s_bytes_check = s.to_bytes();
```
isislovecruft: Maybe we should add a `FieldElement.as_bytes()`?

src/ristretto.rs
```diff
@@ -64,63 +65,65 @@ impl CompressedRistretto {
...
-
-        // Batch inversion: set b = 1/sqrt(us^2 Z^2)
-        let (ussZZ_is_nonzero_square, b) = ussZZ.invsqrt();
-        if (ussZZ_is_nonzero_square | uss.is_zero()) == 0u8 {
-            return None; // us^2 is nonzero nonsquare
-        }
+        let yden = &FieldElement::one() + &ss;
+        let ynum = &FieldElement::one() - &ss;
+        let yden_sqr = yden.square();
+        let xden_sqr = &(&(-&constants::d) * &ynum.square()) - &yden_sqr;
```
isislovecruft: Comments would be good…

src/ristretto.rs
```diff
@@ -191,151 +194,43 @@ pub struct RistrettoPoint(pub ExtendedPoint);
...
-        //       =  invsqrt(a-d) * invsqrt(Z^2 - Y^2)
-        // otherwise:
-        //   (X : Y : Z : T) <--- (i*Y_0 : i*X_0 : Z_0 : -T_0)
-        //   r <--- invsqrt(a-d) * (T_0 * Z_0) * W
-        //       =  invsqrt(a-d) * invsqrt(Z_0^2 + X_0^2)
-        //       =  invsqrt(a-d) * invsqrt(Z^2 - Y^2)
-        //
-        // The rest of the compression follows the steps in the
-        // appendix of the Decaf paper.
-
```
isislovecruft: Why are we removing all the comments explaining the encoding?

src/ristretto.rs
```diff
@@ -757,7 +652,13 @@ mod test {
     fn decompress_id() {
         let compressed_id = CompressedRistretto::identity();
         let id = compressed_id.decompress().unwrap();
-        assert_eq!(id.0.compress(), CompressedEdwardsY::identity());
+        let mut identity_in_coset = false;
+        for P in &id.coset4() {
```
isislovecruft: Maybe we should later add some syntactic sugar for determining whether something is in the coset?

src/ristretto.rs
```diff
@@ -187,7 +188,7 @@ impl<'de> Deserialize<'de> for RistrettoPoint {
 
 /// A point in a prime-order group.
 ///
-// XXX think about how this API should work
+/// XXX think about how this API should work
```
isislovecruft: "XXX" comments should not be docstrings, imho.

PR #83, opened 2017-10-31
Prune constants file

hdevalence:
This depends on the Ristretto PR (#81), just because Ristretto uses different constants than Decaf.

This PR removes all unused constants from the `constants` module, renames them to have `DESCRIPTIVE_UPPER_CASE_NAMES`, and adds some missing docs.

It also makes nearly all of the numeric constants `pub(crate)`, so that we're not bound to providing them forever. (Most computations people would do with, e.g., the Edwards `d` value, should IMO be done in-crate).

### DIFF
--- a/src/constants_32bit.rs
+++ b/src/constants_32bit.rs
@@ -20,46 +20,46 @@ use edwards::AffineNielsPoint;
 use edwards::EdwardsBasepointTable;
 
 /// Edwards `d` value, equal to `-121665/121666 mod p`.
-pub const EDWARDS_D: FieldElement32       = FieldElement32([
+pub(crate) const EDWARDS_D: FieldElement32       = FieldElement32([
     -10913610,  13857413, -15372611,   6949391,    114729,
     -8787816,   -6275908,  -3247719, -18696448, -12055116, ]);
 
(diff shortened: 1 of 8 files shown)


src/constants_32bit.rs
```diff
@@ -71,35 +51,18 @@ pub const MSQRT_M1: FieldElement32 = FieldElement32([
...
 pub const APLUS2_OVER_FOUR: FieldElement32 = FieldElement32([121666, 0, 0, 0, 0, 0, 0, 0, 0, 0]);
 
-/// `SQRT_MINUS_A` is sqrt(-486662)
-// XXX I think that this was used in Adam's code for his elligator
-// implementation, but that should maybe be using sqrt(-486664)
-// instead...?  - hdevalence
-pub const SQRT_MINUS_A: FieldElement32 = FieldElement32([ // sqrtMinusA
-    12222970,    8312128,  11511410,  -9067497,  15300785,
-    241793,    -25456130, -14121551,  12187136,  -3972024, ]);
-
```
isislovecruft: This is used in the elligator2 implementation which matches AGL's golang implementation (albeit none of the others, since all of them do different things).

I guess I should add it back in if/when we merge the elligator2 implementations?

PR #84, opened 2017-11-03
Refactor 32bit field arithmetic code

hdevalence:
Refactor the 32bit field arithmetic code to have clearer code (and use unsigned limbs), and use a multi-stage build to eliminate most of the hardcoded constants.  This is dependent on PR #83.


isislovecruft:
@hdevalence Should we maybe add a Travis target for `--feature=precomputed-tables`?

src/field_32bit.rs
```diff
@@ -30,14 +30,11 @@ use core::ops::Neg;
...
-use utils::{load3, load4};
-
 /// A `FieldElement32` represents an element of the field GF(2^255 - 19).
 ///
 /// In the 32-bit implementation, a `FieldElement32` is represented in
-/// radix 2^25.5 as ten `i32`s, so that an element t, entries
-/// t[0],...,t[9], represents the integer t[0]+2^26 t[1]+2^51
-/// t[2]+2^77 t[3]+2^102 t[4]+...+2^230 t[9].
+/// radix 2^25.5 as ten `u32`s, so that an element t, entries
+/// t[0],...,t[9], represents `sum(t[i]*2^ceil(i*51/2))`.
```
isislovecruft: This comment doesn't actually get clearer (IMHO) unless you already understand that the radix is alternating between 25 and 26 bits.

src/field_32bit.rs
```diff
@@ -108,86 +118,101 @@ impl<'b> MulAssign<&'b FieldElement32> for FieldElement32 {
...
     type Output = FieldElement32;
     fn mul(self, _rhs: &'b FieldElement32) -> FieldElement32 {
-        // Notes preserved from ed25519.go (presumably originally from ref10):
+        /// Helper function to multiply two 32-bit integers with 64 bits
+        /// of output.
+        #[inline(always)]
+        fn m(x: u32, y: u32) -> u64 { (x as u64) * (y as u64) }
+
+        // Alias self, _rhs for more readable formulas
+        let x: &[u32;10] = &self.0; let y: &[u32;10] = &_rhs.0;
```
isislovecruft: Whitespace: s/&[u32;10]/&[u32; 10]/g

hdevalence:
Looks like I need to rebase this to account for the new Scalar code in the build system.

PR #86, opened 2017-11-18
Refactor internal code

hdevalence:
Refactor the codebase to make it more clear what is supposed to be an implementation detail and what's supposed to be public API.

The `backend` module contains submodules corresponding to the different implementations.  The `curve_models` module contains internal representations.  The other files make up the public API.


src/edwards.rs
```diff
@@ -269,10 +269,76 @@ impl ExtendedPoint {
...
 
-    /// Convert this point to its equivalent on the Montgomery form of the
-    /// curve.
+    /// Convert this `ExtendedPoint` on the Edwards model to the
+    /// corresponding `MontgomeryPoint` on the Montgomery model.
+    ///
+    /// Note that this is a one-way conversion, since the Montgomery
+    /// model does not retain sign information.
+    ///
+    // XXX need to figure out how to keep this in internal docs, and
```
isislovecruft: why does this need to be in internal docs?

src/edwards.rs
```diff
@@ -647,17 +647,13 @@ impl EdwardsBasepointTable {
 
 impl ExtendedPoint {
     /// Multiply by the cofactor: compute `8 * self`.
-    ///
-    /// Convenience wrapper around `mult_by_pow_2`.
-    #[inline]
```
isislovecruft: why are we not inlining this?

src/edwards.rs
```diff
@@ -690,14 +690,14 @@ impl ExtendedPoint {
     /// Returns `Some<[u8;32]>` if `self` is in the image of the
     /// Elligator2 map.  For a random point on the curve, this happens
     /// with probability 1/2.  Otherwise, returns `None`.
-    pub fn to_uniform_representative(&self) -> Option<[u8; 32]> {
+    fn to_uniform_representative(&self) -> Option<[u8; 32]> {
```
isislovecruft: fwiw, we can probably just remove these entirely since I was planning to refactor my elligator2 code to implement `std::convert::From`.

isislovecruft:
Finally, we have `traits` module. Cool.

src/edwards.rs
hdevalence: just because I would rather not go down the road of "mess with llvm inlining cost model" unless necessary

src/edwards.rs
hdevalence: my reasoning was that the formulas that it uses internally are an implementation detail, not something the API consumer has to know about.

PR #88, opened 2017-11-23
Refactor scalar api

hdevalence:
This changeset converts the `Scalar` struct from a tuple struct to an opaque struct and ensures that every `Scalar` meets the preconditions for scalar multiplication.


Makefile
```diff
@@ -1,3 +1,7 @@
 
 doc:
 	cargo rustdoc --features "nightly yolocrypto" -- --html-in-header rustdoc-include-katex-header.html
+
+doc-internal:
+	cargo rustdoc --features "nightly yolocrypto" -- --html-in-header rustdoc-include-katex-header.html --no-defaults --passes "collapse-docs" --passes "unindent-comments"
```
isislovecruft: Should `make doc-internal` be what runs on docs.rs?

PR #90, opened 2017-11-28
Use clear_on_drop to erase scalar digits in multiscalar_mult

hdevalence:
This relies on cesarb/clear_on_drop#16 so it will need to be fixed up before it could be merged, but it's here for review anyways due to the below:

I'm not sure if there's a great way to use `ClearOnDrop` with the lookup tables of points, since they're not zeroable, and we'd want something like `impl<T: Default> InitializableFromZeroed for [T]` which isn't compatible with the existing `impl<T: ZeroSafe> InitializableFromZeroed for [T]`.  In fact I think we can't implement `InitializableFromZeroed for [T]` for any `T` due to the orphan rules.

We could put the lookup table in a newtype wrapper and impl `ZeroSafe` on that, but the downside is that since "all zero bytes" isn't a valid point, it would, in theory, be possible to write code in `dalek` that would use a lookup table after it had been zeroed and get invalid points.  Since the type wouldn't be exported I don't think this is a big deal.


hdevalence:
This branch now relies on the released version of `clear_on_drop` and erases the point lookup tables after they're used.

PR #92, opened 2017-12-01
Texify more formulas + Scalar API tweak

hdevalence:
This changeset converts more docs to use KaTeX.  Also there is a tweak to the `Scalar` API: `reduce_wide` is renamed to `from_bytes_mod_order_wide` to fit better with the other `Scalar` constructors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
+FEATURES := nightly yolocrypto
 
 doc:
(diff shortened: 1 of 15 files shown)


hdevalence:
(Sorry for mixing an API tweak with doc changes, but the reason I thought it was a good idea was because I saw it while reviewing the rest of the docs).

PR #94, opened 2017-12-04
Experimental ('yolocrypto') AVX2 implementation using 4-way parallel Edwards formulas

hdevalence:
This branch adds an experimental (`yolocrypto`) backend using the unstable SIMD intrinsics from the [stdsimd](https://github.com/rust-lang-nursery/stdsimd) crate.  The code has a bunch of rough edges that should probably be fixed up before merging it, and it's definitely `yolocrypto`.

It uses AVX2 to implement 4-way parallel Edwards formulas from the Hisil-Wong-Carter-Dawson paper that introduced extended Edwards formulas.  The implementation notes are done using Rustdoc -- there's a temporary copy of the rendered notes [here](https://s3-us-west-2.amazonaws.com/hdevalence-tmp-docs-bucket/doc/curve25519_dalek/backend/avx2/index.html).

The scalar multiplication algorithms are still a work-in-progress -- they still use the same parameters for window size, etc, but using different parameters could probably give performance improvements. I only really put thought into the lowest-level point operations so far.  In particular, in the `avx2` backend the doubling performance isn't as good relative to the point addition performance.  It gives a speedup over the `u64` backend of between `1.5x` and `1.8x`, depending on the operation.

There's some glue code and `cfg`s to hook up the `ExtendedPoint` API to the `avx2` backend. From the CI it looks like there's some issues with getting that to work on stable.

I compared `ed25519-dalek` verification (using this branch) with Tor's benchmark of their copy of `ed25519-donna`.  The `avx2` code achieves parity, but the `-dalek` `u64` backend is 40% slower. Linking to `-donna` directly from Rust and using FFI gives different results (`-dalek` `u64` is on par) so it would be good to find out what Tor is doing.


hdevalence:
After checking in more detail, the problem was an error in the way that the FFI `-donna` was being configured and built -- the Tor numbers are the accurate ones, and the FFI ones aren't.  I hacked in some extra timers to the `ed25519-donna` benchmarks (here: https://github.com/hdevalence/ed25519-donna/tree/hack-internal-bench) to investigate.  The `donna` benchmarks were compiled with
```
clang -O3 -march=skylake ed25519.c test.c -DED25519_TEST -lssl -lcrypto -o ed25519-donna-test-clang-modified
```

The double-base scalarmul takes `140 kcy`, and point decompression takes `12.7 kcy`.  In contrast, the AVX2 code in `dalek` does a double-base scalarmul in `101 kcy`, but takes `18 kcy` for decompression.  I guess this means that the implementation strategy for the inversion ladder of "rely on rustc to do move forwarding" doesn't work well, and we should optimize that part.

src/backend/avx2/edwards.rs
```diff
@@ -0,0 +1,1057 @@
...
+            t0.0[3] = t0.0[3] + t1.0[3];
+            t0.0[4] = t0.0[4] + t1.0[4];
+
+            // Set t0 = (X1 Y1 Z1 X1+Y1)
+            // why does this intrinsic take an i32 for the imm8 ???
+            t0.0[0] = _mm256_blend_epi32(P.0[0].into(), t0.0[0].into(), D_LANES as i32).into();
+            t0.0[1] = _mm256_blend_epi32(P.0[1].into(), t0.0[1].into(), D_LANES as i32).into();
+            t0.0[2] = _mm256_blend_epi32(P.0[2].into(), t0.0[2].into(), D_LANES as i32).into();
+            t0.0[3] = _mm256_blend_epi32(P.0[3].into(), t0.0[3].into(), D_LANES as i32).into();
+            t0.0[4] = _mm256_blend_epi32(P.0[4].into(), t0.0[4].into(), D_LANES as i32).into();
```
isislovecruft: What are we `into()`ing into? And why are we doing `as` type conversions every single time?

src/backend/avx2/edwards.rs
```diff
@@ -0,0 +1,1057 @@
...
+            tmp = tmp.double();
+        }
+        tmp
+    }
+}
+
+impl<'a, 'b> Add<&'b CachedPoint> for &'a ExtendedPoint {
+    type Output = ExtendedPoint;
+
+    /// Uses a slight tweak of the parallel unified formulas of HWCD'08
```
isislovecruft: Maybe document what was tweaked more explicitly?

src/backend/avx2/edwards.rs
hdevalence: The problem is that the `_mm256_blend_epi32` intrinsic is defined to take an `i32x8` instead of a `u32x8`, even though the actual instruction works on any 256-bit integer data, shuffling 32-bit lanes according to the immediate. All of these casts disappear from the generated code. It would be nice to file bugs in `stdsimd` to try to make them go away.

src/edwards.rs
```diff
@@ -432,32 +432,41 @@ impl<'a, 'b> Mul<&'b Scalar> for &'a ExtendedPoint {
     /// For scalar multiplication of a basepoint,
     /// `EdwardsBasepointTable` is approximately 4x faster.
     fn mul(self, scalar: &'b Scalar) -> ExtendedPoint {
-        // Construct a lookup table of [P,2P,3P,4P,5P,6P,7P,8P]
-        let lookup_table = LookupTable::<ProjectiveNielsPoint>::from(self);
+        // If we built with AVX2, use the AVX2 backend.
+        #[cfg(all(feature="nightly", all(feature="avx2_backend", target_feature="avx2")))] {
```
isislovecruft: What is the reasoning behind having the `#[cfg(…)]` inside the function rather than just have different functions with the same interface that are chosen at compile time?

src/backend/avx2/edwards.rs
hdevalence: I can add a note that the tweaks are documented in the module-level docs.

src/edwards.rs
```diff
@@ -1049,7 +1086,7 @@ mod test {
 
     /// Test precomputed basepoint mult
     #[test]
-    #[cfg(feature="basepoint_table_creation")]
+    #[cfg(feature="precomputed_tables")]
```
isislovecruft: Nice

src/lib.rs
```diff
@@ -11,10 +11,9 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(feature = "alloc", feature(alloc))]
 #![cfg_attr(feature = "nightly", feature(i128_type))]
+#![cfg_attr(feature = "nightly", feature(cfg_target_feature))]
 #![cfg_attr(feature = "bench", feature(test))]
-#![cfg_attr(all(feature = "nightly", feature = "std"), feature(zero_one))]
```
isislovecruft: We got rid of `zero_one`? (What even was using it?)

src/lib.rs
```diff
@@ -11,10 +11,9 @@
...
 
-#![allow(unused_features)]
```
isislovecruft: FWIW, this was added originally because whatever ancient version of the rustc I had at the time we first started working on this would say that features were unused that _were_ actually used and I just wanted it to STFU. Glad we can remove it now. :)

isislovecruft:
I'm merging with explicit, careful code review of the integration of the new backend to the rest of the user-facing code. The backend itself is "yolocrypto" for now: it _looks_ correct to me, but if you use it and your shit breaks in some crazy way, it's not my fault.

src/edwards.rs
hdevalence: Right now the backend selection is done totally statically, choosing the AVX2 backend if `yolocrypto` is on and the target supports AVX2. In the future we might want to do dynamic dispatch, so that the library could have two codepaths, and in that case the selection has to be done inside the highest-level user-facing API.

It would probably be cleaner to keep the non-vectorized implementation in another, seperate function, but it's a bit awkward because that implementation works with some backends (the `u64` and `u32` ones) but not this one.

src/lib.rs
hdevalence: I think we used it for `FieldElement`s a year ago

PR #95, opened 2017-12-16
Update digest => 0.7, generic_array => 0.9

tarcieri:
These are the latest releases of these crates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ optional = true
 version = "0.3"
 
 [dependencies.digest]
-version = "0.6"
+version = "0.7"
 
 [dependencies.subtle]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -33,27 +33,27 @@ optional = true
...
-version = "0.6"
+version = "0.7"
 
 [dependencies.subtle]
 version = "^0.3"
 default-features = false
 
 [dependencies.generic-array]
 # same version that digest depends on
-version = "^0.8"
```
tarcieri: The `^` character is redundant in this case, as semver is the default matching used by cargo

isislovecruft:
Thanks @tarcieri!

PR #96, opened 2017-12-31
Speed up Montgomery ladder by 20-30% by combining differential add and double

isislovecruft:


isislovecruft:
It still needs a docstring but I'm not sure how much to detail since it's not ever going to be publicly exposed.

isislovecruft:
Cool I'll merge and tag a 0.14.2

PR #100, opened 2018-01-20
Optimize 64-bit code

hdevalence:
Clean up the 64-bit code and be a bit tighter with bounds estimations:

```text
 name                                                                bench-ed7345 ns/iter  bench-0e9b8e ns/iter  diff ns/iter    diff %  speedup
 edwards::bench::basepoint_mult                                      13,144                11,695                      -1,449   -11.02%   x 1.12
 edwards::bench::edwards_compress                                    4,634                 3,924                         -710   -15.32%   x 1.18 
 edwards::bench::edwards_decompress                                  4,928                 4,176                         -752   -15.26%   x 1.18
 edwards::bench::scalar_mult                                         51,228                44,954                      -6,274   -12.25%   x 1.14
 edwards::bench::ten_fold_scalar_mult                                191,991               171,588                    -20,403   -10.63%   x 1.12
 edwards::bench::vartime::bench_double_scalar_mult_basepoint         52,876                46,450                      -6,426   -12.15%   x 1.14
 edwards::bench::vartime::ten_fold_scalar_mult                       139,645               126,082                    -13,563    -9.71%   x 1.11
 field::bench::batch_1024_inv                                        302,223               269,882                    -32,341   -10.70%   x 1.12
 field::bench::batch_128_inv                                         31,325                28,042                      -3,283   -10.48%   x 1.12
 field::bench::batch_16_inv                                          6,685                 5,803                         -882   -13.19%   x 1.15
 field::bench::fieldelement_a_inv                                    4,573                 3,874                         -699   -15.29%   x 1.18
 montgomery::bench::montgomery_compress                              4,612                 3,905                         -707   -15.33%   x 1.18
 montgomery::bench::montgomery_ladder                                55,061                48,786                      -6,275   -11.40%   x 1.13
 ristretto::bench::compression                                       5,175                 4,357                         -818   -15.81%   x 1.19
 ristretto::bench::decompression                                     5,117                 4,324                         -793   -15.50%   x 1.18
 ristretto::bench::double_and_compress_1024_random_points            752,041               614,705                   -137,336   -18.26%   x 1.22
 ristretto::bench::double_and_compress_128_random_points             87,811                71,580                     -16,231   -18.48%   x 1.23
 ristretto::bench::double_and_compress_16_random_points              13,756                11,318                      -2,438   -17.72%   x 1.22
```
Benchmarks are on an `Intel(R) Core(TM) i7-7800X CPU @ 3.50GHz` without Turbo Boost.

PR #102, opened 2018-01-24
update subtle dependency to 0.5

isislovecruft:


isislovecruft:
Ugh, github isn't showing 460355d5e06d590ab289ed063045793734cc4bd4 which was already accidentally included in merging #100 and then subsequently reverted. That commit is also part of this, but I think it might require a revert-the-revert commit.

PR #103, opened 2018-01-25
Rename ExtendedPoint to EdwardsPoint

isislovecruft:
Produced via `find src/ -iname "*" -exec sed --in-place -e 's/ExtendedPoint/EdwardsPoint/g' {} \;`.

In an in-person meeting, @hdevalence and I discussed renaming `ExtendedPoint` (points in extended twisted Edwards coordinates) to just `EdwardsPoint`. (Another suggestion from others as well as a thing we ourselves suggested and considered was just `Point`) We didn't really discuss why we agreed on `EdwardsPoint`, we just both agreed on the change, so here's _my_ opinion for why:

Often in programming, when we move from a generalisation to a specific case, our naming reflects that. As a concrete, personal example, in tor code, we have a `circuit_t`, which is then "subclassed" (as much as one _can_ "subclass" in C) as `origin_circuit_t` and `or_circuit_t` to differentiate between paths through the network which "originate" at a "client"(-like thing) and those which are merely circuits _through_ an OR ("onion _relay_"). There's plenty of examples of this, and in general, we've been trained as programmers to name things according to this convention, and also been trained to recognise it when we see it. Thus, `Point` would appear to be somehow some sort of superclass of a `MontgomeryPoint` or a `RistrettoPoint`… which is entirely not the case, it's just a different coordinate system.

My other compliant is one of project goals. We're a low-level cryptographic library made by cryptographers… _for cryptographers_. Naming something super generically as a `Point` seems (IMHO) like an attempt to appeal to both cryptographers (who should understand the idea of different representations or isogenisations of the same curve… or at least be capable of [reading the literature](http://eprints.qut.edu.au/33233/1/H%C3%BCseyin_Hi%C5%9Fil_Thesis.pdf)) as well as developers (to be clear, the sort of person who is looking for the API level of "I wish to sign/encrypt a thing").  We're totally not aiming for the latter group as users, and thus our naming should not pander to them.

FWIW, as an ergonomics issue, I'm okay with an alias of `pub type Point = EdwardsPoint;` but I think my first complaint about clarity/type-derivation/"subclassing" above still applies.


isislovecruft:
This is a breaking change and requires moving to 0.15 (or 1.0.0!), just to state the obvious.

hdevalence:
I think this is a good idea, but I would like to keep the `ExtendedPoint` name for the AVX2-format point type in the AVX2 backend. So I'll change that part back and merge.

hdevalence:
@isislovecruft Does that commit look OK to you?

isislovecruft:
LGTM

PR #107, opened 2018-02-02
Simplify and optimize Montgomery code.

hdevalence:
The `MontgomeryPoint` struct is now a point on the Kummer line of the Montgomery curve.

The `ProjectivePoint` struct is made private, since its only purpose is
internal to the Montgomery ladder.

The Montgomery ladder takes affine input, making it faster, and produces affine output.

The Edwards-Montgomery correspondence is simplified.


src/lib.rs
```diff
@@ -21,7 +21,7 @@
 //
 // This means that missing docs will still fail CI, but means we can use
 // README.md as the crate documentation.
-#![cfg_attr(feature = "nightly", deny(missing_docs))]
+//#![cfg_attr(feature = "nightly", deny(missing_docs))]
```
isislovecruft: Why did this get commented out?

src/montgomery.rs
```diff
@@ -36,285 +56,145 @@ use field::FieldElement;
...
-/// necessary to use `v` for a Diffie-Hellman key exchange.
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub struct CompressedMontgomeryU(pub [u8; 32]);
-
-impl CompressedMontgomeryU {
-    /// View this `CompressedMontgomeryU` as an array of bytes.
+/// Holds the \\(u\\)-coordinate of a point on the Montgomery form of
+/// Curve25519 or its twist.
+#[derive(Copy, Clone, Debug)]
+pub struct MontgomeryPoint(pub [u8; 32]);
```
isislovecruft: Should this be `pub(crate)`?

isislovecruft: Err, I mean, should this be `pub struct MontgomeryPoint(pub(crate) [u8; 32]);`

src/lib.rs
hdevalence: That's my mistake, it was for refactoring.

src/montgomery.rs
hdevalence: No, because the `MontgomeryPoint` works similarly to, say, a `CompressedEdwardsY`, in that it's now a typed wrapper around bytes.

src/lib.rs
hdevalence: This is fixed in a fixup commit now.

PR #108, opened 2018-02-02
Add basic integration tests

isislovecruft:
This should probably only be merged just before we tag 1.0.0, to avoid a mess if we do break any parts of the API before stabilising.


hdevalence:
Maybe this would be better to do as part of the `dalek-workspace` repo?

PR #114, opened 2018-03-19
Add batch scalar inversion

hdevalence:
This PR adds a function for batch `Scalar` inversion. The temporary scalars are erased from the heap in case they're secret.

The `Scalar` code got pretty messy and should be refactored later, but this just minimally adds one function.


isislovecruft:
Okay, LGTM other than the one `debug_assert!()` thing

PR #115, opened 2018-03-19
Rename `multiscalar_mult`->`multiscalar_mul` to match `Mul` traits

hdevalence:
None of the other Rust functions use `mult`, only `mul`, so we should too.


isislovecruft:
LGTM! Maybe rename the other "mult" functions to, like `mult_by_cofactor` also?

PR #117, opened 2018-03-22
Remove accumulated warnings

hdevalence:


isislovecruft:
Yeah! Maybe we should make the two `square()` fns be pub(crate) at some point?

PR #122, opened 2018-04-04
Move from external `stdsimd` crate to `core::{arch, simd}` intrinsics

hdevalence:
This change required some work, because the to-be-stabilized SIMD functions
don't allow non-constant `imm8`s.  Previously, the `stdsimd` functions had a
constifying macro that ensured that the immediates were known.  The dalek code
used this to build helper functions which would be inlined into different
places where the immediates were known.  Unfortunately, since constexprs aren't
fully supported in Rust yet, this is done by a hidden compiler attribute, and
there's no way to propagate constness.

To deal with this, some of the functions are specialized (e.g.,
`square_and_negate_D` instead of taking a mask), and others use an enum.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,6 @@ digest = "0.7"
 generic-array = "0.9"
 clear_on_drop = "=0.2.3"
 subtle = { version = "0.6", features = ["generic-impls"], default-features = false }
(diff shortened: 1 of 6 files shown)


hdevalence:
I'm going to merge this without review since it fixes the build, and this implementation is still in a draft state and is `yolocrypto`-only.

PR #123, opened 2018-04-04
Drop benchmarks from Travis

hdevalence:
They take a lot longer to run now.


tarcieri:
If you marked them as `allow_failures` in `matrix`, you could also add `fast_finish: true` to `matrix` and they would continue to run, but not block merges.

hdevalence:
That's a good suggestion. I'm not sure it makes sense to run benchmarks on Travis anyways, since there's no control over what the hardware is doing, so I'm not sure how meaningful it is. @isislovecruft ?

tarcieri:
Yeah I have generally stopped benchmarking on Travis because the results have way too much variability for me to care

Issue #76, opened 2017-09-21
Finish Ristretto implementation


isislovecruft:

hdevalence:
I have work in progress on this issue in the following branch: https://github.com/chain/curve25519-dalek/tree/feature/ristretto_r1

isislovecruft:
Some notes are here: https://github.com/isislovecruft/ristretto-paper and I'll update that at some point with more equations and implementation details for Ristretto-based generalised Schnorr signatures.

zmanian:
Ristretto paper paper is still private.

hdevalence:
The PDF notes got subsumed into the Rustdoc notes, since HTML docs are more accessible and live with the implementation.

After #124 is merged, I think the remaining things to document are:

- [ ] more motivation, why multiplying by the cofactor or whatever isn't sufficient for complex protocols, and why there's a problem for basic cases (e.g. according to the Ed25519 paper and RFC, the set of valid signatures is undefined)
- [ ] equality checking
- [ ] elligator
- [ ] the batched double-and-encode function

hdevalence:
Note: a bunch of the work above has moved to the Ristretto website.

isislovecruft:
I'm going to call this done? We have a couple remaining points for this in #148 but the remainder has since moved to https://github.com/Ristretto/website/.

isislovecruft:
Done now, I think?

hdevalence:
Yep!

PR #124, opened 2018-04-05
Update & expand Ristretto docs

hdevalence:
Expand the Ristretto docs (what it provides) and the notes (how it works).  Addresses part of #76.

EDIT: moved checklist to the tracking issue


hdevalence:
Merging this so we can host the rendered docs, but leaving #76 open (and leaving these edits subject to further change)

PR #128, opened 2018-04-09
Update docs for AVX2 backend

hdevalence:
Reorganizes the AVX2 writeup into the same `docs/` subdirectory as the Ristretto notes, then rearranges the content so it's in a more sensible order.


hdevalence:
I cut out the performance results, but I would like to add them back in after pinning to a specific `dalek` version, specific `rustc` version, etc.

PR #129, opened 2018-04-09
Pippenger multiscalar multiplication algorithm for very large inputs

oleganza:
## ⚠️ REPLACED BY https://github.com/dalek-cryptography/curve25519-dalek/pull/249 ⚠️

Currently supported Straus algorithm does not improve performance as input size grows and saturates at ≈4x improvement over naïve multiplication. Pippenger’s algorithm takes advantage of very large amount of points (>190) by avoiding premultiplication of points and instead placing points in the buckets indexed by the multiplication factor. Then, buckets are cleverly added up to have their multipliers applied automatically. The process is repeated for each "digit" (that are 6 to 15 bits wide, depending on number of input points). As a result, the cost of multiplication grows slower than linearly with respect to the input size. For 1024 points Pippenger is >60% faster than Straus.

This patch adds:

1. An implementation of `VartimeMultiscalarMul` using Pippenger algorithm.
2. Dynamic switch based on input's `size_hint()` from Straus to Pippenger (at 190 points).
2. New `DigitsBatch` API to enumerate digits of scalars of arbitrary radix in a batch. Digits are produced via an iterator API on the fly, without pre-allocation: this allows us to have wider digits (`i16`) for more efficient computation of over 10K points.

## TODO

- [ ] There is no consttime version of Pippenger. It should be pretty straightforward to add (there is only one conditional point negation that depends on the value of a scalar), but I need some help in figuring this out.
- [ ] there is no ClearOnDrop support yet.

This addresses #130.


oleganza:
Some notes on parallelizing the current implementation:

1. Each scalar is represented in fixed radix. This means, we can have `256/w` parallel threads, one per window, to add points into buckets. Once all buckets are summed up, we can join the threads and serially add-with-doublings those sums from the high window to the low one (`w` doublings, 1 addition per window). For low number of processors (e.g. CPU) this could be enough parallelization, and points being (re)added are all read-only and shared among threads in the cache.
2. In each window, points are (re)added into `2^w/2-1` buckets, based on the digit value. We can have that many threads, one per bucket, to perform these readditions in parallel using very little data. together with (1) parallelization, this gives approximately `256*(2^w/2-1)/w` threads. For `w=8..10` (1K-10K points) this is `4000-13000` threads total.
3. Adding buckets can also be parallelized with a slighly less optimal formula for serial execution, but amenable to parallelization. E.g. we can divide the triangle of `A+B+B+C+C+C+D+D+D+D+...` into sub-triangles and large squares by saving some additions and having more doublings. However, since the (1) and (2) already add enough parallelism, this may not be effective.

oleganza:
Similarly for Straus:

1. Premultiplication for dynamic bases can be parallelized (1 thread per point).
2. If we use fixed-window scalar representation, we can have 1 thread per window and add up premultiplied points in parallel. This is similar to pippenger, but we don't use buckets as points are already premultiplied. The resulting sums are added serially with `w` doublings applied in-between.

Straus has enough parallelism for CPU, but not for GPU. But that's not an issue as Straus is less effective for large input sizes than Pippenger anyway.

oleganza:
Replaced the branch with a rebase done by @hdevalence.

src/backend/avx2/scalar_mul/pippenger.rs
```diff
@@ -0,0 +1,172 @@
...
+            }
+
+            // Step 3. Prepare 2^w/2 buckets.
+            // buckets[i] corresponds to a multiplication factor (i+1).
+            let mut buckets = vec![ExtendedPoint::identity(); buckets_count];
+
+            // Step 4. Iterate over pairs of (point, scalar)
+            // and add/sub the point to the corresponding bucket.
+            // Note: when we add support for precomputed lookup tables,
+            // we'll be adding/subtractiong point premultiplied by `digits[i]` to buckets[0].
```
oleganza: Note 2: adding precomputed points to buckets[0] won't really help us with performance if we have at least one non-precomputed point. This is because we'd have to add up all buckets anyway.

benches/dalek_benchmarks.rs
```diff
@@ -14,8 +14,17 @@ use curve25519_dalek::constants;
...
+static VARTIME_MULTISCALAR_SIZES: [usize; 32] = [
+    1, 2, 4, 8,
+    16, 32, 64, 128,
+    152, 181, 215, 256,
+    304, 362, 431, 512,
+    609, 724, 861, 1024,
+    1218, 1448, 1722, 2048,
+    2435, 2896, 3444, 4096,
+    4871, 5793, 6889, 8192
+];
```
hdevalence: Where do these sizes come from? Is there a reason not to use powers-of-two?

src/lib.rs
```diff
@@ -20,7 +20,7 @@
 //
 // This means that missing docs will still fail CI, but means we can use
 // README.md as the crate documentation.
-#![cfg_attr(feature = "nightly", deny(missing_docs))]
+#![cfg_attr(all(feature = "nightly"), deny(missing_docs))]
```
hdevalence: This `all` seems unnecessary?

src/scalar.rs
```diff
@@ -24,6 +24,12 @@ use core::borrow::Borrow;
 
 use rand::{Rng, CryptoRng};
 
+#[cfg(feature = "std")]
+use std::vec::Vec;
+
+#[cfg(feature = "alloc")]
+use alloc::Vec;
```
hdevalence: I'm not sure it's necessary to import these explicitly.

benches/dalek_benchmarks.rs
oleganza: This is powers-of-two with 3 values in-between each powers to give more fine-grained benchmarks. I used it at one point to help me find the thresholds, but maybe we can simplify it now.

src/lib.rs
oleganza: Yeah, it's a leftover from some temporary flags that I used.

src/scalar.rs
```diff
@@ -683,6 +689,65 @@ impl Scalar {
...
 
+    /// Write this scalar in radix \\(2\^r\\), with \\(n = ceil(256/r)\\) coefficients in \\([-(2\^r)/2,(2\^r)/2)\\),
+    /// i.e., compute \\(a\_i\\) such that
+    /// $$
+    ///    a = a\_0 + a\_1 2\^1r + \cdots + a_{n-1} 2\^{r*(n-1)},
+    /// $$
+    /// with \\(-2\^r/2 \leq a_i < 2\^r/2\\) for \\(0 \leq i < (n-1)\\) and \\(-2\^r/2 \leq a_{n-1} \leq 2\^r/2\\).
+    /// Radix power \\(r\\) can be between 1 and 15.
+    #[cfg(any(feature = "alloc", feature = "std"))]
+    pub(crate) fn to_arbitrary_radix(&self, mut radix_power: usize) -> Vec<i16> {
```
hdevalence: Returning a `Vec<i16>` means we're doing a separate allocation for every single scalar.

Since the digits can be computed in least-significant to most-significant order without backtracking, I wonder whether it makes sense for this function to return an iterator of the digits.  This way the caller code can decide how to allocate space for the results.

Ideally this would allow getting rid of the `to_radix_16` method entirely, but I think that this isn't easy to do, since it's not possible to `.collect` into an array. (We need to store those digits without heap allocations, which means using an array. I guess we could do it manually...)

src/scalar.rs
```diff
@@ -683,6 +689,65 @@ impl Scalar {
...
+        use byteorder::{ByteOrder, LittleEndian};
+        // The last word is half the size, so it always <= 2^r/2 including the carry
+        debug_assert!(self[31] <= 127);
+
+        // Ensure that radix is in [2^1, 2^15]
+        radix_power = if radix_power > 15 { 15 } else { radix_power };
```
hdevalence: These should be `debug_assert`s rather than silently changing bad parameters, IMO.

src/scalar_mul/pippenger.rs
```diff
@@ -0,0 +1,211 @@
...
+            4
+        } else if size < 190 {
+            5
+        } else if size < 500 {
+            6
+        } else if size < 800 {
+            7
+        } else if size < 3_000 {
+            8
+        } else if size < 6_000 {
```
hdevalence: How much of a performance degradation is there by capping the window size at 8? It means that the scalar digits can fit in `i8` rather than `i16`, so they will take half the space in cache.  For instance, for `2048` scalars, we have `256/8 = 32` digits each for `64K` digits total, so using 1 byte per digit rather than 2 would save `64KB` of cache.

On the other hand, if we arrange the scalar digits properly, we may be able to achieve linear access, in which case this might not matter.

src/scalar_mul/pippenger.rs
```diff
@@ -0,0 +1,211 @@
...
+            .collect();
+        let ps: Vec<ProjectiveNielsPoint> = points
+            .into_iter()
+            .map(|p| p.borrow().to_projective_niels())
+            .collect();
+
+        // Iterate the digits from last to first so we can shift the result by w bits
+        // at the end of each iteration (e.g. w doublings).
+        let mut result = EdwardsPoint::identity();
+        for i in (0..windows_count).rev() {
```
hdevalence: Is `windows_count` supposed to be the number of digits in each scalar, so that `(0..windows_count).rev()` iterates over the digits in reverse order?

oleganza: Yeah. Window <=> digit.

src/scalar_mul/pippenger.rs
```diff
@@ -0,0 +1,211 @@
...
+    ///
+    fn vartime_multiscalar_mul<I, J>(scalars: I, points: J) -> EdwardsPoint
+    where
+        I: IntoIterator,
+        I::Item: Borrow<Scalar>,
+        J: IntoIterator,
+        J::Item: Borrow<EdwardsPoint>,
+    {
+        let mut scalars = scalars.into_iter();
+        let size = scalars.by_ref().size_hint().0;
```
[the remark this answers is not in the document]
oleganza: Nope, the estimation script only gives a rough idea. I was manually tweaking the test to fine-tune where the threshold should be. The script does not account for the real code overhead, so sometime it's off by 10-20%.

src/scalar_mul/pippenger.rs
```diff
@@ -0,0 +1,211 @@
...
+            // Step 5. Add the buckets applying the multiplication factor to each bucket.
+            // The most efficient way to do that is to have a single sum with two running sums:
+            // an intermediate sum from last bucket to the first, and a sum of intermediate sums.
+            //
+            // For example, to add buckets 1*A, 2*B, 3*C we need to add these points:
+            //   C
+            //   C B
+            //   C B A   Sum = C + (C+B) + (C+B+A)
+            let mut buckets_intermediate_sum = buckets[buckets_count - 1].clone();
+            let mut buckets_sum = buckets[buckets_count - 1].clone();
```
hdevalence: It would be good to do allocate the scratch space for this up-front, rather than doing two allocations per loop iteration

src/scalar.rs
oleganza: Yeah, it makes even more sense if we do not expose these through the API.

hdevalence:
Merged `develop` into this branch.

src/scalar_mul/pippenger.rs
oleganza: Good point, fixed in 67a6a97.

src/scalar_mul/pippenger.rs
oleganza: According to an estimation using the provided ruby script, the difference is like 5-7% if you double the points from 3K->6K. It gets to 25%-35% difference when you go to 30K and 300K points respectively.

oleganza: Also, interestingly, if we reduce the support to `max(w)=8`, then we only need `w = {6,7,8}`, because w < 6 are optimal for n<190 where Straus is faster anyway. I wonder if having only three possible window sizes is beneficial in some sense?

src/scalar.rs
oleganza: Done.

oleganza:
How reasonable are inputs of over 10K points (where availability of digits wider than 8 bits makes a difference)? 

Case 1: Bulletproof range proofs use 17 dynamic points and ≈130 static points. We need to batch-verify 588 range proofs in order to use Straus for 130 shared static points and Pippenger for the remaining 10K dynamic points.

Case 2: Aggregated range proof with 74 participants also requires ≈10K-point multi-scalar multiplications.

Case 3: 10K Schnorr signatures verified in a batch.

Finally, batch verification of 100 transactions with 64-party aggregated range proofs in each would require multiplication of over 800K points, where the speedup from wider digits could be as big as 2x.

oleganza:
Correction: we don’t need to scan digits hi->lo if we delay adding and shifting intermediate results. We’d only have to add memory for 30-40 points (one per digit), which will be processed hi->lo after we finish with the main pippenger logic.

src/scalar_mul/pippenger.rs
```diff
@@ -0,0 +1,238 @@
...
+///
+/// The algorithm works as follows:
+///
+/// Let `n` be a number of point-scalar pairs.
+/// Let `w` be a window of bits (5..15, chosen based on `n`, see cost factor).
+///
+/// 1. Prepare `2^(w-1) - 1` buckets with indices `[1..2^(w-1))` initialized with identity points.
+///    Bucket 0 is not needed as it would contain points multiplied by 0.
+/// 2. Convert scalars to a radix-`2^w` representation with signed digits in `[-2^w/2, 2^w/2]`.
+///    Note: all but last digit will never be equal `2^w/2`, but that's irrelevant to us here.
```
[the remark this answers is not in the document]
oleganza: Yeah, that's my favorite mistake (I'm not a native English speaker), thanks for pointing it out!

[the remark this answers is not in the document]
oleganza: Fixed in 5051777

oleganza:
Holy shit, this PR is one year old.

oleganza:
Going to try converting scalars to u8 digits, per Henry's suggestion above. https://github.com/dalek-cryptography/curve25519-dalek/pull/129#discussion_r188717199

hdevalence:
Cool! I can help with rebasing if you'd like. Since then I restructured the source tree so that there are `serial` and `vector`  backends, each with a `scalar_mul` module (where this code should go).

hdevalence:
@oleganza Can I help revitalize this branch?

oleganza:
⚠️ **Replacing this PR by #249**

hdevalence:
Closing in favor of #249.

PR #134, opened 2018-05-01
Implement Product and Sum traits

### LINKED ISSUE #131: Iterators should implement `Sum` and `Product`
Rust defines [`Sum`]( https://doc.rust-lang.org/std/iter/trait.Sum.html) and [`Product`](https://doc.rust-lang.org/std/iter/trait.Product.html) traits that allow sums and products of iterators.

I think it makes sense to implement:

- [ ] `impl Sum for RistrettoPoint`
- [ ] `impl Sum for EdwardsPoint`
- [ ] `impl Sum for Scalar`
- [ ] `impl Product for Scalar`

These can be implemented using `fold` internally.

Mandragorian:
This PR implements the [Product](https://doc.rust-lang.org/std/iter/trait.Product.html) and [Sum](https://doc.rust-lang.org/std/iter/trait.Sum.html) traits for Scalar.

Partially resolves #131 

I made this PR before implement the Sum trait on the other types so that you can comment on my code before trying to implement the other traits. If you would like me completely close issue #131 before making a PR just tell me.

### DIFF
--- a/src/edwards.rs
+++ b/src/edwards.rs
@@ -98,6 +98,7 @@ use core::iter::Iterator;
 use core::ops::{Add, Sub, Neg};
 use core::ops::{AddAssign, SubAssign};
 use core::ops::{Mul, MulAssign};
+use core::iter::Sum;
 use core::borrow::Borrow;
 
 use subtle::ConditionallyAssignable;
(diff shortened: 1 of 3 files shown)


hdevalence:
This looks great, especially the point tests are well-done! 

The only change I would ask for is to just implement the point types directly, rather than using a macro.  The reason is that the macro adds an extra layer of complexity, but there are only two point types and we don't have any plans to add more.

Mandragorian:
Ok, moved the definitions in each file and removed the commit introducing the macro.

hdevalence:
Thanks! The CI is failing for what looks like an unrelated reason. I'll try to fix it, then review your PR.

hdevalence:
(The CI failure is an unrelated issue in today's nightly cargo related to optional dependencies, and all of the other tests pass).

This looks great! Thanks a lot for the PR.

PR #135, opened 2018-05-03
Implement Sum for RistrettoPoint and EdwardsPoint

Mandragorian:
This PR implements the Sum trait for RistrettoPoint and EdwardsPoint.

This PR and #134 together close issue #131


hdevalence:
Could this be combined into a single PR with #134 ?

PR #138, opened 2018-05-10
Disable default features in byteorder dependency to be no_std-compatible

oleganza:
When building with [Rust SGX SDK](https://github.com/baidu/rust-sgx-sdk) `byteorder` was pulling in `std` crate which conflicted with `sgx_tstd`. Disabling default features in `byteorder` prevents this.


hdevalence:
Thanks! Sorry for introducing this problem.

PR #141, opened 2018-05-14
[WIP] upgrade to rand 0.5

### LINKED ISSUE #132: Move to `rand` 0.5 (breaking change)
The `rand` API was reworked: see https://github.com/rust-lang-nursery/rand/blob/master/UPDATING.md

We should upgrade to use it. It has a bunch of nice stuff, like a `CryptoRng` marker trait.

hdevalence:
This specifies the git repository, because `impl CryptoRng for OsRng` was only
added after the last pre-release (in https://github.com/rust-lang-nursery/rand/pull/414)

Once the 0.5 version of `rand` is released, we can edit this commit to remove
the git dependency, and then merge it.

Closes #132

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,28 +41,27 @@ harness = false
 # match exactly, since the build.rs uses the crate itself as a library.
 
 [dependencies]
-byteorder = {version = "1", default-features = false }
+rand = { version = "0.5.0-pre.2", default-features = false }
+byteorder = { version = "1", default-features = false }
 digest = "0.7"
 generic-array = "0.9"
 clear_on_drop = "=0.2.3"
 subtle = { version = "0.6", features = ["generic-impls"], default-features = false }
 serde = { version = "1.0", optional = true }
-rand = { version = "0.4", optional = true }
 
 [build-dependencies]
-byteorder = "1"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -41,28 +41,27 @@ harness = false
 # match exactly, since the build.rs uses the crate itself as a library.
 
 [dependencies]
-byteorder = {version = "1", default-features = false }
+rand = { version = "0.5.0-pre", default-features = false, git = "https://github.com/rust-lang-nursery/rand" }
```
isislovecruft: So tor is currently pinned to rand-0.5.0-pre.1 because I needed to implement an RNG using rand_core, which then required using 0.5.0… so this is still not going to resolve to the same trait, making my current task (https://bugs.torproject.org/23886) impossible. Probably I should go ask nicely for another pre-release and we should pin to that before releasing curve25519-dalek-0.17?

isislovecruft: Okay, it's done already in https://github.com/rust-lang-nursery/rand/issues/447! I'll fix up your branch to use `rand-0.5.0-pre.1`.

hdevalence: I think it should be `0.5.0-pre.2`.

PR #146, opened 2018-06-18
Refactor AVX2 code and remove its "yolocrypto" designation

### LINKED ISSUE #142: Refactor / polish AVX2 code
The current AVX2 code is a hacky mess for two reasons: it was a proof of concept, and the API it uses is itself unstable.

It should be refactored, polished, and have the `yolocrypto` feature removed.

hdevalence:
This changeset rewrites the AVX2 code from a proof-of-concept to something that matches the quality standards of the rest of our code.

It also documents the formulas used, the implementation strategy, the design tradeoffs, and the coefficient bounds.

It removes the `yolocrypto` flag from the `avx2_backend`.

Closes #142.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
   # Tests the u64 backend
   - TEST_COMMAND=test EXTRA_FLAGS='--no-default-features' FEATURES='std u64_backend'
   # Tests the avx2 backend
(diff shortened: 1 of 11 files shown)


hdevalence:
The docs for the AVX2 backend don't build because Rustdoc isn't aware of `target_feature` (upstream bug https://github.com/rust-lang/rust/issues/46317). In the meantime maybe it would be good to add a `documentation` feature that we could use to build docs for all internal modules?

PR #149, opened 2018-06-30
Sequential montgomery trick

ebfull:
The current implementation of batch inversion for field elements and scalars is using this tree-based thing which doesn't seem to ever be more efficient (both in space and time) than the sequential version that I've seen before. Local benchmarks confirm this, though I'd like others to double-check.

This PR just replaces everything with the sequential version: simpler, half as much scratch space, not (as?) sensitive to input length.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -142,58 +142,34 @@ impl FieldElement {
     /// Given a slice of public `FieldElements`, replace each with its inverse.
     ///
     /// All input `FieldElements` **MUST** be nonzero.
-    ///
-    /// This function is most efficient when the batch size (slice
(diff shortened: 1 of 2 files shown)


ebfull:
The performance gains appear to be great for an input size of `1024` on local benchmarks.

```
change: [-59.805% -59.709% -59.604%] (p = 0.00 < 0.05)
```

ebfull:
Whoops, opened against `master` even after reminding myself not to...

hdevalence:
No worries, in the future you can just edit the PR to change the target branch (I do it pretty often on people's PRs because the workflow isn't obvious).

PR #150, opened 2018-06-30
Sequential Montgomery trick

ebfull:
The current implementation of batch inversion for field elements and scalars is using this tree-based thing which doesn't seem to ever be more efficient (both in space and time) than the sequential version that I've seen before. Local benchmarks confirm this, though I'd like others to double-check.

This PR just replaces everything with the sequential version: simpler, half as much scratch space, not (as?) sensitive to input length.

The performance gains appear to be great for an input size of `1024` on local benchmarks.

```
change: [-59.805% -59.709% -59.604%] (p = 0.00 < 0.05)
```

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -142,58 +142,34 @@ impl FieldElement {
     /// Given a slice of public `FieldElements`, replace each with its inverse.
     ///
     /// All input `FieldElements` **MUST** be nonzero.
-    ///
-    /// This function is most efficient when the batch size (slice
-    /// length) is a power of 2.
     #[cfg(any(feature = "alloc", feature = "std"))]
     pub fn batch_invert(inputs: &mut [FieldElement]) {
-        // First, compute the product of all inputs using a product
-        // tree:
-        //
-        // Inputs: [x_0, x_1, x_2]
-        //
-        // Tree:
-        //
-        //                 x_0*x_1*x_2*1         tree[1]
-        //                   /       \
-        //               x_0*x_1     x_2*1       tree[2,3]
-        //                / \        / \
-        //              x_0  x_1   x_2  1        tree[4,5,6,7]
-        //
-        //  The leaves of the tree are the inputs.  We store the tree in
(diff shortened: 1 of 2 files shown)


ebfull:
I made a change to reduce the diff after @daira pointed out that my PR originally performed a debug assertion *after* inversion rather than before, unlike the previous API. This is fine because the existing API for inversion handles the zero case in a particular way, but is not fine because we would prefer to localize invariants in debug assertions and minimize the diff.

In any case, @daira and I are concerned that the existing API for inversion is behaving as though zero is an inverse of itself. This is not ideal for reducing mistakes downstream, and deserves further discussion.

hdevalence:
Thanks for writing this up! It looks good to me.

For the (internal) field element API, we use the fact that the inverse and inverse square root functions send 0 to 0 to avoid edge cases in some formulas, and I think this is documented. But that API isn't public so there aren't any downstream consequences.

I'm not sure what a better behavior for the scalar inversion API would be. If we return a `Scalar`, we have to choose something to return on input 0, and I think 0 is the best choice. Alternatively, we could return an `Option<Scalar>`, check for zero, and return `Some(inverse)` or `None`. Do you think that would be better? (In any case we are missing documentation on what the behaviour is, which should be fixed).

For batch inversion, the idea was that the results would be written in place to avoid an allocation. I'm not sure what the best way to signal failure is there. Maybe the API should be changed.

ebfull:
I do think it would be better to return `Option<Scalar>` but I don't know what the batch inversion API should look like if some values are zero and others aren't. Perhaps a `NonzeroScalar` wrapper type would help?

isislovecruft:
Hi @ebfull and @daira! Thanks for this! These changes also look good to me. The code looks much cleaner as well. Also thanks for adding tests!

For the `Option<Scalar>` API to guard against inversions of zero, I feel like it would be clunky to handle unwraps for the overwhelming majority of non-zero cases, combined with not really making sense to do an inversion unless you're already assured the things you're inverting is non-zero. We should definitely document that expectation better.

PR #152, opened 2018-07-04
Update rand dependency to 0.5.3.

isislovecruft:


Cargo.toml
```diff
@@ -41,7 +41,7 @@ harness = false
 # match exactly, since the build.rs uses the crate itself as a library.
 
 [dependencies]
-rand = { version = "0.5.0", default-features = false }
+rand = { version = "0.5.3", default-features = false }
```
hdevalence: Any reason not to specify just `0.5`?

isislovecruft: Oh, we might actually want "=0.5.3", but I'm not sure yet. Tor is also using `rand` and `rand_core` (pinned to specific versions for the offline builds) and we are about to pull in `ed25519-dalek`, so it'd be nice if all the versions lined up nicely.

tarcieri: If Tor is `= 0.5.3` Cargo will select `0.5.3` for `{curve,ed25519}-dalek` as well as it's semver-by-default

hdevalence: If `curve25519-dalek` specifies `0.5`, then `cargo` is allowed to resolve to any compatible version, so Tor (or anyone else) can still pin to a specific version, but we don't have to change the `curve25519-dalek` dependencies when that changes.

Issue #148, opened 2018-06-22
1.0.0-pre code review


hdevalence:
We need to re-review our code before releasing a 1.0 prerelease.

hdevalence:
Checklist of docs related changes:

- [ ] remove the WARNING note in the README
- [ ] add a performance section to the README
- [ ] decide what to do about imports
- [ ] move `yolocrypto` description to the bottom
- [ ] clarify wording on nightly compiler
- [ ] add a note on side channels
- [ ] change description of `EdwardsBasepointTable`, remove XXX
- [ ] remove XXX on `EdwardsPoint::vartime_double_scalar_mul_basepoint` (or remove that function)
- [ ] update edwards module docs to point to the multiscalar traits
- [ ] cherrypick multiscalar length detection from pippenger branch
- [ ] ristretto docs should point to ristretto.group for "why / what" and focus just on the API
- [ ] update `double_and_compress_batch` to remove the power of two note
- [ ] update `RistrettoPoint` docs to note that `Eq` checking is cheaper
- [ ] should `ristretto::notes` still exist?
- [ ] the `scalar` module should explain stuff about the scalar apis (distinction between integers, integers mod group order, etc).
- [ ] add examples for `Scalar` creation
- [ ] copy batch invert warning to invert function
- [ ] remove extra line in batch invert example

isislovecruft:
I started attacking some of the things on the TODO list in my `148-cleanup-for-1.0.0-pre.0` [branch](https://github.com/isislovecruft/curve25519-dalek/tree/feature/148-cleanup-for-1.0.0-pre.0).

hdevalence:
The only thing left is a note on performance.  Since that's just a README change I'd rather do it as a point release.

PR #155, opened 2018-07-05
some pre-1.0.0 cleanup

isislovecruft:
This fixes some issues in #148

### DIFF
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -535,6 +535,32 @@ impl Scalar {
     /// # Returns
     ///
     /// The multiplicative inverse of the this `Scalar`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use curve25519_dalek::scalar::Scalar;
+    ///
+    /// // x = 2238329342913194256032495932344128051776374960164957527413114840482143558222
+    /// let X: Scalar = Scalar::from_bytes_mod_order([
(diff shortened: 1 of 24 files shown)


src/scalar.rs
```diff
@@ -11,6 +11,134 @@
...
+//!
+//! let mut hasher: Sha512 = Sha512::default();
+//! let mut hash: [u8; 64] = [0u8; 64];
+//!
+//! hasher.input(b"Abolish ICE");
+//! hash.copy_from_slice(hasher.result().as_slice());
+//!
+//! let a: Scalar = Scalar::from_bytes_mod_order_wide(&hash);
+//! # }
+//! ```
```
hdevalence: I'm not sure if we want to have an example of doing this (instead of just using the scalar hashing methods), since I don't think that we want to encourage people to use this API -- it's there if they need it, but I don't think it should be highlighted.

src/scalar.rs
```diff
@@ -415,6 +525,42 @@ impl Scalar {
     }
 
     /// Compute the multiplicative inverse of this scalar.
+    ///
+    /// # Warning
+    ///
+    /// All input `Scalars` **MUST** be nonzero.  If you cannot
```
hdevalence: This should just say that `self` must be nonzero (since unlike the batch case we just have one scalar).

src/scalar.rs
hdevalence: I can change the example tomorrow.

hdevalence:
@isislovecruft Does the updated example look OK?

PR #159, opened 2018-07-17
Feature/148 cleanup for 1.0.0 pre.0 1

isislovecruft:
More various doctests and cleanups for #148.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,6 +37,9 @@ extern crate core;
 #[cfg(feature = "alloc")]
 extern crate alloc;
 
+#[cfg(test)]
+extern crate sha2;
+
 extern crate rand;
 extern crate clear_on_drop;
 extern crate byteorder;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,9 +37,6 @@ extern crate core;
 #[cfg(feature = "alloc")]
 extern crate alloc;
 
-#[cfg(test)]
-extern crate sha2;
-
 extern crate rand;
 extern crate clear_on_drop;
 extern crate byteorder;
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -533,11 +533,31 @@ impl Scalar {
     }
 
     /// Convert this `Scalar` to its underlying sequence of bytes.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use curve25519_dalek::scalar::Scalar;
+    ///
+    /// let s: Scalar = Scalar::zero();
+    ///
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -37,6 +37,9 @@ extern crate core;
 #[cfg(feature = "alloc")]
 extern crate alloc;
 
+#[cfg(test)]
+extern crate sha2;
```
hdevalence: Where is this used? I don't see anything that needs it...

src/scalar.rs
```diff
@@ -479,21 +494,70 @@ impl Scalar {
...
+    ///
+    /// # fn main() {
+    /// let mut h = Sha512::default();
+    ///
+    /// h.input(b"To really appreciate architecture, you may even need to commit a murder.");
+    /// h.input(b"While the programs used for The Manhattan Transcripts are of the most extreme");
+    /// h.input(b"nature, they also parallel the most common formula plot: the archetype of");
+    /// h.input(b"murder. Other phantasms were occasionally used to underline the fact that");
+    /// h.input(b"perhaps all architecture, rather than being about functional standards, is");
+    /// h.input(b"about love and death.");
```
hdevalence: nice

src/scalar.rs
```diff
@@ -479,21 +494,70 @@ impl Scalar {
...
+    /// let s: Scalar = Scalar::zero();
+    ///
+    /// assert!(s.to_bytes() == [0u8; 32]);
+    /// ```
     pub fn to_bytes(&self) -> [u8; 32] {
         self.bytes
     }
 
-    /// View this `Scalar` as a sequence of bytes.
+    /// View this `Scalar` as its underlying sequence of bytes.
```
hdevalence: Maybe this could be something like

> View the little-endian byte encoding of the integer representing this `Scalar`.

hdevalence: I'm not sure about saying "underlying" since it's the integer that represents the scalar, rather than the other way around.

src/scalar.rs
isislovecruft: ;)

src/scalar.rs
isislovecruft: Sure, that's more clear! Fixed in 7b22fe6.

src/lib.rs
isislovecruft: Ah, only in a doctest, good catch. I've removed it in 73a5f47. (I originally had the doctest as an actual test so that I could print output.)

isislovecruft:
Merging? I think I've addressed everything.

PR #162, opened 2018-07-17
More pre 1.0 cleanups

### LINKED ISSUE #157: Please document what constant-time guarantees this crate provides
Since this crate provides pure-Rust implementations of cryptographic primitives, the very first question I was trying to find the answer to in the README was whether it had appropriate protection against timing attacks, via constant-time primitives. (And, if so, how it managed to do so using pure Rust code.)

After some digging, it *looks* like this crate does use appropriate constant-time operations through the `subtle` crate. That crate, in turn, mentions the need to use the `nightly` feature for full protection.

Could you please document, in the README for both this crate and the ed25519/x25519 crates (either directly or via a link), the timing-attack resistance and constant-time properties, and in particular the requirement to use the `nightly` feature to fully achieve those properties?

Thank you for your work on these crates; much appreciated.

hdevalence:
Summary of changes:

- Add description of safety guarantees to the README (closes #157)
- Remove various FIXME notes
- Add missing Ristretto double-base-vartime function
- Fix ExtendedPoint validity check to check all coordinates

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -96,17 +96,10 @@ impl FieldElement {
 
     /// Compute (self^(2^250-1), self^11), used as a helper function
     /// within invert() and pow22523().
-    ///
-    /// XXX This returns an extra intermediate to save computation in
-    /// finding inverses, at the cost of an extra copy when it's not
-    /// used (e.g., when raising to (p-1)/2 or (p-5)/8). Good idea?
     fn pow22501(&self) -> (FieldElement, FieldElement) {
         // Instead of managing which temporary variables are used
-        // for what, we define as many as we need and trust the
-        // compiler to reuse stack space as appropriate.
-        //
-        // XXX testing some examples suggests that this does happen,
-        // but it would be good to check asm for this function.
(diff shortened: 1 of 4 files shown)


src/field.rs
```diff
@@ -177,12 +170,9 @@ impl FieldElement {
     ///
     /// The inverse is computed as self^(p-2), since
     /// x^(p-2)x = x^(p-1) = 1 (mod p).
-    //
-    // XXX do we want the debug assertion to check for zero? it breaks behaviour
-    // such as that such as in curve25519_dalek::montgomery::test::identity_to_monty.
+    ///
+    /// This function returns zero on input zero.
```
isislovecruft: Not sure if this should be in a `# Warning` section, or at least a different section? Up to you.

isislovecruft:
Merging now, but feel free to change that docstring if you feel it's appropriate and push the change to the same branch and I'll re-merge.

src/field.rs
hdevalence: It's only in internal docs, so I don't think it really matters.

PR #163, opened 2018-07-17
Allow Options in the VartimeMultiscalarMul trait

hdevalence:
This changes the primary function for the `VartimeMultiscalarMul` trait
to an `optional_multiscalar_mul` trait that accepts
`Option<Self::Point>` (and returns `None` if any input points are
`None`).

The existing `vartime_multiscalar_mul` is changed to be a wrapper around
this function to avoid code duplication.  This may result in an
extra copy of each input point, but that cost is probably not
significant compared to the cost of the multiscalar multiplication.

The motivation is to allow performing multiscalar multiplications with
inline decompression.  Currently, API consumers have to allocate
temporary buffers for all of their points, decompress into those
buffers, then pass (iterators over) those buffers into the multiscalar
multiplication code, which then creates new buffers for lookup tables.

### DIFF
--- a/src/edwards.rs
+++ b/src/edwards.rs
@@ -556,12 +556,11 @@ impl MultiscalarMul for EdwardsPoint {
 impl VartimeMultiscalarMul for EdwardsPoint {
     type Point = EdwardsPoint;
     
-    fn vartime_multiscalar_mul<I, J>(scalars: I, points: J) -> EdwardsPoint
+    fn optional_multiscalar_mul<I, J>(scalars: I, points: J) -> Option<EdwardsPoint>
     where
         I: IntoIterator,
         I::Item: Borrow<Scalar>,
-        J: IntoIterator,
-        J::Item: Borrow<EdwardsPoint>,
+        J: IntoIterator<Item = Option<EdwardsPoint>>,
     {
         // XXX later when we do more fancy multiscalar mults, we can
         // delegate based on the iter's size hint -- hdevalence
@@ -570,13 +569,13 @@ impl VartimeMultiscalarMul for EdwardsPoint {
         #[cfg(all(feature="avx2_backend", target_feature="avx2"))]
         {
             use backend::avx2::scalar_mul::straus::Straus;
-            Straus::vartime_multiscalar_mul(scalars, points)
+            Straus::optional_multiscalar_mul(scalars, points)
         }
         // Otherwise, proceed as normal:
         #[cfg(not(all(feature="avx2_backend", target_feature="avx2")))]
         {
             use scalar_mul::straus::Straus;
-            Straus::vartime_multiscalar_mul(scalars, points)
+            Straus::optional_multiscalar_mul(scalars, points)
         }
     }
 }
--- a/src/scalar_mul/straus.rs
+++ b/src/scalar_mul/straus.rs
@@ -152,12 +152,11 @@ impl VartimeMultiscalarMul for Straus {
     /// The non-adjacent form has signed, odd digits.  Using only odd
     /// digits halves the table size (since we only need odd
     /// multiples), or gives fewer additions for the same table size.
-    fn vartime_multiscalar_mul<I, J>(scalars: I, points: J) -> EdwardsPoint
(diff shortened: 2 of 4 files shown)


hdevalence:
I think that this API is still a little awkward -- mixing non-`Option` points with `Option`al ones requires an extra `map` call. But I'm not sure of a better way to do it. Just allowing compressed points wouldn't work, since then you wouldn't be able to mix compressed points (from proofs to be verified) with non-compressed points.

On the other hand, if you think about decompression as being a kind of precomputation, maybe that means that the "right" thing to do is have a precomputation API where the "dynamic" points are all compressed points, which get decompressed on the fly.

burdges:
`RefCell` has `try_borrow` and `Option` has `as_ref` that return `Option<&T>`s.  Would some `TryBorrow` trait abstracting this sort of thing help?  Or maybe you want to avoid having everything as `Option<&...>`?

hdevalence:
Using an `Option<&...>` doesn't work, because the goal is to allow iterator combinators to construct the `Scalar`s and `Point`s as needed -- meaning that the iterators have ownership of the values they construct.

src/edwards.rs
```diff
@@ -556,12 +556,11 @@ impl MultiscalarMul for EdwardsPoint {
 impl VartimeMultiscalarMul for EdwardsPoint {
     type Point = EdwardsPoint;
     
-    fn vartime_multiscalar_mul<I, J>(scalars: I, points: J) -> EdwardsPoint
+    fn optional_multiscalar_mul<I, J>(scalars: I, points: J) -> Option<EdwardsPoint>
```
isislovecruft: What is the reasoning for eliminating the `vartime_multiscalar_mul` method entirely? What if the user has already validated the points?

src/scalar_mul/straus.rs
```diff
@@ -167,10 +166,15 @@ impl VartimeMultiscalarMul for Straus {
...
             .map(|c| c.borrow().non_adjacent_form(5))
             .collect();
-        let lookup_tables: Vec<_> = points
+
+        let lookup_tables = match points
             .into_iter()
-            .map(|P| NafLookupTable5::<ProjectiveNielsPoint>::from(P.borrow()))
-            .collect();
+            .map(|P_opt| P_opt.map(|P| NafLookupTable5::<ProjectiveNielsPoint>::from(&P)))
+            .collect::<Option<Vec<_>>>()
```
isislovecruft: Your lil `<_>>>` dude looks like he grew a few extra eyeballs.

src/traits.rs
```diff
@@ -106,11 +106,70 @@ pub trait VartimeMultiscalarMul {
...
+    /// // It's also possible to mix compressed and uncompressed points:
+    /// let A3 = RistrettoPoint::optional_multiscalar_mul(
+    ///     abc.iter()
+    ///         .chain(abc.iter()),
+    ///     compressed.iter().map(|pt| pt.decompress())
+    ///         .chain(PQR.iter().map(|&pt| Some(pt))),
+    /// );
+    ///
+    /// assert_eq!(A3, Some(A1+A1));
+    /// ```
```
isislovecruft: These are some nice doctests. :)

burdges:
I meant roughly:
```
pub trait TryBorrow<Borrowed> where Borrowed: ?Sized, {
    fn try_borrow(&self) -> Option<&Borrowed>;
}
impl<T> TryBorrow<T> for T {
    fn try_borrow(&self) -> Option<&T> { Some(self) }
}
impl<'a,T> TryBorrow<T> for &'a T {
    fn try_borrow(&self) -> Option<&T> { Some(&*self) }
}
impl<T> TryBorrow<T> for Option<T> {
    fn try_borrow(&self) -> Option<&T> { self.as_ref() }
}
impl<'a,T> TryBorrow<T> for Option<&'a T> {
    fn try_borrow(&self) -> Option<&T> { self }
}
```
so any of `T`, `&T`, `Option<T>`, and `Option<&T>` can be treated as a `Option<&T>` generically.  You build the `IntoIterator<Item: TryBorrow<T>>` like usual but with any of these types.  I guess the cost is not significant like you say.  

I suppose `Result<T,E>` would be more generally useful, and fits the `Try*` name better, but lacks the right semantics for this use case.

PR #165, opened 2018-07-19
Replace Scalar::from_u64 with From impls

hdevalence:
Unfortunately, Rust selects `i32` as the type for an integer literal
when the literal has no other type constraints.  This means that someone
cannot write `Scalar::from(1)`, as Rust will choose `i32` as the type for
`1`, and we don't `impl From<i32> for Scalar`.

We could implement `From` conversions for signed integers, but since
`Scalar` operations should be constant-time by default, this would
require us to extract the sign bit of the integer and use it to
conditionally select between the positive and negative of Scalar
constructed from the value bits.  This is more expensive than the
unsigned operation, and I don't think it's what anyone really wants.

Making API consumers specify that their literals are unsigned is
slightly annoying, but better than the above alternative.

It would also be nice to change `Scalar::from_hash` to be
`impl<D: Digest<OutputSize = U64>> From<D> for Scalar`,
but this isn't currently allowed by Rust (since that `impl` "could"
conflict with the `impl From<u8>` if someone decided that `u8` should
`impl Digest`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ harness = false
 
 [dependencies]
 rand = { version = "0.5", default-features = false }
-byteorder = { version = "1", default-features = false }
+byteorder = { version = "1", default-features = false, features = ["i128"] }
 digest = "0.7"
 generic-array = "0.9"
 clear_on_drop = "=0.2.3"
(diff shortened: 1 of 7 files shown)


isislovecruft:
It's a little weird, but I think this is fine. In the earlier case we'd have users typing `Scalar::from_u64(42)`, whereas now they'll write `Scalar::from(42u64)` which is weirdish, but actually one less character and two less keystrokes. I don't think that's so bad.

Issue #166, opened 2018-07-19
Scalar::random shouldn't require std anymore


isislovecruft:
A robigdalia (https://robigalia.org/) developer reported to me via IRC that they were confused why the `Scalar::random` function required std. I know it used to be the case that the rand traits we were using required std, but I believe this is no longer the case. I'll try removing it and see what happens.

hdevalence:
While we're fixing up `Scalar::random`, maybe we should get rid of it entirely? I think there's a trait from the `rand` crate that types can implement in order to be "randomly generable" by any RNG.

hdevalence:
Looking into it a bit, I think that what that would look like is implementing a `Distribution` https://docs.rs/rand/0.5.4/rand/distributions/trait.Distribution.html for `Scalar`s, which people would call `sample` on.  There's also a provided `sample_iter`, which could be more convenient for generating sequences than what we're doing now.

isislovecruft:
I'm not sure exactly what that would look like? `impl rand::distributions::Distribution<Scalar> for Uniform`? But `Uniform` is their type, so the orphan rules will hate us, so we'd need to create our own `Uniform` type.. it sounds slightly messy. That should probably be a new issue in any case.

hdevalence:
We'd define our own distribution type and implement the trait for it.

PR #167, opened 2018-07-19
Make Scalar::random work with nostd

isislovecruft:


isislovecruft:
This fixes #166

src/scalar_mul/straus.rs
```diff
@@ -12,11 +12,16 @@
 
 #![allow(non_snake_case)]
 
+#[cfg(any(feature = "alloc", feature = "std"))]
```
oleganza: Why not guard the entire `use straus;` in the `scalar_mul/mod.rs`, or maybe the whole `use scalar_mul`?

hdevalence: We can't guard the entire `use scalar_mul` since we want to be able to do some scalar muls with no_std, but the guards could probably be cleaner.

src/edwards.rs
```diff
@@ -119,7 +119,10 @@ use scalar_mul::window::LookupTable;
 
 use traits::{Identity, IsIdentity};
 use traits::ValidityCheck;
+
+#[cfg(any(feature = "alloc", feature = "std"))]
 use traits::MultiscalarMul;
+#[cfg(any(feature = "alloc", feature = "std"))]
```
hdevalence: This is just to prevent unused import warnings, or?

PR #168, opened 2018-07-20
impl Default for remaining point types

### LINKED ISSUE #154: Point types should `impl Default`
We can't write a generic `impl<T: Identity> Default for T` because of the orphan rules, but we should provide `Default` impl for every type implementing `Identity`.

isislovecruft:
Closes #154.

### DIFF
--- a/src/montgomery.rs
+++ b/src/montgomery.rs
@@ -155,6 +155,12 @@ impl Identity for ProjectivePoint {
     }
 }
 
+impl Default for ProjectivePoint {
+    fn default() -> ProjectivePoint {
+        ProjectivePoint::identity()
+    }
+}
+
 impl ConditionallyAssignable for ProjectivePoint {
     fn conditional_assign(&mut self, that: &ProjectivePoint, choice: Choice) {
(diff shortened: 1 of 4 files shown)


src/montgomery.rs
```diff
@@ -155,6 +155,12 @@ impl Identity for ProjectivePoint {
     }
 }
 
+impl Default for ProjectivePoint {
+    fn default() -> ProjectivePoint {
+        ProjectivePoint::identity()
+    }
+}
+
```
hdevalence: Should we add this? We don't expose the `ProjectivePoint` externally, and we don't use the `Default` impl internally, so this function is dead code.

isislovecruft:
I'll remove it.

isislovecruft:
Should I also remove the impls for `ProjectiveNielsPoint` and `AffineNielsPoint`?

hdevalence:
Nope, they're used when building lookup tables. I don't have a problem with having a `Default` impl for an internal point type, it's just that since it's not used by the `ProjectivePoint` I would rather not add it unless we needed it for something.

PR #170, opened 2018-07-20
Unify `alloc` and `std` cargo features

tarcieri:
This change provides a common convention for using allocator-dependent features with:

```rust
#![cfg(feature = "alloc")]
```

When available, `Vec` is imported consistently as `prelude::Vec`, which means modules that need access to `Vec` can simply do:

```rust
use prelude::*;
```

...and if an allocator is available, `Vec` will be in the crate prelude.

This allows all `alloc` vs `std` gating to be handled in `lib.rs`, `build.rs`, and `prelude.rs` so the rest of the codebase doesn't have to do any gating whatsoever.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,16 +15,18 @@ env:
   - TEST_COMMAND=test EXTRA_FLAGS='' FEATURES='serde'
   # Tests building without std. We have to select a backend, so we select the one
   # most likely to be useful in an embedded environment.
-  - TEST_COMMAND=build EXTRA_FLAGS=--no-default-features FEATURES='u32_backend'
+  - TEST_COMMAND=build EXTRA_FLAGS='--no-default-features' FEATURES='u32_backend'
+  # Tests no_std+alloc usage using the most embedded-friendly backend
+  - TEST_COMMAND=test EXTRA_FLAGS='--lib --no-default-features' FEATURES='alloc u32_backend'
 
 matrix:
   exclude:
     # Test the avx2 backend only on nightly
     - rust: stable
       env: TEST_COMMAND=test EXTRA_FLAGS='--no-default-features' FEATURES='std avx2_backend'
-    # Test no_std only on nightly.
+    # Test no_std+alloc only on nightly
     - rust: stable
-      env: TEST_COMMAND=build EXTRA_FLAGS=--no-default-features FEATURES='u32_backend'
+      env: TEST_COMMAND=test EXTRA_FLAGS='--lib --no-default-features' FEATURES='alloc u32_backend'
 
 script:
   - cargo $TEST_COMMAND --features="$FEATURES" $EXTRA_FLAGS
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -21,6 +21,12 @@
 //! `32bit` since identifiers can't start with letters, and the backends
 //! do use `u32`/`u64`, so this seems like a least-bad option.
 
+#[cfg(not(any(feature = "u32_backend", feature = "u64_backend", feature = "avx2_backend")))]
+compile_error!(
+    "no curve25519-dalek backend cargo feature enabled! \
+     please enable one of: u32_backend, u64_backend, avx2_backend"
+);
+
 #[cfg(feature = "u32_backend")]
 pub mod u32;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,10 +8,9 @@
 // - Isis Agora Lovecruft <isis@patternsinthevoid.net>
 // - Henry de Valence <hdevalence@hdevalence.ca>
 
-#![cfg_attr(not(feature = "std"), no_std)]
+#![no_std]
 
-#![cfg_attr(feature = "alloc", feature(alloc))]
-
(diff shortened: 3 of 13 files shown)


src/backend/mod.rs
```diff
@@ -21,6 +21,12 @@
 //! `32bit` since identifiers can't start with letters, and the backends
 //! do use `u32`/`u64`, so this seems like a least-bad option.
 
+#[cfg(not(any(feature = "u32_backend", feature = "u64_backend", feature = "avx2_backend")))]
+compile_error!(
+    "no curve25519-dalek backend cargo feature enabled! \
+     please enable one of: u32_backend, u64_backend, avx2_backend"
+);
```
tarcieri: Threw this in too... it results in:

```
   Compiling curve25519-dalek v0.18.0 (file:///Users/bascule/src/curve25519-dalek)
error: no curve25519-dalek backend cargo feature enabled! please enable one of: u32_backend, u64_backend, avx2_backend
  --> src/backend/mod.rs:25:1
   |
25 | / compile_error!(
26 | |     "no curve25519-dalek backend cargo feature enabled! \
27 | |      please enable one of: u32_backend, u64_backend, avx2_backend"
28 | | );
   | |__^

error: aborting due to previous error
```

...if no backend is selected.

src/lib.rs
```diff
@@ -8,10 +8,9 @@
 // - Isis Agora Lovecruft <isis@patternsinthevoid.net>
 // - Henry de Valence <hdevalence@hdevalence.ca>
 
-#![cfg_attr(not(feature = "std"), no_std)]
-
-#![cfg_attr(feature = "alloc", feature(alloc))]
+#![no_std]
```
tarcieri: This provides consistent semantics between `no_std` and `std` and makes the `std` feature purely additive

src/scalar.rs
```diff
@@ -328,7 +328,6 @@ impl Scalar {
     /// # Returns
     ///
     /// A random scalar within ℤ/lℤ.
-    #[cfg(feature = "std")]
```
tarcieri: This works on `no_std` as of rand 0.5

.travis.yml
```diff
@@ -15,7 +15,9 @@ env:
   - TEST_COMMAND=test EXTRA_FLAGS='' FEATURES='serde'
   # Tests building without std. We have to select a backend, so we select the one
   # most likely to be useful in an embedded environment.
-  - TEST_COMMAND=build EXTRA_FLAGS=--no-default-features FEATURES='u32_backend'
+  - TEST_COMMAND=build EXTRA_FLAGS='--no-default-features' FEATURES='u32_backend'
+  # Tests no_std+alloc usage using the most embedded-friendly backend
+  - TEST_COMMAND=test EXTRA_FLAGS='--lib --no-default-features' FEATURES='alloc u32_backend'
```
tarcieri: The test suite passes with the 'alloc' feature, however for some reason `rand` isn't available in the test environment unless `std` is enabled?

src/ristretto.rs
```diff
@@ -937,6 +940,7 @@ impl Debug for RistrettoPoint {
 
 #[cfg(all(test, feature = "stage2_build"))]
 mod test {
+    #[cfg(feature = "rand")]
```
tarcieri: Not sure why, but the `rand` crate isn't available during tests unless `std` is enabled. This gating is necessary to make the tests pass with `alloc` alone.

src/backend/mod.rs
hdevalence: Nice!!

src/lib.rs
```diff
@@ -32,11 +31,14 @@
 // External dependencies:
 //------------------------------------------------------------------------
 
-#[cfg(feature = "std")]
-extern crate core;
-#[cfg(feature = "alloc")]
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+#[macro_use]
```
tarcieri: `vec![]` will be available wherever `alloc` is enabled

src/prelude.rs
```diff
@@ -0,0 +1,8 @@
+//! Crate-local prelude (for alloc-dependent features like `Vec`)
+
+// TODO: switch to alloc::prelude
```
tarcieri: I can go ahead and do this now if you're happy with requiring recent nightlies

.travis.yml
hdevalence: Hmm, I'm not sure what's happening there. Do we want to run tests without `std`? (Can tests be run without `std`?)

src/prelude.rs
hdevalence: I'm happy with merging this PR without that and adding it later.

.travis.yml
tarcieri: `std` is still linked, but this is testing the configuration when the `std` feature is not enabled. It can and does work and pass (aside from the doc tests, hence the `--lib` flag).

src/prelude.rs
tarcieri: Sounds good. The only thing you're using is `Vec` anyway, so no need to pull in the whole prelude.

hdevalence:
Rebased onto `develop` in #174

PR #172, opened 2018-07-20
impl Default for CompressedEdwardsY

isislovecruft:
So this is a little weird, but it's nice syntactically because it lets me do:

```rust
use clear_on_drop::clear::Clear;

#[derive(Default)]
pub struct Keypair {
    secret: Scalar,
    public: CompressedEdwardsY,
}
```

I'm not sure about the `Identity` part though, since a compressed point doesn't really have an identity, it only has a byte sequence that when decompressed would equal the identity point for the Edwards model.

### DIFF
--- a/src/edwards.rs
+++ b/src/edwards.rs
@@ -152,6 +152,12 @@ impl Identity for CompressedEdwardsY {
     }
 }
 
+impl Default for CompressedEdwardsY {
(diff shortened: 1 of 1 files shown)


hdevalence:
Don't these types already implement `Identity` and `Default`?

hdevalence:
I think that these changes were already included in #168, closing.

PR #173, opened 2018-07-23
Revert "Unify `alloc` and `std` cargo features"

hdevalence:
Reverts dalek-cryptography/curve25519-dalek#170


hdevalence:
Merged to `master` instead of `develop`

PR #177, opened 2018-07-25
Fix distribution of curve points for hashing to Ristretto points.

samscott89:
Pretty sure this was just a typo (based on the signature of the function and the code for `random`), so I skipped opening an issue.

Current code basically just computes the same encoding twice instead of sampling from different domains.


hdevalence:
Yikes. Thanks!

hdevalence:
Rebased onto `develop`.

PR #178, opened 2018-07-26
Migrate to packed_simd from core::simd

### LINKED ISSUE #175: AVX2 build broken due to deletion of `core::simd` upstream
The `core::simd` module was deleted last week (discussion: https://github.com/rust-lang-nursery/stdsimd/issues/525), so the AVX2 backend won't build any more.

For now it's possible to fix the build by setting last week's nightly in `rust-toolchain`:
```
echo "nightly-2018-07-19" > rust-toolchain
```

hdevalence:
Closes #175.


hdevalence:
@isislovecruft Since the build is broken on nightly, we maybe want to add a `0.19.0` before the `1.0.0-pre.0` release?

PR #185, opened 2018-08-16
ristretto decoding cleanup

isislovecruft:


src/ristretto.rs
```diff
@@ -245,21 +245,25 @@ impl CompressedRistretto {
         // Step 2.  The rest.  (XXX write comments)
         let one = FieldElement::one();
         let ss = s.square();
-        let yden = &one + &ss; // 1 - a*s^2
-        let ynum = &one - &ss; // 1 + a*s^2
-        let yden_sqr = yden.square();
-        let xden_sqr = &(&(-&constants::EDWARDS_D) * &ynum.square()) - &yden_sqr;
+        let u1 = &one + &ss;      // 1 + as²
```
hdevalence: Since `a = -1`, `1 + as^2` is `1 - s^2`, not `1 + s^2`, so I think this should be `u2` (and the variable named `u2` is actually `u1`.

src/ristretto.rs
```diff
@@ -245,21 +245,25 @@ impl CompressedRistretto {
...
+        let Dy = &I * &(&Dx * &v); // 1/u2
 
-        let mut x = &(&s + &s) * &xden_inv; // 2*s*xden_inv
-        let x_is_negative = x.is_negative();
-        x.conditional_negate(x_is_negative);
-        let y = &ynum * &yden_inv;
+        // x == | 2s/sqrt(v) | == + sqrt(4s²/(ad(1+as²)² - (1-as²)²))
+        let mut x = &(&s + &s) * &Dx;
+        let x_neg = x.is_negative();
+        x.conditional_negate(x_neg);
```
hdevalence: AFAIR the reason to have `x_neg` (instead of `x.conditional_negate(x.is_negative())` was because of borrowck problems (can't mutably borrow x (for cneg) and immutably borrow x (for is_neg) on the same line).

Is this still the case? Is it the case with NLL?

If it goes away with NLL we should add an XXX to remove the variable once NLL lands.

src/ristretto.rs
hdevalence: I think that rather than swapping the roles of `u1` and `u2` in the formulas here: https://github.com/Ristretto/website/pull/24/files, the `u1` and `u2` have been swapped here, because the code is using the fact that `a = -1` to compute `1 \pm as^2` as `1 \mp s^2` (i.e., avoiding a multiplication by `-1`).

src/ristretto.rs
isislovecruft: What's NLL stand for? I tried to remove `x_neg` (hence the variable renaming when I later put it back) but the compiler yelled at me about moving mutable variables.

src/ristretto.rs
isislovecruft: I've now flipped the comments back and added a note that `a=-1`.

src/ristretto.rs
hdevalence: Sorry, non-lexical lifetimes (`#![feature(nll)]`).

I'm not 100% sure, but I think that the problem with 
```
x.conditional_negate(x.is_negative())
```
is that the lexical lifetime rules require that the borrow of `x` as `&self` in `x.is_negative()` live for the whole line, which conflicts with the `&mut self` borrow in `x.conditional_negate()`. But since this isn't a real borrowing conflict, just a limitation of the current borrow checker analysis, maybe it would go away if we enabled NLL in the crate root.

hdevalence:
LGTM

PR #186, opened 2018-08-16
Feature gate byteorder/i128.

### LINKED ISSUE #184: Bump byteorder version?
We may need/want to bump the byteorder version to avoid this problem with `#[feature]`s, pointed out by @ruuda: https://github.com/dalek-cryptography/curve25519-dalek/commit/1e74cb3e56c1b2dba7d72b5bf8ed69c094417e0b#r29958498

isislovecruft:
* CLOSES #184.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ harness = false
 
 [dependencies]
 rand = { version = "0.5", default-features = false }
-byteorder = { version = "1", default-features = false, features = ["i128"] }
+byteorder = { version = "1", default-features = false }
 digest = "0.7"
 generic-array = "0.9"
 clear_on_drop = "=0.2.3"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -61,7 +61,7 @@ serde = { version = "1.0", optional = true }
 packed_simd = { version = "0.1.0", features = ["into_bits"], optional = true }
 
 [features]
-nightly = ["subtle/nightly", "clear_on_drop/nightly"]
+nightly = ["subtle/nightly", "clear_on_drop/nightly", "byteorder/i128"]
```
hdevalence: Why do we want to feature-gate our `u128` support on `nightly`? They're stable now. Why not bump the byteorder version instead?

hdevalence:
Used #189 instead

PR #191, opened 2018-09-05
Bump subtle dependency version to 0.8.0

isislovecruft:


isislovecruft:
Actually nevermind, I'm bumping it to 0.9.0 instead.

PR #193, opened 2018-09-05
Add `impl From<usize> for Scalar`

isislovecruft:


hdevalence:
Why not use `Scalar::from(x as u64)`?

hdevalence:
As a design point, I think it's better for callers to explicitly convert from the platform-dependent `usize` type to a platform-independent type: if there's platform-dependent types being used in a crypto protocol, that's something that should be handled explicitly, not implicitly.

But if we did add this, I think the implementation should be `Scalar::from(x as u64)`, which is simpler and safer.

PR #201, opened 2018-10-02
Update to digest 0.8 and sha2 0.8

tarcieri:
Vicariously updates to `generic-array` 0.12, however this change also removes `generic-array` as a direct dependency, as it can be sourced from the `digest` crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,6 @@ packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 rand = { version = "0.5", default-features = false }
 byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] }
 digest = { version = "0.8", default-features = false }
-generic-array = "0.9"
 clear_on_drop = "=0.2.3"
 subtle = { version = "1", default-features = false }
 serde = { version = "1.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ features = ["nightly"]
 travis-ci = { repository = "dalek-cryptography/curve25519-dalek", branch = "master"}
 
 [dev-dependencies]
-sha2 = "0.7"
+sha2 = { version = "0.8", default-features = false }
 bincode = "1"
 criterion = "0.2"
 
@@ -43,8 +43,7 @@ harness = false
 [dependencies]
 rand = { version = "0.5", default-features = false }
 byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] }
-digest = "0.7"
-generic-array = "0.9"
+digest = { version = "0.8", default-features = false }
 clear_on_drop = "=0.2.3"
 subtle = { version = "1", default-features = false }
 serde = { version = "1.0", optional = true }
@@ -53,7 +52,7 @@ packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 [build-dependencies]
 rand = { version = "0.5", default-features = false }
 byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] }
-digest = "0.7"
+digest = { version = "0.8", default-features = false }
 generic-array = "0.9"
 clear_on_drop = "=0.2.3"
 subtle = { version = "1", default-features = false }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,12 +44,7 @@ extern crate packed_simd;
 extern crate rand;
 extern crate clear_on_drop;
 extern crate byteorder;
-
-// The `Digest` trait is implemented using `generic_array`, so we need it
-// too. Hopefully we can eliminate `generic_array` from `Digest` once const
-// generics land.
-extern crate digest;
-extern crate generic_array;
+pub extern crate digest;
 
 // Used for traits related to constant-time code.
 extern crate subtle;
--- a/src/ristretto.rs
+++ b/src/ristretto.rs
@@ -167,7 +167,7 @@ use core::borrow::Borrow;
 use rand::{Rng, CryptoRng};
 
 use digest::Digest;
-use generic_array::typenum::U64;
+use digest::generic_array::typenum::U64;
 
 use constants;
 use field::FieldElement;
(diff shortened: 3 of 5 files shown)


src/scalar.rs
```diff
@@ -581,12 +581,12 @@ impl Scalar {
...
-    /// h.input(b"nature, they also parallel the most common formula plot: the archetype of");
-    /// h.input(b"murder. Other phantasms were occasionally used to underline the fact that");
-    /// h.input(b"perhaps all architecture, rather than being about functional standards, is");
-    /// h.input(b"about love and death.");
+    /// h.input("To really appreciate architecture, you may even need to commit a murder.");
+    /// h.input("While the programs used for The Manhattan Transcripts are of the most extreme");
+    /// h.input("nature, they also parallel the most common formula plot: the archetype of");
+    /// h.input("murder. Other phantasms were occasionally used to underline the fact that");
+    /// h.input("perhaps all architecture, rather than being about functional standards, is");
+    /// h.input("about love and death.");
```
tarcieri: This was fun. The `input()` method now takes `AsRef<[u8]>` as a parameter, however bytestring literals are `Sized` and `AsRef` is only impl'd up to `[T; 32]`. Switching from bytestrings to regular strings gets us the `AsRef<[u8]>` impl we need. Hax, but it's just an example...

src/ristretto.rs
```diff
@@ -167,7 +167,7 @@ use core::borrow::Borrow;
 use rand::{Rng, CryptoRng};
 
 use digest::Digest;
-use generic_array::typenum::U64;
+use digest::generic_array::typenum::U64;
```
tarcieri: Might as well use `generic_array` from `digest` rather that requiring a version-lockstepped `extern crate`, eh?

src/lib.rs
```diff
@@ -44,12 +44,7 @@ extern crate packed_simd;
 extern crate rand;
 extern crate clear_on_drop;
 extern crate byteorder;
-
-// The `Digest` trait is implemented using `generic_array`, so we need it
-// too. Hopefully we can eliminate `generic_array` from `Digest` once const
-// generics land.
-extern crate digest;
-extern crate generic_array;
+pub extern crate digest;
```
tarcieri: By making this `pub`, `ed25519-dalek` can depend on it rather than having its own `extern digest` declaration, and we can be sure the two are in lockstep version-wise.

src/ristretto.rs
```diff
@@ -870,7 +870,7 @@ define_mul_variants!(LHS = Scalar, RHS = RistrettoPoint, Output = RistrettoPoint
 #[cfg(feature = "alloc")]
 impl MultiscalarMul for RistrettoPoint {
     type Point = RistrettoPoint;
-    
+
```
tarcieri: My editor is set up to trim trailing whitespace on files I modify...

src/scalar.rs
newpavlov: You also can demonstrate the new `chain` method by rewriting it as:
```Rust
let h = Sha256::new()
    .chain("To really appreciate architecture, you may even need to commit a murder.");
    .chain("While the programs used for The Manhattan Transcripts are of the most extreme");
    .chain("nature, they also parallel the most common formula plot: the archetype of");
    .chain("murder. Other phantasms were occasionally used to underline the fact that");
    .chain("perhaps all architecture, rather than being about functional standards, is");
    .chain("about love and death.");
```

Cargo.toml
```diff
@@ -26,7 +26,7 @@ features = ["nightly"]
 travis-ci = { repository = "dalek-cryptography/curve25519-dalek", branch = "master"}
 
 [dev-dependencies]
-sha2 = "0.7"
+sha2 = "0.8"
```
newpavlov: Note that by default hash crates now enable `std` feature. It shouldn't matter for tests, but because `std` features are not used, I think it's worth to add `default-features = false`.

tarcieri: Updated, although note this is just a dev dependency

tarcieri:
Updated with @newpavlov's suggestions of using the new `chain` API in the documentation example, and disabling `default-features` for the `digest` and `sha2` crates, since the `std` feature isn't being used and is now enabled by default

mehcode:
Is anything else needed here? I'm using `ed25519-dalek` (awesome work here by the way) and would love to be able to pull in updated hash crates.

hdevalence:
Sorry, I didn't have time to look at it yet, will check it out now.

Cargo.toml
```diff
@@ -53,7 +52,7 @@ packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 [build-dependencies]
 rand = { version = "0.5", default-features = false }
 byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] }
-digest = "0.7"
+digest = { version = "0.8", default-features = false }
 generic-array = "0.9"
```
hdevalence: Should this `generic-array` dependency also be removed?

tarcieri: Indeed. Removed.

tarcieri: Err, too late!

hdevalence:
Thanks for the PR, and sorry for the delay in getting it merged.  Now that the Ristretto work is mostly done I think we will be putting out a `1.0.0-pre` version with these changes Real Soon(tm).

newpavlov:
Are you sure about it? `Digest` trait is part of your public API, so I believe it's not advisable to go with 1.0 until `digest` will be stabilized as well. Currently I don't plan to make digest v1.0.0 until const generics will get stabilized.

hdevalence:
Yeah, it's unfortunate, but we don't really want to wait for const generics to be stabilized.  While it means that we'll be stuck with `0.8` for `digest`, we can release a `2.0` version when `digest` is stabilized.

tarcieri:
Here's an approach I've considered for 1.0+ crates of mine which depend on pre-1.0 crates:

- Gate `digest`-related functionality behind a cargo feature that calls out the feature is not part of the stable 1.0 API, e.g. `digest-preview` or `digest-unstable`. Make this feature opt-in only (i.e. not enabled by default).
- Provide guidelines around how `digest` upgrades will be handled, e.g. every `digest` bump will require AT LEAST the minor version be bumped.
- When `digest` is stabilized, follow the above rule, but change the feature name (e.g. to just `digest`) and optionally enable it by default.

PR #202, opened 2018-10-10
Implement ZeroSafe for Scalar to allow zeroing arrays of packed scalars

oleganza:
`Scalar` already implements `Default`, allowing clearing of individual values. But to allow clearing `Vec<Scalar>` or `&[Scalar]` we need to mark `Scalar` as zeroable because ClearOnDrop zeroes the entire array in one call.

### DIFF
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -251,8 +251,8 @@ impl Scalar {
 }
 
 use clear_on_drop::clear::ZeroSafe;
-// Scalars already implement Default, allowing clearing of individual scalars.
-// But to allow clearing `Vec<Scalar>` or `&[Scalar]` we need to mark Scalars as zeroable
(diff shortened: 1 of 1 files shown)


oleganza:
A PR in bulletproofs repo depends on this patch: https://github.com/dalek-cryptography/bulletproofs/pull/183

hdevalence:
Is this still required? I'd prefer not to merge it, just because it adds the `clear_on_drop` types to the public API, making it impossible for us to treat the clearing-on-drop as an internal implementation detail.

oleganza:
@hdevalence then lets not do that. Clearing manually turned out to be pretty straightforward.

hdevalence:
It might be better for `curve25519_dalek` to implement `Drop` unconditionally (even for non-secret scalars); the cost of wiping 32 bytes is pretty minimal and it would avoid all of these problems.

PR #215, opened 2018-12-10
Restructure vector backends and add a new IFMA backend

hdevalence:
Implements a new IFMA backend targeting Cannonlake processors with AVX512-IFMA. It is almost as fast as FourQ (with endomorphisms).  As far as I am aware, this is the fastest Curve25519 implementation ever.

image from ios

Rendered notes: https://doc-internal.dalek.rs/develop/curve25519_dalek/backend/vector/ifma/index.html

This changeset also restructures the backends into `serial` and `vector`, with common scalar multiplication implementations for each.


hdevalence:
I can try to rebase these changes to collapse a bunch of the WIP commits and clean up the history.

PR #219, opened 2018-12-26
Support passing both borrowed and owned Rngs

burdges:
Rng was designed to support both &mut and owned usage like this.

It's slightly odd to do this on a method since Borrow exists for that, but this helps a lot with data types so one may as well use it.

### DIFF
--- a/src/ristretto.rs
+++ b/src/ristretto.rs
@@ -621,7 +621,7 @@ impl RistrettoPoint {
     /// discrete log of the output point with respect to any other
     /// point should be unknown.  The map is applied twice and the
     /// results are added, to ensure a uniform distribution.
-    pub fn random<T: Rng + CryptoRng>(rng: &mut T) -> Self {
+    pub fn random<T: Rng + CryptoRng>(mut rng: T) -> Self {
         let mut uniform_bytes = [0u8; 64];
         rng.fill(&mut uniform_bytes);
 


hdevalence:
Neat. I'm not sure if we can make this change, since it alters the type parameters on public API which is covered by a stability promise.

burdges:
I think this cannot break any existing code since `impl<'a,R: Rng> Rng for &'a mut R` exists as part of rand's public API.  I've no idea how this might interact with some future edition that revamped the borrow, etc. traits or whatever though.

newpavlov:
>it alters the type parameters on public API which is covered by a stability promise

IIUC you have published only a pre-release 1.0, so changing API at this stage will not count as breaking stability promise.

hdevalence:
Thanks! This should be included in `1.1.0-pre.0`, published as a prerelease to test that generalizing the trait bound doesn't cause problems.

oleganza:
>I think this cannot break any existing code since `impl<'a,R: Rng> Rng for &'a mut R` exists as part of rand's public API.

Looks like it did break code that was taking in `&mut R` and passing it without double-borrowing into two Scalar::random() calls. 

* https://github.com/interstellar/slingshot/pull/89
* https://github.com/dalek-cryptography/curve25519-dalek/issues/232

burdges:
Interesting.  It appears [`Distribution::sample`](https://docs.rs/rand/0.6.5/rand/distributions/trait.Distribution.html#tymethod.sample) also takes a `&mut R` not an `R: Rng`, so the `rand` crate itself might occasionally trigger similar.

Issue #125, opened 2018-04-05
Multiscalar multiplication with precomputation


hdevalence:
The current multiscalar API looks like:
```rust
pub fn multiscalar_mul<I, J>(scalars: I, points: J) -> RistrettoPoint
where
    I: IntoIterator,
    I::Item: Borrow<Scalar>,
    J: IntoIterator,
    J::Item: Borrow<RistrettoPoint>, 
```
This takes an iterator of scalars and an iterator of points, and computes `Q = c_1 P_1 + \cdots + c_n P_n`. This is super-flexible and very useful (it makes it easy to combine statements), but it doesn't allow precomputation.

To do that, we could create a struct (not sure about naming, let's say `MultiscalarPrecomputation` for now) that looks like:
```rust
struct MultiscalarPrecomputation {
    /// Takes points B_1, ..., B_m
    pub fn new<I>(points: I) -> Self
    where
        I: IntoIterator,
        I::Item: Borrow<Scalar>,
    { ... }

    /// Computes (a_1*A_1 + ... + a_n *A_n) + (b_1*B_1 + ... + b_m * B_n)
    pub fn multiscalar_mul<I,J,K>(
        dynamic_scalars: I, 
        dynamic_points: J,
        static_scalars: K,
    ) -> RistrettoPoint
    where
        I: IntoIterator,
        I::Item: Borrow<Scalar>,
        J: IntoIterator,
        J::Item: Borrow<RistrettoPoint>,
        K: IntoIterator,
        K::Item: Borrow<Scalar>,
    { ... }
}
```
This API would let us cover every combination of static/dynamic points, and because the precomputation is opaque, it means that we can pick what kind of static data we want to use depending on the backend implementation. This would supersede #79.

hdevalence:
I think that a basic version of this would be easy to implement using the existing scalar mul code, but I'm not sure if it's worth doing before `1.0`.

hdevalence:
Some work-in-progress on this is in the https://github.com/dalek-cryptography/curve25519-dalek/tree/feature/multiscalar-traits branch.

This has an additional upshot: by eliminating the bare functions and the `vartime` modules, we would have the option (if we wanted) to expose a flat API namespace.

hdevalence:
Work-in-progress traits for constant-time and variable-time multiscalar multiplication here: https://github.com/dalek-cryptography/curve25519-dalek/commit/8f659c32096af0c5593050e0a496b4683225e663

[...]

PR #220, opened 2018-12-26
Improve points being used in data structures

burdges:
In multi party protocols, there are situations where indexing some data structure with a public key comes in handy, so this adds various `#[derive(..)]` to `CompressedRistretto`.  

In addition, this adds a `RistrettoBoth` both convenience type, which requires extensive boilerplate itself, but makes it easy to keep compressed and uncompressed forms together.  If used properly, I believe `RistrettoBoth` should reduce implementation errors, improve readability, simplify indexing data structure, and maybe facilitate performance improvements like https://github.com/dalek-cryptography/ed25519-dalek/issues/61 

In essence, `RistrettoBoth` acts like a `CompressedRistretto` that drags around `RistrettoPoint`, which you access by calling `.as_point()`.  I therefore did not implement constant time or arithmetic operation traits for it, but..

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "curve25519-dalek"
-version = "0.21.0"
+version = "1.0.0-pre.0"
 authors = ["Isis Lovecruft <isis@patternsinthevoid.net>",
            "Henry de Valence <hdevalence@hdevalence.ca>"]
 readme = "README.md"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ rand = { version = "0.5", default-features = false }
 byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] }
 digest = { version = "0.8", default-features = false }
 clear_on_drop = "=0.2.3"
-subtle = { version = "1", default-features = false }
+subtle = { version = "2", default-features = false }
 serde = { version = "1.0", optional = true }
 packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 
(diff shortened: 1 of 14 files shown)


hdevalence:
Thanks for this PR.

The additional derives for `CompressedRistretto` might be reasonable, and it would be good to have them in a separate PR based off of the current `develop` branch for discussion just for those. (I would like to think about what it means for us to advertise a point as being `Ord`, for instance, before merging it).

 I'm not sure if we want to add the `RistrettoBoth` API. In the next few months, I'm planning to update and merge the code I wrote that provides a precomputation API (#125), which might supersede the precomputation use-case, and I'm not sure it's better to allow API clients to not think about the serialization/validation operation.  For Ed25519, it makes sense to factor the public-key decompression into the `PublicKey` struct, but this might not be the case for other protocols, and in our Bulletproofs implementation, we arrange the code so that we don't need to maintain a pair of points.

burdges:
Cool.  Yeah cosmetic things like this should wait until any relevant.  :) 

Yes, it's true protocol implementors could simply accept `RistrettoBoth` types instead of creating their own public key type, which should often have an error type with deserialization errors, like Ed25519, but the protocol implementor can reason about the deserialization failures better than users.  In principle implementors can abuse `RistrettoPoint` similarly though, so at least `RistrettoBoth` gives a venue to explain this worry.  <shrug>

hdevalence:
It might be a good idea - if it ends up being useful in a bunch of libraries that use curve25519-dalek, I wouldn't be opposed to adding it, I would just like to wait on it.

burdges:
There is a minor mistake in the `PartialEq` and `Eq` definition for `RistrettoBoth` here, probably either they should be derived, and thus compare the point too, or else the fields should be kept private to the module.

burdges:
I increasingly find this "both" approach slightly distasteful.  It might be best to provide a both type but actually use some trait so that deserialization can be delayed until `optional_multiscalar_mul`.  I originally considered using a trait that worked for multiple stages of being decompressed to be too much magic but maybe it's fine.

PR #222, opened 2019-01-05
Replace rand with rand_core+rand_os

newpavlov:
You don't have to depend on the whole `rand`, instead `rand_core` and `rand_os` can be used.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 [features]
 nightly = ["subtle/nightly", "clear_on_drop/nightly"]
 default = ["std", "u64_backend"]
-std = ["alloc", "subtle/std"]
+std = ["alloc", "subtle/std", "rand_core/std"]
 alloc = []
 yolocrypto = []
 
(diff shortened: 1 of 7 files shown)


hdevalence:
What's the benefit of depending on `rand_core+rand_os` over `rand`?

Also, is it possible for us to make this change without breaking compatibility? Since the `rand` trait is part of our public API, I'm not sure we can do this without bumping the major version (which we're planning to do at the next time the `rand` crate is updated, but hopefully not beforehand)...

newpavlov:
`rand_core` and `rand_os` are smaller crates, which, in addition to being more lightweight, will make it easier to review the whole dependency tree of your crates. Also API surface is significantly smaller, so they will have breaking changes much less often.

tarcieri:
> What's the benefit of depending on rand_core+rand_os over rand

See #648. The `rand` crate presently pulls in a small zoo of subcrates including non-cryptogrraphically-secure RNGs as well as “90s crypto” RNGs

hdevalence:
Thanks! This should be included in `1.1.0-pre.0`, published as a prerelease to test that generalizing the trait bound doesn't cause problems.

PR #225, opened 2019-01-22
Fix two typos in FieldElement::invsqrt documentation

isislovecruft:
These appeared to be typos?


hdevalence:
Yeah, for the first change the new version is equal to the old version, but for the second, it's a typo.

PR #226, opened 2019-01-23
Check the equality of `EdwardsPoint`s in the projective coordinates

ebfull:
I stumbled on this sentence at [ristretto.group](https://ristretto.group/details/equality.html):

> Equality testing of points on the Edwards curve requires comparing to affine coordinates, which requires an expensive inversion. However, testing whether two points lie in the same coset can be done in projective coordinates, making it actually easier than equality testing in the original non-quotient group.

which led me to check the code here and notice this could be optimized. :smiley_cat:


hdevalence:
Looks like the build failure is related to https://github.com/rust-random/rand/issues/702

ebfull:
Note that the current implementation of `is_small_order` seems to invoke this to compare against the identity, requiring several inversions.

PR #230, opened 2019-02-12
Multiscalar multiplication with precomputation.

### LINKED ISSUE #125: Multiscalar multiplication with precomputation

hdevalence:
Closes #125.

Adds two traits, `PrecomputedMultiscalarMul` and `VartimePrecomputedMultiscalarMul`, provides implementations for `EdwardsPoints` and `RistrettoPoint`s, and reworks the benchmarks to provide precomputation timings.


hdevalence:
I removed the constant-time precomputation support for the reason mentioned in the commit:

> This doesn't (yet) give any speedup over the non-precomputed multiscalar
multiplication, and it's not clear that it's a good idea to commit to
supporting it in the future.  Removing it means that it's not committed-to as
part of the public API, but the source is still there in the tree if we want to
revisit it later.

hdevalence:
The CI failures are related to issues in the rand crate again, merging this before fixing them.

PR #234, opened 2019-02-18
Elligator: use extended coordinates instead of completed coordinates

jedisct1:
This improves clarity by closely matching the Ristretto specification.


hdevalence:
Hi, thanks for the PR, but I would prefer to leave it as-is: I think it's more clear to be explicit that the formulas are coming from the P1xP1 model.  The spec was changed to remove that reference (in order to not depend on a description of the curve models) but since we already include that concept in the source I think it's better to be explicit.

PR #238, opened 2019-03-16
Tests showing that scalar addition and subtraction don't reduce mod l

str4d:
The [documentation for Scalar](https://doc.dalek.rs/curve25519_dalek/scalar/index.html) claims:

> All arithmetic on Scalars is done modulo ℓ.

This PR introduces tests showing that scalar addition and scalar subtraction do not reduce modulo ℓ (but are otherwise arithmetically correct).


str4d:
I think the resolution here is to clarify in the documentation that arithmetic involving canonical Scalars is guaranteed to result in canonical Scalars, but that arithmetic involving non-canonical Scalars is not.

str4d:
To be fair, "done modulo ℓ" is not technically the same as "produces results modulo ℓ", but the distinction is not obvious (it tripped me up for a bit).

hdevalence:
I think the intent was that scalar operations should output canonical results, so I think that this is a bug.

PR #243, opened 2019-04-22
Change short-circuiting logical ORs to bitwise ORs in Ristretto decoding.

isislovecruft:
This prevents a timing side channel which would allow us to determine which check failed. I can't think of a case where this would create problems in practice, but nonetheless I'd argue we should prevent the leak.

Related to https://github.com/dalek-cryptography/subtle/pull/42

### DIFF
--- a/src/ristretto.rs
+++ b/src/ristretto.rs
@@ -266,7 +266,7 @@ impl CompressedRistretto {
             &s_bytes_check[..].ct_eq(self.as_bytes());
         let s_is_negative = s.is_negative();
 
-        if s_encoding_is_canonical.unwrap_u8() == 0u8 || s_is_negative.unwrap_u8() == 1u8 {
+        if (s_encoding_is_canonical.unwrap_u8() == 0u8) | (s_is_negative.unwrap_u8() == 1u8) {
             return None;
         }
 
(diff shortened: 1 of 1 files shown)


hdevalence:
I'm not sure if this is a good change, because the decoding function doesn't operate on secret data, and can't be constant-time anyways because it returns an `Option`: even if we were to consider unvalidated point data as secret, we would need to prevent data flows from that point data into a branch, but it's impossible to use an `Option` without branching.

src/ristretto.rs
```diff
@@ -296,7 +296,7 @@ impl CompressedRistretto {
         // t == ((1+as²) sqrt(4s²/(ad(1+as²)² - (1-as²)²)))/(1-as²)
         let t = &x * &y;
 
-        if ok.unwrap_u8() == 0u8 || t.is_negative().unwrap_u8() == 1u8 || y.is_zero().unwrap_u8() == 1u8 {
+        if (ok.unwrap_u8() == 0u8) | (t.is_negative().unwrap_u8() == 1u8) | (y.is_zero().unwrap_u8() == 1u8) {
```
str4d: Given that `Choice` implements `BitXor` and `Not`, I think this could be written more simply as:
```suggestion
        if ((!ok | t.is_negative() | y.is_zero()).unwrap_u8() == 1u8) {
```

src/ristretto.rs
```diff
@@ -266,7 +266,7 @@ impl CompressedRistretto {
             &s_bytes_check[..].ct_eq(self.as_bytes());
         let s_is_negative = s.is_negative();
 
-        if s_encoding_is_canonical.unwrap_u8() == 0u8 || s_is_negative.unwrap_u8() == 1u8 {
+        if (s_encoding_is_canonical.unwrap_u8() == 0u8) | (s_is_negative.unwrap_u8() == 1u8) {
```
str4d: Similarly:
```suggestion
        if (!s_encoding_is_canonical | s_is_negative).unwrap_u8() == 1u8 {
```

str4d:
If timing is a concern here, the API should probably also be altered to return a `CtOption<RistrettoPoint>`:
```rust
CtOption::new(s, s_encoding_is_canonical & !s_is_negative).and_then(|s| {
    [...]

    CtOption::new(
        RistrettoPoint(EdwardsPoint{X: x, Y: y, Z: one, T: t}),
        ok & !t.is_negative() & !y.is_zero(),
    )
)
```

hdevalence:
Timing isn't a concern here, because point decompression does not operate on secret data.

PR #247, opened 2019-05-17
Fix nostd feature with serde enabled

xoloki:
This library was previously only nostd compliant when serde was not enabled.  To fix this, serde is no longer an optional dependency, but it is properly handled in both std and alloc modes.
This required handling std and alloc modes separately; previously std mode included alloc.  To separate these modes, it was necessary to expand a number of config checks for alloc mode to also include std.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 nightly = ["subtle/nightly", "clear_on_drop/nightly"]
 default = ["std", "u64_backend"]
 std = ["subtle/std", "rand_core/std", "serde/std"]
-alloc = ["subtle/std", "rand_core/std", "serde/alloc"]
+alloc = ["subtle/alloc", "rand_core/alloc", "serde/alloc"]
 yolocrypto = []
 
 # The u32 backend uses u32s with u64 products.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] 
 digest = { version = "0.8", default-features = false }
 clear_on_drop = "=0.2.3"
 subtle = { version = "2", default-features = false }
-serde = { version = "1.0", default-features = false, optional = true }
+serde = { version = "1.0", default-features = false }
 packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 
 [build-dependencies]
@@ -58,14 +58,14 @@ byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] 
 digest = { version = "0.8", default-features = false }
 clear_on_drop = "=0.2.3"
 subtle = { version = "2", default-features = false }
-serde = { version = "1.0", default-features = false, optional = true }
+serde = { version = "1.0", default-features = false }
 packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 
 [features]
 nightly = ["subtle/nightly", "clear_on_drop/nightly"]
 default = ["std", "u64_backend"]
-std = ["alloc", "subtle/std", "rand_core/std"]
-alloc = []
+std = ["subtle/std", "rand_core/std", "serde/std"]
+alloc = ["rand_core/alloc", "serde/alloc"]
 yolocrypto = []
 
 # The u32 backend uses u32s with u64 products.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] 
 digest = { version = "0.8", default-features = false }
 clear_on_drop = "=0.2.3"
 subtle = { version = "2", default-features = false }
-serde = { version = "1.0", optional = true }
+serde = { version = "1.0", default-features = false }
 packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 
 [build-dependencies]
@@ -58,14 +58,14 @@ byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] 
 digest = { version = "0.8", default-features = false }
 clear_on_drop = "=0.2.3"
 subtle = { version = "2", default-features = false }
-serde = { version = "1.0", optional = true }
+serde = { version = "1.0", default-features = false }
 packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 
 [features]
 nightly = ["subtle/nightly", "clear_on_drop/nightly"]
 default = ["std", "u64_backend"]
-std = ["alloc", "subtle/std", "rand_core/std"]
-alloc = []
+std = ["alloc", "subtle/std", "rand_core/std", "serde/std"]
+alloc = ["serde/alloc"]
 yolocrypto = []
 
 # The u32 backend uses u32s with u64 products.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 nightly = ["subtle/nightly", "clear_on_drop/nightly"]
 default = ["std", "u64_backend"]
 std = ["subtle/std", "rand_core/std", "serde/std"]
-alloc = ["subtle/alloc", "rand_core/alloc", "serde/alloc"]
+alloc = ["rand_core/alloc", "serde/alloc"]
 yolocrypto = []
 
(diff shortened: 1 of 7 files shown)


james-darkfox:
Eh. Please don't make serde required.

```rust
#[cfg(all(feature = "serde", not(any(feature = "std",feature = "alloc"))))]
compile_error!("serde feature requires one of std or alloc.");
```

tarcieri:
This seems way more complex than necessary. What problems does adding `default-features = false` to the `serde` dependency not address?

xoloki:
> This seems way more complex than necessary. What problems does adding `default-features = false` to the `serde` dependency not address?
> 
> Making `serde` a hard dependency seems like a bad idea.

The goal is to make the bulletproofs crate (which uses curve25519 with serde) also be nostd compliant.  So it needs to be possible to use curve25519 in nostd mode while also including serde.  To do this, we need to have std and alloc feature targets in curve25519 which appropriately include serde/std and serde/alloc respectively.  The only way I could see to do this was to always depend on serde in curve25519, and separate the std and alloc features; previously, curve25519's std feature target included the alloc target.  Once that happened, it was then necessary to to expand the cfg feature checks to include both std and alloc targets.

If there's a way to do this that doesn't involve making serde a hard dependency for curve25519, then I'm happy to do that instead.

tarcieri:
Is `curve25519-dalek` actually using any `std` or `alloc`-dependent features of `serde`?

If not, adding `default-features = false` to `[dependencies.serde]` should be sufficient.

isislovecruft:
> Is `curve25519-dalek` actually using any `std` or `alloc`-dependent features of `serde`?

Nope.

> If not, adding `default-features = false` to `[dependencies.serde]` should be sufficient.

Doing this and then testing a `no_std` cross-compile with `export TARGET=thumbv7em-none-eabi; xargo build --no-default-features --features=u64_backend --target $TARGET` errors with:

```
∃!isisⒶwintermute:(develop *$=)~/code/rust/curve25519-dalek ∴ xargo build --no-default-features --target $TARGET
   Compiling rand_core v0.4.0
   Compiling typenum v1.10.0
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv7em-none-eabi` target may not be installed

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: Could not compile `rand_core`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

@tarcieri I think it might have to do with the bug you described [here](https://internals.rust-lang.org/t/tools-team-tell-us-your-sorrows/9657/42?u=bascule)?

isislovecruft:
Update: testing this PR with the above cross-compilation commands still errors with the same text.

tarcieri:
@isislovecruft poked at it a bit and offhand I'm not sure what's happening. I can't get it to avoid linking `std` even with `--no-default-features` and eliminating most of the dev/build dependencies (also added `default-features = false` to `serde` in *both* the `build-dependencies` and `dependencies`.

Perhaps bisecting the last time it worked on `no_std` might help? Unfortunately "what's linking `std`" is (perhaps too) hard to diagnose.

isislovecruft:
@tarcieri Ooof.. do we even know when the last time it did work on `no_std` was? I feel like no one's been paying attention to that since you were looking at running it on HSMs.

isislovecruft:
I guess I can git bisect it.. :cry:

tarcieri:
Per #248 it looks like `no_std` builds were generally broken, most likely as a result of https://github.com/rust-lang/cargo/issues/6549

It appears several `dev-dependencies` were activating the `std` feature of both `byteorder` and `rand_core`.

Unfortunately this applies to `serde` too: `bincode`, `criterion`, `csv`, and `tinytemplate` all appear to activate it.

To fix that temporarily (after #248 gets merged), I'd suggest making an (unpublished) `benches/` crate that contains `criterion`. They can share a cargo workspace (somehow different crates in the same workspace are immune to the bug) and a common Cargo.lock resolution, and things like `cargo bench --all` should continue to work.

caemor:
> ∃!isisⒶwintermute:(develop *$=)~/code/rust/curve25519-dalek ∴ xargo build --no-default-features --target $TARGET

You can just use cargo for this target after adding it with `rustup target add thumbv7em-none-eabi`



Cargo nono shows the following std usages:
```
> cargo nono check --no-default-features

curve25519-dalek: FAILURE
  - Source code contains an explicit `use std::` statement.
   --> /home/caemor/git/tests/curve25519-dalek/src/backend/vector/ifma/edwards.rs

  - Source code contains an explicit `use std::` statement.
   --> /home/caemor/git/tests/curve25519-dalek/src/prelude.rs

byteorder: SUCCESS
rand_core: FAILURE
  - Source code contains an explicit `use std::` statement.
   --> /home/caemor/.cargo/registry/src/github.com-1ecc6299db9ec823/rand_core-0.3.1/src/error.rs

  - Source code contains an explicit `use std::` statement.
   --> /home/caemor/.cargo/registry/src/github.com-1ecc6299db9ec823/rand_core-0.3.1/src/error.rs

clear_on_drop: FAILURE
  - Did not find a #![no_std] attribute or a simple conditional attribute like #[cfg_attr(not(feature = "std"), no_std)] in the crate source. Crate most likely doesn't support no_std without changes.
subtle: SUCCESS
digest: FAILURE
  - Source code contains an explicit `use std::` statement.
   --> /home/caemor/.cargo/registry/src/github.com-1ecc6299db9ec823/digest-0.8.0/src/lib.rs

  - Source code contains an explicit `use std::` statement.
   --> /home/caemor/.cargo/registry/src/github.com-1ecc6299db9ec823/digest-0.8.0/src/dyn_digest.rs

  - Source code contains an explicit `use std::` statement.
   --> /home/caemor/.cargo/registry/src/github.com-1ecc6299db9ec823/digest-0.8.0/src/errors.rs

  - Source code contains an explicit `use std::` statement.
   --> /home/caemor/.cargo/registry/src/github.com-1ecc6299db9ec823/digest-0.8.0/src/lib.rs

```

tarcieri:
This should probably get fixed:

```
  - Source code contains an explicit `use std::` statement.
   --> /home/caemor/git/tests/curve25519-dalek/src/backend/vector/ifma/edwards.rs
```

`clear_on_drop` is a false positive.

Somehow it's missing `byteorder` and `serde`

caemor:
When using `cargo nono` without the `--no-default-features` flag it also found byteorder (as a SUCCESS) but serde is still missing. Most likely since it's optional on the current master branch.

When using it on this pr serde is included: 

```
//For all features
serde: FAILURE
  - Crate supports no_std if "std" feature is deactivated.
    - Caused by feature flag "std" in crate "serde:1.0.91"
      - Caused by feature flag "serde/std" in crate "curve25519-dalek:1.1.4"
        - Caused by feature flag "default" in crate "curve25519-dalek:1.1.4"
          - Caused by implicitly enabled default feature from "curve25519-dalek:1.1.4"

//For no_default_features
serde: SUCCESS
```
So serde seems to be included correctly. Everything else stays the same for this pr.

tarcieri:
> So serde seems to be included correctly.

Except this PR makes `serde` a hard dependency...

caemor:
Thats true. I was only focusing on the no_std part.

tarcieri:
This branch still doesn't link on a `no_std` target:

```
$ git fetch xoloki
remote: Enumerating objects: 38, done.
remote: Counting objects: 100% (38/38), done.
remote: Compressing objects: 100% (23/23), done.
remote: Total 38 (delta 22), reused 27 (delta 15), pack-reused 0
Unpacking objects: 100% (38/38), done.
From https://github.com/xoloki/curve25519-dalek
 * [new branch]      add-compressed-cteq-impls -> xoloki/add-compressed-cteq-impls
 * [new branch]      develop                   -> xoloki/develop
 * [new branch]      master                    -> xoloki/master
 * [new branch]      nostd                     -> xoloki/nostd
$ git co nostd
Branch nostd set up to track remote branch nostd from xoloki.
Switched to a new branch 'nostd'
$ cargo build --release --target thumbv7em-none-eabihf --no-default-features --features=u32_backend
   Compiling serde v1.0.70
   Compiling rand_core v0.4.0
   Compiling byteorder v1.2.3
error[E0463]: can't find crate for `std`
```

hdevalence:
> The goal is to make the bulletproofs crate (which uses curve25519 with serde) also be nostd compliant.

Since the bulletproofs crate is not designed to work with `no_std`, it might be good to create an issue on that repo first to propose & discuss changes before diving into changes to dependencies...

xoloki:
Closing this in favor of a new PR #255, which keeps serde as an optional dependency and just turns off default features for it.

PR #248, opened 2019-05-21
Fix no_std builds: remove byteorder; gate rand_core on std

tarcieri:
It looks like you're encountering this bug:

https://github.com/rust-lang/cargo/issues/1796

There are many dev-dependencies in Cargo.lock which activate the `std` features of both `byteorder` and `rand_core`.

To address `byteorder` I've switched to `to_le_bytes`/`from_le_bytes` which are available on integer primitives since 1.32. To support earlier versions of Rust while using core in 1.32+, we can add an extension trait and impl it on the relevant primitive types. This commit, however, at least demonstrates that things will link with `byteorder` removed.

For `rand_core` I just ended up gating it on `std`. It's similarly impacted by the cargo feature activation bug.

Finally, I added a step to the Travis build matrix which cross-compiles to a `no_std`-only target (`thumbv7`), which should ensure future `no_std` regressions are spotted in CI.


.travis.yml
```diff
@@ -1,9 +1,13 @@
 language: rust
+cache: cargo
```
tarcieri: Thought I'd add this... unless you deliberately have it off it should speed up builds (particularly since I'm adding even more stuff)

hdevalence:
I'm not sure what the better alternative is, but feature-gating `rand_core` on `std` is ... really not great. It means, for instance, that the bulletproofs and x25519-dalek libraries will never work on std, and neither will any other library that uses the recommended APIs.

tarcieri:
I fixed `no_std` + `rand_core` support in 1c7c310 by moving all of the benchmarks to an in-tree `benchmarks/` crate and adding it to a Cargo workspace (strangely Cargo workspaces are immune to this bug for whatever reason).

I had to completely remove `rand` and `rand_os` as `[dev-dependencies]` of `curve25519-dalek` (temporarily disabling the tests that were using them), but after doing so I was able to make `rand_core` an always-on non-optional dependency again, and remove all feature gating around it so it will work in `no_std` environments as well.

caemor:
If it is just about the crosscompiling errors for no_std then just pinning the version to 0.3.0 also fixes this issue for me (while also having included the 2 random functions once more) when crosscompiling for thumbv7m
```
[dependencies]
rand_core = { version = "=0.3.0", default-features = false}

[build-dependencies]
rand_core = { version = "=0.3.0", default-features = false}
```

hdevalence:
I made a test library that uses `rand_core` and `curve25519-dalek` with a `no_std` target: https://github.com/hdevalence/no_std_test_lib

It seems to work fine using the currently published versions of the crates... what exactly are we trying to fix?

tarcieri:
@hdevalence ok, confirm, that works. So the bug only applies to `dev-dependencies` in the project consuming the crate, which was breaking linking if you tried to do a `no_std` build within the `curve25519-dalek` workspace.

Based on that I think we can close this PR out, although there are a few little things in here you might like otherwise (notably `cache: cargo`), and it might be worth making a note about all of this in `lib.rs` and/or the README.md.

PR #250, opened 2019-05-24
Make rand_core an optional dependency

tarcieri:
Currently it's confusing to build curve25519-dalek on `no_std` targets due to the following Cargo bug:

https://github.com/rust-lang/cargo/issues/4866

If the `rand_core/std` feature is activated by any of the `build-dependencies` or `dev-dependencies` of the crate attempting to include `curve25519-dalek` (even vicariously through`ed25519-dalek`) then `no_std` builds will fail to link as `rand_core` will link `std`.

Making `rand_core` an optional dependency improves the ergonomics for `no_std` users by eliminating one of the two problematic crates which `curve25519-dalek` has hard dependencies on (the other is `byteorder`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ harness = false
 # match exactly, since the build.rs uses the crate itself as a library.
 
 [dependencies]
-rand_core = { version = "0.3.0", default-features = false }
+rand_core = { version = "0.3.0", default-features = false, optional = true }
 byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] }
 digest = { version = "0.8", default-features = false }
 clear_on_drop = "=0.2.3"
@@ -53,7 +53,7 @@ serde = { version = "1.0", optional = true }
 packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
 
 [build-dependencies]
-rand_core = { version = "0.3.0", default-features = false }
+rand_core = { version = "0.3.0", default-features = false, optional = true }
 byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] }
 digest = { version = "0.8", default-features = false }
 clear_on_drop = "=0.2.3"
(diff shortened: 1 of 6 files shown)


tarcieri:
Note: I can see how this might be considered a break in semver guarantees which could impact users of `curve25519-dalek` who are using `default-features = false` in order to switch backends.

No worries if you reject this PR for that reason.

hdevalence:
Right now there's a clear and type-safe way for people to use a random number generator to construct random points and scalars, and it works both in `std` and `no_std` cases.  Importing both this crate and `rand_core` and compiling for `no_std` already [just works](https://github.com/hdevalence/no_std_test_lib). Although the `rand` and `rand_core` crates have had a number of issues, this isn't anything problematic about the `rand_core` crate, but a problem with the Cargo workflow for `no_std`.

If we do this change and make the mechanism for generating random points and scalars optional, we encourage users to duplicate their RNG-handling logic (have one codepath for `std` and another codepath for `no_std`), which I think makes the API worse.

However, the change will not actually solve the problem for the user, because the unintuitive Cargo behaviour is a behaviour in Cargo, not local to this crate.  If their crate has any dev-dependencies that use `std` features, those will still cause the client crate to have the same problem, and they will have to know about https://github.com/rust-lang/cargo/issues/4866 to fix it.

So, my feeling is that it would be better to solve this issue not by making changes to `curve25519-dalek`, but by documenting (in the near term) and fixing (in the long term) the `no_std` workflow for Cargo, which is the actual source of the issue here.

tarcieri:
> If we do this change and make the mechanism for generating random points and scalars optional, we encourage users to duplicate their RNG-handling logic (have one codepath for std and another codepath for no_std), which I think makes the API worse.

The way a downstream consumer could support `*::random` for `no_std` and std:

```
curve25519-dalek = { version = "X.Y.Z", default-features = false, features = ["u32_backend", "rand_core"] }
```

and if they actually use `std`-dependent features of dalek:

```
[features]
default = ["std"]
std = ["curve25519-dalek/std"]
```

PR #251, opened 2019-05-26
Quench snake case warning

fabric-and-ink:
Non-functional change. Allow non-snake case in whole file to suppress warning in the following line.

https://github.com/dalek-cryptography/curve25519-dalek/blob/f630041af28e9a405255f98a8a93adca18e4315b/src/traits.rs#L328

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -10,6 +10,8 @@
 
 //! Module for common traits.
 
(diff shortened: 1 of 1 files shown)


hdevalence:
Thanks for the PR!

PR #254, opened 2019-05-29
Fix doc comment typo

fabric-and-ink:
There seems to be a mix-up for the documentation of the return value for the function `sqrt_ratio_i`.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -203,9 +203,9 @@ impl FieldElement {
     /// # Return
     ///
     /// - `(Choice(1), +sqrt(u/v))  ` if `v` is nonzero and `u/v` is square;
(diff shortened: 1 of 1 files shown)


hdevalence:
Hmm, the documentation seems to be correct to me... it's tested here: https://github.com/dalek-cryptography/curve25519-dalek/blob/master/src/field.rs#L355-L392 , did you get a result that didn't match the existing documentation?

fabric-and-ink:
Sorry, it's a misunderstanding on my part.

hdevalence:
No worries! Thanks for reading the docs well enough to actually check them :)

fabric-and-ink:
I think they are wonderful! Thank you for adding the formula rendering.

PR #257, opened 2019-06-04
Use upstream intrinsics

### LINKED ISSUE #256: Remove `simd_ffi` for IFMA backend
The IFMA backend was written before I implemented the IFMA intrinsics for Rust, so it uses `simd_ffi`, but now that the intrinsics are upstream, this should be removable.

hdevalence:
Closes #256.  This is a work-in-progress.


hdevalence:
Tested locally on an i3-8121U (Travis does not have Cannonlake or Icelake CPUs).

PR #258, opened 2019-06-05
Hello Rust 2018 + Zeroize Flag For simple support

goldenMetteyya:
Hello,

I have updated the crate to rust 2018. Also I have seen there is some discussion already with removing the clear_on_drop crate for zeroize. As the scalar type can be used as a secret key in constructions like Schnorr it is important that it support such a feature. During this transition period from clear_on_drop I think that adding zeroize behind a feature tag allowing crate users to explicitly call it when needed upstream is a good place to start and doesn't break anyone else usage. I do think that overtime to consolidate behind one library.

thanks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ exclude = [
     ".travis.yml",
 ]
 build = "build.rs"
+edition = "2018"
 
 [package.metadata.docs.rs]
 # Disabled for now since this is borked; tracking https://github.com/rust-lang/docs.rs/issues/302
@@ -51,6 +52,8 @@ clear_on_drop = "=0.2.3"
 subtle = { version = "2", default-features = false }
 serde = { version = "1.0", default-features = false, optional = true }
 packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
+zeroize = "0.9.1"
+
 
 [build-dependencies]
 rand_core = { version = "0.3.0", default-features = false }
@@ -60,6 +63,8 @@ clear_on_drop = "=0.2.3"
 subtle = { version = "2", default-features = false }
 serde = { version = "1.0", default-features = false, optional = true }
 packed_simd = { version = "0.3.0", features = ["into_bits"], optional = true }
+zeroize = { version = "0.9.1", optional = true }
+
 
 [features]
 nightly = ["subtle/nightly", "clear_on_drop/nightly"]
@@ -67,6 +72,9 @@ default = ["std", "u64_backend"]
 std = ["alloc", "subtle/std", "rand_core/std"]
 alloc = []
 yolocrypto = []
+
+# Clear Types
+zeromem = ["zeroize"]
 
 # The u32 backend uses u32s with u64 products.
 u32_backend = []
--- a/build.rs
+++ b/build.rs
@@ -68,7 +68,7 @@ mod prelude;
 #[path = "src/window.rs"]
 mod window;
 
-use edwards::EdwardsBasepointTable;
+use crate::edwards::EdwardsBasepointTable;
 
 fn main() {
     // Enable the "stage2_build" feature in the main build stage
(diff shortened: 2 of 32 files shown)


tarcieri:
Hi, I'm the author of `zeroize`. I have a local PR to migrate `curve25519-dalek` over to `zeroize`, however I haven't pushed it up yet because I've been working through some ergonomic issues in its API.

In particular it leverages all of the new features I've added recently, including `Zeroizing<Z>` and the custom derive support, which eliminates the need for manual `Drop` impls.

I've talked with @hdevalence about moving it over and we generally agreed it would probably be best after a `zeroize` 1.0 release. If people are interested, I can push up my local PR in the meantime though.

goldenMetteyya:
> Hi, I'm the author of `zeroize`. I have a local PR to migrate `curve25519-dalek` over to `zeroize`, however I haven't pushed it up yet because I've been working through some ergonomic issues in its API.
> 
> In particular it leverages all of the new features I've added recently, including `Zeroizing<Z>` and the custom derive support, which eliminates the need for manual `Drop` impls.
> 
> I've talked with @hdevalence about moving it over and we generally agreed it would probably be best after a `zeroize` 1.0 release. If people are interested, I can push up my local PR in the meantime though.

Hi @tarcieri , thanks for your work on zeroize! I think that if that is the case please do share, Besides a stable api is there any features that you want to add that will make it 1.0 ready? 

Even if you derive Drop scalar has a copy trait in its derive that will continue to be an issue no ?

I like to be explicit with these things so I wrote the short zeroize trait implementation so its very clear.

thanks

tarcieri:
> Besides a stable api is there any features that you want to add that will make it 1.0 ready?

I've mostly been working through some ergonomic and design issues around the procedural macro, which it turns out is quite important for zeroizing nontrivial structs.

> Even if you derive Drop scalar has a copy trait in its derive that will continue to be an issue no ?

The current proc macro in `0.9` uses an attribute to indicate you wish to derive a `Drop` impl:

```
#[derive(Zeroize)]
#[zeroize(drop)]
struct MyStruct([u8; 32]);
```

This allows `Zeroize` to still be derived on `Copy` types without receiving a `Drop` impl.

> I like to be explicit with these things so I wrote the short zeroize trait implementation so its very clear.

The problem is it's imperative rather than declarative: if additional fields are ever added to any struct you write a manual `Drop` impl for, they will not be automatically zeroized.

Though proc macros may seem a bit magical, they're based on `synstructure` which has a nice testing story:

https://github.com/iqlusioninc/crates/blob/develop/zeroize_derive/src/lib.rs#L106

goldenMetteyya:
I agree that the way you are developing the library for generic use cases will make it more useful and easier to consume. But here the case is a scalar and I dont think more fields will be added to it. Furthermore the sensitivity of this scalar is pretty high if it is used to build a secret key so makes life easier as you know exactly what's going on from my view. 

I think that having the scalar type have the zeroize implementation behind a feature flag for now until you reach 1.0 won't affect others usage of the crate but also begin the availability for upstream to use the main crate instead of branches and not be on the main release.

goldenMetteyya:
@tarcieri Hey so I have tested it in another place which is a symmetric key https://github.com/stichtingorganism/mohan/blob/master/src/secretbox/mod.rs#L44

and it is very clean and intuitive to use.

hdevalence:
The `zeroize` crate is designed around a public trait, `Zeroize`, which must be implemented by zeroizable types.  This means that `curve25519-dalek` cannot use it internally (unlike `clear_on_drop`): as soon as we release a version that supports `Zeroize`, we are stuck with it until the next major version (= forever).  For this reason my preference would be to wait a while until we're sure that the `zeroize` crate is stable.

tarcieri:
I think it might be helpful if I opened a PR where we can work through the API warts in `zeroize` until we're happy with what it looks like, and after that I can cut a 1.0.

I've also been waiting on const generics, but I'd like to feature gate those and support older Rust versions (by maintaining the macros that presently define the trait on arrays up to `[Z; 64]`).

goldenMetteyya:
hey @tarcieri how far is zeroize from reaching 1.0?

tarcieri:
I'd like to release v0.10 fairly soon which may or may not be the last release before v1.0, but hopefully close

burdges:
Just fyi, if you want `Zeroize` in your own crate now, then you can easily add zeroize without zeroize_derive using
```
#[inline(always)]
fn zeroize_hack<Z: Default>(z: &mut Z) {
    use core::{ptr, sync::atomic};
    unsafe { ptr::write_volatile(z, Z::default()); }
    atomic::compiler_fence(atomic::Ordering::SeqCst);
}
```

goldenMetteyya:
Im Closing this as I see @tarcieri has another PR as the author of zeroize.

PR #259, opened 2019-06-06
Pippenger bugfix

hdevalence:
The Pippenger implementation in 1.2.0 (yanked) had a bug, diagnosed by @oleganza; at the largest problem sizes (using w=8), the signed digits fill the value range of an i8, and so doing computation on them to calculate the bucket index could hit an overflow.

The CI missed catching this because the test suite didn't check all the problem sizes, which I fixed in this PR.  Adding a check that the multiscalar functions work on manually-constructed extremal scalar values (which are allowed by the `Scalar` type for X/Ed25519 compatibility but aren't used as inputs to the multiscalar functions) revealed a bug in the NAF-using computation, which has no security content but has been present since the first version of the library (cc @isislovecruft); there's more details in the commit messages.

### DIFF
--- a/src/backend/serial/scalar_mul/pippenger.rs
+++ b/src/backend/serial/scalar_mul/pippenger.rs
@@ -125,7 +125,8 @@ impl VartimeMultiscalarMul for Pippenger {
             // Note: if we add support for precomputed lookup tables,
             // we'll be adding/subtracting point premultiplied by `digits[i]` to buckets[0].
             for (digits, pt) in scalars_points.iter() {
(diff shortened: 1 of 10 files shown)


oleganza:
No, my mistake. In the common case, the last additional digit (w=8) is zero, which adds no point operations overhead (point is not added to any bucket).

hdevalence:
Yes, exactly -- in the Pippenger implementation, the loop iterates over the scalar digits and takes an action if the digits are nonzero.  So the cost is maybe one or two missed branches or something.

PR #260, opened 2019-06-07
Remove unneeded check for negativity in edwards point decompression

fabric-and-ink:
The function `FieldElement::sqrt_ratio_i` always returns a positive root
by definition. Therefore the test for negativity in the edwards point
decompression function always returns false and we only need to flip its
sign if `compressed_sign_bit` is set.

### DIFF
--- a/src/edwards.rs
+++ b/src/edwards.rs
@@ -192,9 +192,7 @@ impl CompressedEdwardsY {
 
         // Flip the sign of X if it's not correct
         let compressed_sign_bit = Choice::from(self.as_bytes()[31] >> 7);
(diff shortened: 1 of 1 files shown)


src/edwards.rs
```diff
@@ -192,9 +192,7 @@ impl CompressedEdwardsY {
 
         // Flip the sign of X if it's not correct
```
hdevalence:
```suggestion
        // FieldElement::sqrt_ratio_i always returns the nonnegative square root,
        // so we negate according to the supplied sign bit.
```

hdevalence:
Thanks!

PR #262, opened 2019-06-20
Update `rand` dependency to `rand_os`, update `rand_core`

zone117x:
Remove dependency on monolithic `rand` lib.


hdevalence:
Hi there, thanks for the PR! Unfortunately I don't think we can merge this, because it changes the `rand_core` version, and the `rand_core` traits are part of this crate's public API (and so they're covered by this crate's stability guarantees).

We could update the rand versions in a version 2.0 of this crate, but there's a lot of churn in the rand crates so I would rather wait until they release a 1.0 version.

PR #271, opened 2019-08-04
Remove byteorder-dependency

lukaslueg:
This removes the `byteorder`-dependency, which is not needed since 1.32 (raising the minimum Rust version to that). Dropping a dependency is worth the change IMHO.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,6 @@ harness = false
 
 [dependencies]
 rand_core = { version = "0.3.0", default-features = false }
-byteorder = { version = "^1.2.3", default-features = false, features = ["i128"] }
 digest = { version = "0.8", default-features = false }
 clear_on_drop = "=0.2.3"
 subtle = { version = "2", default-features = false }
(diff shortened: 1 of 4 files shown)


tarcieri:
I tried to do this in an earlier PR, but also included a shim/polyfill for the functionality in `byteorder` which was actually being used so code could be written with the new API and the shim/polyfill removed at a later date:

https://github.com/dalek-cryptography/curve25519-dalek/pull/248

Otherwise I think they don't plan on bumping MSRV until 2.0. That said, it seems there are a lot of compelling reasons to release a 2.0 already.

hdevalence:
I'm not sure I feel super strongly that removing `byteorder` is worthwhile.  The replacement is less readable (it does not signal intent) and requires maintaining a custom macro (or removing it and making the code even less readable).  I am not really concerned with MSRV (I'm not sure that this crate actually has one as a policy).

On the other hand I don't think that there's a particular downside to depending on the `byteorder` crate: it's stable, it's a core part of the Rust ecosystem, it does exactly one thing and does it well. So, thanks for the PR, but I would prefer to close this issue for now.

PR #279, opened 2019-08-14
JSON serialization for Scalar / Ristretto / Edwards types

oleganza:
## Problem

Current version of the library encodes scalars and points as serde type "byte array". In `serde_json` this gets encoded into array of numbers. During deserialization this becomes a "sequence" serde type (aka "vector of json values"), which is not provided when the dalek deserializer asks for "byte array", and causes a deserialization failure.

Another issue with JSON is that `HashMap<CompressedRistretto,...>` currently attempts to serialize with array as a key, while JSON only permits strings as keys. This is also fixed with this patch.

## Suggestion

To fix this problem I suggest having built-in serialization of binary data as hex string for "human-readable" serializers. This means that users of binary serializers such as bincode, XDR, protobufs etc could still use efficient binary representation (which is also backwards compatible). But the users of the human-readable serializers such as JSON, YAML, TOML, XML do not have to worry if their format specifies "binary string" type (usually they don't).

## Issues

Seems like some binary serializers do not set `is_human_readable` flag to `false`, which is defaulted to `true` ¯\\(ツ)/¯. Some people already filed issues, but it might be impossible to fix this w/o breaking compatibility: https://github.com/3Hren/msgpack-rust/issues/176

Another problem: if someone is successfully using a serializer marked as human-readable (irrespective of whether it's actually true), this patch will alter encoding for them, making binary strings encoded as hex-encoded utf-8 strings. **This may be a blocker for this PR.**

## Alternatives

**1st alternative** is to enable hex-encoding in this crate as a feature-flag, separate from `"serde"` feature.

**2nd alternative** is to have an alternative version of the `serde_json` crate (or an opt-in knob) that does not support schemaless encoding, and explicitly converts byte arrays to hex (or base64) strings and back. Here's PR on `serde_json` introducing optional hex serialization: https://github.com/serde-rs/json/pull/557

**3rd alternative** is to use a serialization proxy such as https://github.com/slowli/hex-buffer-serde, but in a several layers of abstraction it's not always clear at which layer such proxy should be applied: if it's too low, it'll be having same issues as this PR; but if it's too high, it might be hard to penetrate through several nested types to encode specific item in hex.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
   # Tests the avx2 backend
   - TEST_COMMAND=test EXTRA_FLAGS='--no-default-features' FEATURES='std avx2_backend'
   # Tests serde support and default feature selection
-  - TEST_COMMAND=test EXTRA_FLAGS='' FEATURES='serde serde_json'
+  - TEST_COMMAND=test EXTRA_FLAGS='' FEATURES='serde'
   # Tests building without std. We have to select a backend, so we select the one
   # most likely to be useful in an embedded environment.
   - TEST_COMMAND=build EXTRA_FLAGS='--no-default-features' FEATURES='u32_backend'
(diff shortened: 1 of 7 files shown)


oleganza:
Here's PR for the alternative #2: https://github.com/serde-rs/json/pull/557

hdevalence:
This is fixed by #297 and released in `2.0.0-alpha.0`

PR #285, opened 2019-09-19
try to fix panic at CompressedRistretto from_slice

HaoXuan40404:
Hello, I found that the function "from_silce"  in src/ristretto.rs will happen a panic when receiving a slice of incorrect length. Maybe this unreasonable. And then I try to fix it.

Test main.rs like this
```rust
extern crate curve25519_dalek;
use curve25519_dalek::ristretto;
use curve25519_dalek::scalar::Scalar;

fn main() {
    let test_u8 = [1, 2, 3, 4];
    let test_point = ristretto::CompressedRistretto::from_slice(&test_u8);
    println!("test_point = {:?}", test_point);
}
```

Then I got this:
```rust
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `32`,
 right: `0`: destination and source slices have different lengths', src/libcore/slice/mod.rs:2119:9
stack backtrace:
```

### DIFF
--- a/src/ristretto.rs
+++ b/src/ristretto.rs
@@ -235,9 +235,13 @@ impl CompressedRistretto {
     /// If the input `bytes` slice does not have a length of 32.
     pub fn from_slice(bytes: &[u8]) -> CompressedRistretto {
         let mut tmp = [0u8; 32];
-
-        tmp.copy_from_slice(bytes);
-
+        if bytes.len() != 32 {
+            let byte_tmp =&[0u8; 32];
+            tmp.copy_from_slice(byte_tmp);
+        }
+        else {
+            tmp.copy_from_slice(bytes);
+        }
         CompressedRistretto(tmp)
     }
 


tarcieri:
I think the correct solution to avoiding a panic here is to have a fallible API, not zero-padding. Your “solution” is still panicking because `copy_from_slice` requires that the two slice lengths are equal, and you’re not slicing the temporary buffer. But even if you did, this approach can’t handle an overlength string, nor does zero padding actually help in this context.

HaoXuan40404:
So maybe return ”result” would be better?

HaoXuan40404:
Ok,  so this fallible API is in expected, right?

tarcieri:
Yes, although it’d be a breaking API change.

If @hdevalence is ok with an MSRV bump, you could add a new fallible API by adding a `TryFrom` impl perhaps.

HaoXuan40404:
Ok, thank you

PR #289, opened 2019-09-30
Switch from `clear_on_drop` to `zeroize` (fixes #281)

### LINKED ISSUE #281: compiling to wasm fails due to clear_on_drop without no_cc ?
I'm experimenting with using [magic-wormhole.rs](https://github.com/warner/magic-wormhole.rs/) in a browser context, for which I'm compiling it to WASM (https://github.com/warner/magic-wormhole.rs/issues/2). This is my first attempt at doing anything with WASM, so please tell me if I'm missing something obvious, but I'm seeing the `wasm-pack build` fail because `clear_on_drop` is trying to compile a C file, and I guess that doesn't work when the target is WASM. The dependency chain is `magic-wormhole` -> `spake2` -> `curve25519-dalek` -> `clear_on_drop`.

It looks like `clear_on_drop` has a feature (`no_cc`) that would disable the C compilation, which might allow this to work. But there's no way for `magic-wormhole` to ask `curve25519-dalek` to ask `clear_on_drop` to use that feature.

So I *think* what I'm asking for is some `curve25519-dalek` feature that would then modify its `clear_on_drop` dependency's features to avoid this C compile. Then I could add a similar feature to `spake2` that affects its dependency upon `curve25519-dalek`, and then again with `magic-wormhole` affecting its dependency upon `spake2` to pass the request along.

Does that make sense? It sounds like a lot of work spread over many crates, and I have to imagine that every "feature" added to a Cargo.toml is a long-lived maintenance burden, so I don't really want to make a high-cost feature request for a fringe use case. And clear-on-drop is obviously the right thing to be doing, I don't want to see it become harder to achieve that security property. Is there a better way that people deal with this cross-crate feature handling?

For magic-wormhole's own clear-the-old-secrets purposes, I'm considering the [`zeroize`](https://github.com/iqlusioninc/crates/tree/develop/zeroize) crate, which I think is pure-Rust. But I don't really know what the tradeoffs are; I have to imagine `clear_on_drop` is more reliable or more thorough when it can use lower-level tools.

tarcieri:
`zeroize` is WASM-friendly as it has no dependencies on C compilers.

Instead uses Rust's own volatile write semantics and compiler fences to ensure zeroization is not elided by the compiler.

### DIFF
--- a/src/backend/serial/curve_models/mod.rs
+++ b/src/backend/serial/curve_models/mod.rs
@@ -128,6 +128,8 @@ use core::ops::{Add, Neg, Sub};
 use subtle::Choice;
 use subtle::ConditionallySelectable;
 
+use zeroize::Zeroize;
+
 use constants;
 
 use edwards::EdwardsPoint;
@@ -182,6 +184,14 @@ pub struct AffineNielsPoint {
     pub xy2d:      FieldElement,
 }
 
+impl Zeroize for AffineNielsPoint {
+    fn zeroize(&mut self) {
+        self.y_plus_x.zeroize();
+        self.y_minus_x.zeroize();
+        self.xy2d.zeroize();
+    }
+}
+
 /// A pre-computed point on the \\( \mathbb P\^3 \\) model for the
 /// curve, represented as \\((Y+X, Y-X, Z, 2dXY)\\) in "Niels coordinates".
 ///
@@ -193,6 +203,15 @@ pub struct ProjectiveNielsPoint {
     pub Y_minus_X: FieldElement,
     pub Z:         FieldElement,
     pub T2d:       FieldElement,
+}
+
+impl Zeroize for ProjectiveNielsPoint {
+    fn zeroize(&mut self) {
+        self.Y_plus_X.zeroize();
+        self.Y_minus_X.zeroize();
+        self.Z.zeroize();
+        self.T2d.zeroize();
+    }
 }
 
 // ------------------------------------------------------------------------
--- a/src/backend/serial/scalar_mul/straus.rs
+++ b/src/backend/serial/scalar_mul/straus.rs
@@ -106,7 +106,7 @@ impl MultiscalarMul for Straus {
         J: IntoIterator,
         J::Item: Borrow<EdwardsPoint>,
     {
-        use clear_on_drop::ClearOnDrop;
+        use zeroize::Zeroizing;
 
         use backend::serial::curve_models::ProjectiveNielsPoint;
         use window::LookupTable;
@@ -119,12 +119,12 @@ impl MultiscalarMul for Straus {
 
         // This puts the scalar digits into a heap-allocated Vec.
         // To ensure that these are erased, pass ownership of the Vec into a
-        // ClearOnDrop wrapper.
+        // Zeroizing wrapper.
         let scalar_digits_vec: Vec<_> = scalars
             .into_iter()
             .map(|s| s.borrow().to_radix_16())
             .collect();
-        let scalar_digits = ClearOnDrop::new(scalar_digits_vec);
+        let scalar_digits = Zeroizing::new(scalar_digits_vec);
 
         let mut Q = EdwardsPoint::identity();
         for j in (0..64).rev() {
@@ -137,6 +137,7 @@ impl MultiscalarMul for Straus {
                 Q = (&Q + &R_i).to_extended();
             }
         }
+
         Q
     }
 }
--- a/src/backend/serial/u32/scalar.rs
+++ b/src/backend/serial/u32/scalar.rs
@@ -13,6 +13,8 @@
 use core::fmt::Debug;
 use core::ops::{Index, IndexMut};
 
+use zeroize::Zeroize;
+
 use constants;
 
 /// The `Scalar29` struct represents an element in ℤ/lℤ as 9 29-bit limbs
@@ -22,6 +24,12 @@ pub struct Scalar29(pub [u32; 9]);
 impl Debug for Scalar29 {
     fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
         write!(f, "Scalar29: {:?}", &self.0[..])
+    }
+}
+
+impl Zeroize for Scalar29 {
+    fn zeroize(&mut self) {
+        self.0.zeroize();
     }
 }
 
--- a/src/window.rs
+++ b/src/window.rs
@@ -25,6 +25,8 @@ use edwards::EdwardsPoint;
 use backend::serial::curve_models::ProjectiveNielsPoint;
 use backend::serial::curve_models::AffineNielsPoint;
 
+use zeroize::Zeroize;
+
 /// A lookup table of precomputed multiples of a point \\(P\\), used to
 /// compute \\( xP \\) for \\( -8 \leq x \leq 8 \\).
 ///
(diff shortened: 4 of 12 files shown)


tarcieri:
Note: This uses `zeroize` v1.0.0-pre. I just wanted to sanity check if there are any changes curve25519-dalek would benefit from before cutting a v1.0.0.

src/backend/serial/curve_models/mod.rs
```diff
@@ -182,6 +184,14 @@ pub struct AffineNielsPoint {
     pub xy2d:      FieldElement,
 }
 
+impl Zeroize for AffineNielsPoint {
+    fn zeroize(&mut self) {
+        self.y_plus_x.zeroize();
+        self.y_minus_x.zeroize();
+        self.xy2d.zeroize();
```
tarcieri: Alternatively this could be implemented using a proc macro with `zeroize_derive`, if you'd prefer.

src/backend/serial/curve_models/mod.rs
```diff
@@ -195,6 +205,15 @@ pub struct ProjectiveNielsPoint {
     pub T2d:       FieldElement,
 }
 
+impl Zeroize for ProjectiveNielsPoint {
+    fn zeroize(&mut self) {
+        self.Y_plus_X.zeroize();
+        self.Y_minus_X.zeroize();
+        self.Z.zeroize();
+        self.T2d.zeroize();
```
tarcieri: This could also potentially use `zeroize_derive`.

src/backend/serial/scalar_mul/straus.rs
```diff
@@ -118,15 +118,14 @@ impl MultiscalarMul for Straus {
...
         // This puts the scalar digits into a heap-allocated Vec.
-        // To ensure that these are erased, pass ownership of the Vec into a
-        // ClearOnDrop wrapper.
-        let scalar_digits_vec: Vec<_> = scalars
+        // to be erased below with `Zeroize`.
+        let mut scalar_digits: Vec<_> = scalars
             .into_iter()
             .map(|s| s.borrow().to_radix_16())
             .collect();
-        let scalar_digits = ClearOnDrop::new(scalar_digits_vec);
```
tarcieri: I wasn't able to use `Zeroizing` here because of the current `Zeroize` trait bounds for `Vec`, which are `Clone` + `Default` + `Zeroize`, and `Default` isn't impl'd for `[T: Default; N]` where `N > 32`, which is unfortunate.

src/backend/serial/u32/scalar.rs
```diff
@@ -13,12 +13,16 @@
...
+use zeroize::DefaultIsZeroes;
+
 use constants;
 
 /// The `Scalar29` struct represents an element in ℤ/lℤ as 9 29-bit limbs
-#[derive(Copy,Clone)]
+#[derive(Copy, Clone, Default)]
 pub struct Scalar29(pub [u32; 9]);
 
+impl DefaultIsZeroes for Scalar29 {}
```
tarcieri: This could perhaps be an explicit `Zeroize` impl instead of adding a `Default` bound.

src/window.rs
```diff
@@ -38,28 +40,11 @@ use backend::serial::curve_models::AffineNielsPoint;
...
-///
-/// Since the `ZeroSafe` trait is only used by `clear_on_drop`, the only
-/// situation where this would be a problem is if code attempted to use
-/// a `ClearOnDrop` to erase a `LookupTable` and then used the table
-/// afterwards.
-///
-/// Normally this is not a problem, since the table's storage is usually
-/// dropped too.
-///
-/// XXX is this a good compromise?
```
tarcieri: Wheee...

goldenMetteyya:
Hello @tarcieri when do you think you can cut a 1.0 for zeroize ?

tarcieri:
@goldenMetteyya mostly I've been trying to look at what isn't working in real-world use cases such as this one. The main thing that bothers me right now is the current bounds for `Vec` and how they are preventing the use of `Zeroizing` in several places in this PR.

I think I'm going to relax them. There are a few other small issues but other than that I'd say sometime before the end of October is my goal for 1.0.

goldenMetteyya:
With vectors if your allocating secret variables and if this isn't fixed you still can't know the memory allocator didn't use the same continuation so maybe a special zeroizing native supported collection data structure can be written at some point.

tarcieri:
@goldenMetteyya trying to zero out the rest of the existing capacity safely is the reason for the current bounds, but I don't really like it and I think the best option is to avoid reallocation to begin with.

> maybe a special zeroizing native supported collection data structure can be written at some point.

There are a few types like this in the `secrecy` crate, such as `SecretVec` and `SecretBytes`:

https://docs.rs/secrecy/0.4.0/secrecy/type.SecretVec.html

src/backend/serial/scalar_mul/straus.rs
tarcieri: Update: fixed in `zeroize` 1.0

tarcieri:
Updated this PR to use `zeroize` 1.0 final.

goldenMetteyya:
Hey @tarcieri  so I was trying todo some testing and I merged some of the new changes including this commit and also the new commit for removing the build.rs dep. 

https://github.com/stichtingorganism/curve25519-dalek/tree/organism 

Its giving some errors regarding missing iterators for types wrapped in

zeroize::Zeroizing<std::vec::Vec<backend::serial::u64::scalar::Scalar52>>

for example.

dignifiedquire:
any chance this shipping soon? I am having trouble with arm builds because of clear_on_drop

goldenMetteyya:
> any chance this shipping soon? I am having trouble with arm builds because of clear_on_drop

hey @dignifiedquire https://crates.io/crates/curve25519-dalek-organism is temporary fork I have that has some of the recent changes. Only downside is you need to set the features your self. the default is simd & u64 but you can easily override. when upstream merges to 2.0 should be automatic. but try it out if you like

goldenMetteyya:
@dignifiedquire  its at version 0.0.5

curve25519-dalek-organism = "0.0.5"

hdevalence:
@dignifiedquire I'm working on it now, but in the meantime I believe that the issues with arm builds can be solved by either: a) using the `nightly` feature of curve25519-dalek, which uses a `nightly` feature of `clear_on_drop`, or b) installing an arm toolchain for the system C compiler used to build C dependencies of Rust code.

hdevalence:
@dignifiedquire see https://github.com/dalek-cryptography/curve25519-dalek/issues/281#issuecomment-527566767

hdevalence:
Rebasing....

hdevalence:
@tarcieri I fixed some conflicts in rebasing and pushed to `tarcieri-zeroize` on this repo -- can you update the PR with the contents of that branch? I'll delete it later, but I don't think I can force-push to your fork.

tarcieri:
@hdevalence done

src/window.rs
hdevalence: nice

src/window.rs
```diff
@@ -38,28 +40,11 @@ use backend::serial::curve_models::AffineNielsPoint;
...
-unsafe impl<T> ZeroSafe for LookupTable<T> {}
+pub struct LookupTable<T: Zeroize>(pub(crate) [T; 8]);
```
hdevalence: Do we need to require that `T: Zeroize` in general? (see below)

src/window.rs
```diff
@@ -120,6 +111,15 @@ impl<'a> From<&'a EdwardsPoint> for LookupTable<AffineNielsPoint> {
     }
 }
 
+impl<T> Zeroize for LookupTable<T>
+where
+    T: Copy + Default + Zeroize
```
hdevalence: Could this just be
```
impl<T: Zeroize> Zeroize for LookupTable<T>
```
so that only the `Zeroize` impl required the table entries to be zeroized?  Or are the `Copy + Default` bounds used somewhere in the array impl?

If only this impl had a `Zeroize` bound, then it seems like `LookupTable<T>` would not need `T: Zeroize` and we could skip adding new bounds to some of the where clauses above.

tarcieri: Yep, you're right, PR updated

hdevalence: Great, happy to merge on green CI

Issue #265, opened 2019-07-12
Points and scalars are serialized as 40 bytes (instead of 32)


trevp:
RistrettoPoint, CompressedRistretto, and Scalar are serialized into 40 bytes (instead of 32) when using serde with efficient encodings like bincode.

This is because they are handled as variable-length "bytes" in serde terminology, instead of fixed-length "tuples", so are given a 8-byte length field.

This is fairly easy to fix using serialize_tuple / deserialize_tuple (I've done so in a local branch; could submit a PR), but I don't know what fixes are possible given stability guarantees.

hdevalence:
I'm not sure that we can fix this without breaking API stability, but I recall that the overhead is reduced when using other encodings (like CBOR, which I think uses as little as 1 byte).

I think that some amount of overhead is inevitable when using Serde, because Serde operates on self-describing data structures, not ones specified in advance.  If the data format is specified elsewhere, it should be possible to use the to/from byte conversion methods to achieve a precise and compact encoding for a more complex object like a proof or a signature (we do this in Bulletproofs). This aggregate can then implement the Serde traits, using its own encoding internally, instead of using a derived encoding of all of its components that may not be as efficient.

trevp:
> I think that some amount of overhead is inevitable when using Serde, because Serde operates on self-describing data structures, not ones specified in advance.

I don't think that's true, Serde works great with non-self-describing formats like bincode.

https://serde.rs/impl-deserialize.html

Indeed, if points and scalars serialize to 32 bytes it's easy to have structs consisting of points and scalar and use bincode to #derive efficient, packed representations of signatures, proofs, etc.

To me, it seems better to serialize crypto objects like (keys, signatures, proofs) into fixed-length, non-self-describing, efficiently-packed blobs.

These blobs can then be embedded into (JSON, Protobufs, XML, etc), keeping a separation between your crypto format and code and your application protocol.  Serializing crypto directly into self-describing formats is usually a worse idea (think ASN-1 encoded DSA sigs, or XML-DSIG).

> If the data format is specified elsewhere, it should be possible to use the to/from byte conversion methods to achieve a precise and compact encoding for a more complex object like a proof or a signature (we do this in Bulletproofs).

Yes, one could write serialize/deserialize code by hand, but that defeats the point of serde.  If points and scalars serialize to 32 bytes, then you just #derive these implementations instead of writing them.

tarcieri:
[...]

PR #291, opened 2019-10-10
Update to latest Rand Core

goldenMetteyya:


hdevalence:
Hi there, because this is a breaking change, it can't be merged without a `2.0.0` release, but as per https://github.com/dalek-cryptography/curve25519-dalek/issues/265#issuecomment-536132620 , the Serde support was implemented incorrectly and fixing it will require a breaking change, so this can be included then too.  However, it may take some time to do this, because it's blocked on some other work, but I would like to merge this eventually.

PR #295, opened 2019-10-17
Cleanup for v2

goldenMetteyya:
Hello,

I have merged some recent changes into one PR to help on the path to v2.0.0-pre

- rand crate updates
- replace clear_on_drop with Zeroize 
- Rust 2018
- Remove Build.rs 

I think whats missing is adding a build.rs that has the feature selection and also making sure serde is correctly encoding.


hdevalence:
Hi, unfortunately this isn't helpful, because it's not possible to review the code in each component PR.

PR #301, opened 2019-10-28
Added constants to reduce recalculations in elligator_risteretto_flavor

### LINKED ISSUE #224: Make Ristretto constants actually constants
The Ristretto Elligator map uses constants which are computed on every run, instead of being loaded as constants.

The lines can be found in this part of the `develop` branch: https://github.com/dalek-cryptography/curve25519-dalek/blob/develop/src/ristretto.rs#L580-L581

These constants should be extracted into `backend/u32` and `backend/u64`. Doing this requires dumping the limb values. Since field elements implement `Debug`, this could be done just by adding some `println!`s, getting the limb values, putting them into the source code, and then removing the printlns.

VictorKoenders:
This should fix #224. 

The link in that issue refers to a different line in the codebase, so it's possible I have updated the wrong functions.

The values are based on the original calculation, but please double-check them.

### DIFF
--- a/src/backend/serial/u32/constants.rs
+++ b/src/backend/serial/u32/constants.rs
@@ -18,6 +18,11 @@ use backend::serial::u32::scalar::Scalar29;
 use edwards::{EdwardsBasepointTable, EdwardsPoint};
 use window::{LookupTable, NafLookupTable8};
 
+/// The value of minus one, equal to `-&FieldElement::one()`
+pub(crate) const MINUS_ONE: FieldElement2625 = FieldElement2625([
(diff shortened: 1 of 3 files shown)


isislovecruft:
Hi @VictorKoenders, thanks for the PR! I've checked the constants, they are correct.

hdevalence:
It would be nice to have some `#[test]`s that the constants are correct in the source code.

VictorKoenders:
And some `#[bench]` for the `elligator_risterotto_flavor`, we currently don't know if it's actually faster

PR #305, opened 2019-10-28
Fix a warning that a temporary value isn't used.

isislovecruft:
rustc doesn't seem smart enough to figure out that when doing:

```rust
let mut tmp1 = T;
let mut tmp2 = T;
let mut tmp3 = T;
for i in 0..63 {
    // ...
    tmp1 = tmp2.double();
    tmp3 = tmp1.to_extended();
    tmp1 = tmp3 + foo;
}
```

that the tmp2 variable is actually used.  This silences that erroneous warning.

### DIFF
--- a/src/backend/serial/scalar_mul/variable_base.rs
+++ b/src/backend/serial/scalar_mul/variable_base.rs
@@ -25,7 +25,7 @@ pub(crate) fn mul(point: &EdwardsPoint, scalar: &Scalar) -> EdwardsPoint {
     // We sum right-to-left.
 
     // Unwrap first loop iteration to save computing 16*identity
-    let mut tmp2 = ProjectivePoint::identity();
+    let mut tmp2;
     let mut tmp3 = EdwardsPoint::identity();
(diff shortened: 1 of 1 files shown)


hdevalence:
The warning is not erroneous, and `rustc` is correct here -- the assignment is unused and could be removed.

isislovecruft:
If it's unused but we don't care, then shouldn't we still silence the warning anyway?

src/backend/serial/scalar_mul/variable_base.rs
```diff
@@ -25,6 +25,7 @@ pub(crate) fn mul(point: &EdwardsPoint, scalar: &Scalar) -> EdwardsPoint {
     // We sum right-to-left.
 
     // Unwrap first loop iteration to save computing 16*identity
+    #[allow(unused_assignments)] // Silence warning since rustc can't tell if tmp2 is used
     let mut tmp2 = ProjectivePoint::identity();
```
tarcieri: I believe this should eliminate the warning:

```suggestion
    let mut tmp2;
```

isislovecruft:
Ah, good catch @tarcieri, thanks!

PR #307, opened 2019-11-15
Remove nightly recommendation now that subtle has stable optimization barriers.

hdevalence:
This bumps the required `subtle` version to `2.2.1`, which contains those changes.


isislovecruft:
(Also now that we use `zeroize` instead of `clear_on_drop`, which was the other reason we recommended nightly, iirc.)

PR #316, opened 2020-02-24
Bump criterion version

huitseeker:
Note this allows the use of benchmark comparison tooling such as [critcmp](https://github.com/BurntSushi/critcmp).


hdevalence:
Thanks! This will be included in a 2.1 release.

PR #321, opened 2020-03-19
implemented Zeroize for CompressedRistretto and RistrettoPoint

gedigi:


isislovecruft:
I think this would be a significant speed regression, considering that, in the majority of protocols, group elements are not commonly secrets. I suggest if your protocol does have secret group elements, that you create a wrapper type to use zeroize, [such as the SecretKey type for a algebraic MAC issuer does in my `aeonflux` crate](https://github.com/isislovecruft/aeonflux/blob/e566dd24cb239724e3854d5fabe9b98c047721c5/src/amacs.rs#L61_L74).

burdges:
There is no regression here since @gedigi does not `impl Drop for` those types, while lacking this your crate needs to manually [reimplement](https://docs.rs/zeroize/1.2.0/src/zeroize/lib.rs.html#240-248) zeroize on `self.W` using atomic operations.

isislovecruft:
Ah thanks, sorry. On cursory review I assumed Drop was also implemented.

isislovecruft (approved):
@gedigi Thanks for the PR!

isislovecruft:
I should note that I changed the implementation to instead "zeroize" points to the identity element (i.e. (±1,0) in affine coordinates), which we were previously using in our implementations of `Default`, instead of the invalid "point" (0,0).

PR #325, opened 2020-05-28
Make Scalar::from_bits a const fn.

rubdos:
`const_fn` is stable since Rust 1.31 (https://github.com/rust-lang/rust/pull/54835) and enables calling `Scalar::from_bits(..)` from other const fn contexts, potentially saving some overhead here and there.

Especially useful in contexts where constants are being built from a bit pattern.

---

My use case is a constructor for a certain large number, where I'd rather start from a slice and set the correct bits that just hardcode in all 32 bytes at once.

I'm not 100% sure about the compatibility guarantees when switching an existing function to `const`, but I imagine it shouldn't interfere anywhere, since the whole test suite still passes.

The other constructors use a call to `.reduce()`, so I didn't bother looking deeper whether they can be made `const` too.

### DIFF
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -243,7 +243,7 @@ impl Scalar {
     /// This function is intended for applications like X25519 which
     /// require specific bit-patterns when performing scalar
     /// multiplication.
-    pub fn from_bits(bytes: [u8; 32]) -> Scalar {
+    pub const fn from_bits(bytes: [u8; 32]) -> Scalar {
         let mut s = Scalar{bytes};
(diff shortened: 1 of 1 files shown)


hdevalence:
Sounds good to me, thanks for the PR!

hdevalence:
@rubdos This change is published in 2.1.0

rubdos:
> @rubdos This change is published in 2.1.0

We're mostly running against git/develop anyway, doesn't matter too much to me at this time. Thanks for the feedback!

PR #335, opened 2020-10-21
Make crate feature alloc work with stable Rust

stephaneyfx:
`feature(alloc)` is not needed and removing it allows to compile with stable Rust when the crate `alloc` feature is on.

`println!` was removed in the test as the corresponding test for `u32` does not have it.


stephaneyfx:
The build failure seems to be due to a spurious network error downloading a crate.

isislovecruft (approved):
Hi @stephaneyfx, thanks for reporting the issue and providing this fix.

PR #336, opened 2020-10-21
Elligator2 mapping for curve25519

### LINKED ISSUE #188: Consider adding a method "from_uniform_bytes" to EdwardsPoint
I want to implement a Pedersen Commitment based on curve25519. For Ristretto I can use  `from_uniform_bytes` for the blinding factor base point (like you do in bulletproof) but I can't find such option for `EdwardsPoint`. 
Is there a way to do create such blind points for `EdwardsPoint`? If currently there is none I would suggest a `from_bytes` method for `CompressedEdwardsY` followed by call to `decompress`  to test if the point is valid and return `Result<EdwardsPoint, Error>`. This is similar to what you can find for secp256k1 (https://docs.rs/secp256k1/0.9.2/secp256k1/key/struct.PublicKey.html#method.from_slice).

Thanks for the great work!

huitseeker:
This present PR:
- can be seen as a more conservative variant of PR #326 (it performs no double mapping), 
- implements the vanilla Elligator2 mapping (matching the [draft RFC](https://tools.ietf.org/id/draft-irtf-cfrg-hash-to-curve-10.html) short of the `expand_message_xmd` expansion¹)
- is, as such, subject to the known distribution biases of Elligator2 (see #326 and https://eprint.iacr.org/2014/043 )

While I'm a fan of Ristretto and would definitely advocate for its use over mapping-to-curve25519 in most situations, this present mapping is still useful in cases where compatibility with other libraries is required — the present mapping is compatible with the one performed in [libsignal](https://github.com/signalapp/libsignal-protocol-c/blob/master/src/curve25519/ed25519/additions/elligator.c) and passes its test vectors.

¹: While this uses an HKDF, I'm happy to attach an implementation of `expand_message_xmd` (here for, essentially, domain separation) should it be requested.

Fixes #188

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ travis-ci = { repository = "dalek-cryptography/curve25519-dalek", branch = "mast
 sha2 = { version = "0.9", default-features = false }
 bincode = "1"
 criterion = "0.3.0"
+hex = "0.4.2"
 rand = "0.7"
 
 [[bench]]
--- a/src/backend/serial/u32/constants.rs
+++ b/src/backend/serial/u32/constants.rs
@@ -63,6 +63,10 @@ pub(crate) const SQRT_M1: FieldElement2625 = FieldElement2625([
 pub(crate) const APLUS2_OVER_FOUR: FieldElement2625 =
     FieldElement2625([121666, 0, 0, 0, 0, 0, 0, 0, 0, 0]);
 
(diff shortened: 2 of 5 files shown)


isislovecruft (approved):
Hi @huitseeker, thanks for the PR.

I'm in favour of merging this for 3.1. I think my only slight preference would be for the function to be called something like `elligator2_encode`, so that if we ever implement the reverse map we have an intuitive naming option available.

@cathieyun @oleganza @hdevalence Any opinions?

PR #342, opened 2021-01-11
Serial backends w/formally-verified field arithmetic for 32 & 64 bits

huitseeker:
## Contribution

This sets up two non-default serial backends, `fiat_u32_backend`and `fiat_u64_backend`. Those backends use field arithmetic generated in Rust from proofs of functional correctness checked by the Coq theorem prover, as part of [the fiat-crypto project](https://github.com/mit-plv/fiat-crypto). More details can be found [in the fiat-crypto paper](http://adam.chlipala.net/papers/FiatCryptoSP19/).

This Rust code is located in the [fiat-crypto crate](https://crates.io/crates/fiat-crypto) [maintained by fiat-crypto team](https://github.com/mit-plv/fiat-crypto/blob/master/.github/workflows/publish.yml), and the present PR simply provides the plumbing necessary to integrate with curve25519-dalek's backend architecture.

## Motivation

This contribution could offer to curve25519-dalek an opt-in backend providing *a "gold standard" of correctness*, and [already used for this reason in Wireguard, Go, BoringSSL, etc](https://github.com/mit-plv/fiat-crypto/issues/902). Dalek developers could also use these backends as a *fuzzing target*, in order to check the correctness upcoming optimizations of the serial code.

## Performance

The performance is lower, but within a single digit of the existing basic operations, and only rises to a ~15% slowdown on composite operations.

https://gist.github.com/huitseeker/d3c7a30fa90fdb43d2815eb4b63a7205

## Maintenability

The backends collectively comprise less than [600 SLoC](https://gist.github.com/huitseeker/57819eb28f3749d62f1acbd48e7ed969). They reuse the scalar and constant definitions to offer a minimal code increment.

The fiat-crypto project [already runs the curve25519-dalek unit tests on the fiat_u64 backend as part of its CI](https://github.com/mit-plv/fiat-crypto/blob/master/etc/ci/test-fiat-rust-curve25519-dalek.sh). Hence the probability that it would break this backend is very low. I would be happy to contribute the few lines of adjustments necessary there to make it run on the fiat_u32 backend as well, should the present PR be merged.


isislovecruft (approved):
Thanks @huitseeker! I'm pleasantly surprised by how small the LOC footprint is to add Fiat support. Hopefully I'll get more time at some point to dig into their output Rust code again and see if there's other optimisations to be made, but ~15% slow down is pretty impressive.

spitters:
@isislovecruft @huitseeker This paper speeds up curve25519: https://cs.adelaide.edu.au/~yval/pdfs/KuepperEGSWCCGWY23.pdf

They say it's the fastest implementation on Intel i9 12G.

PR #346, opened 2021-02-10
implement Zeroize trait on Ristretto curve point types

garbageslam:
This is helpful for hardening some of our cryptographic implementations
that use Ristretto curve points

Please let me know if you would like this done differently, or if there is
some reason you don't want to use `zeroize_derive`. Thank you!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ serde = { version = "1.0", default-features = false, optional = true, features =
 # The original packed_simd package was orphaned, see
 # https://github.com/rust-lang/packed_simd/issues/303#issuecomment-701361161
 packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_bits"], optional = true }
-zeroize = { version = "1", default-features = false }
+zeroize = { version = "1", default-features = false, features = ["zeroize_derive"] }
 
 [features]
 nightly = ["subtle/nightly"]
--- a/src/edwards.rs
+++ b/src/edwards.rs
@@ -105,6 +105,8 @@ use subtle::ConditionallyNegatable;
 use subtle::ConditionallySelectable;
 use subtle::ConstantTimeEq;
 
+use zeroize::Zeroize;
+
 use constants;
 
 use field::FieldElement;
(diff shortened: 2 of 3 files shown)


burdges:
As noted in #321, there are no regressions here since types do not become `Drop`.

I'll note https://github.com/zkcrypto/curve25519-dalek-ng exists @garbageslam

garbageslam:
Yes that is my belief also: this doesn't make the types zeroize on drop, it just tells us how to zeroize them if we need to do that. If you don't need to zeroize curve points then this can be stripped out when the linker sees no one calls it.

Zeroizing via a wrapper is problematic because the wrapper doesn't have private access to fields either. Maybe it can just do some volatile write of a default value ristretto public? At that point I'm basically reimplementing part of Zeroize crate, the volatile write thing is an implementation detail.

Zeroizing curve points seems important because a curve point in your memory may reveal the identity of someone you are interacting with.

garbageslam:
@isislovecruft now that mobilecoin fog is open sourced, I can try to point more specifically to the motivation:

the fog-ingest enclave must decrypt a special fog hint value which reveals the recipient of a transaction, and then do some more cryptography to tag the TxOut so that that person can find their transaction quickly.

https://github.com/mobilecoinfoundation/fog/blob/e16c5eb53f3d35ccd19084210d2a160c150db5f5/fog/ingest/enclave/impl/src/lib.rs#L125

I would like to be able to Zeroize that value so that the enclave will clean up sensitive values after they are no longer needed, and make it harder to conduct a side-channel attack that steals this value from the enclave. However, unless dalek `RistrettoPoint` will implement Zeroize, I'm not sure how I can do that.

isislovecruft (approved):
Hi @garbageslam, thanks for the PR!

I've merged #321 so part of this is superfluous (although I'd prefer the Ristretto implementation to use the Edwards implementation). I also have a slight preference for not using the derive crate, but that's a trivial amount of work and I'm happy to do it myself on top of this branch, if that's okay with you?

garbageslam:
sure, thanks!

PR #349, opened 2021-03-26
[benchmarks-only] Updates the benchmarks

huitseeker:
- removes usages fo the (deprecated) `bench_function_over_inputs`
- introduces a few benchmark groups.

There's many ways to re-organize benchmark groups, happy to take some feedback in consideration in review.

Sample output:
- [before this PR](https://gist.github.com/eb68ce45c895bf2699da5e80622ead1e)
- [after this PR](https://gist.github.com/93ce32deae4051e617951d9a5a9998fd)

### DIFF
--- a/benches/dalek_benchmarks.rs
+++ b/benches/dalek_benchmarks.rs
@@ -7,8 +7,10 @@ use rand::thread_rng;
 #[macro_use]
 extern crate criterion;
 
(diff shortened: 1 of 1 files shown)


isislovecruft (approved):
Thanks @huitseeker! The output is way cleaner now. :)

PR #350, opened 2021-04-13
Merge configurably sized basepoint multiplication lookup tables

isislovecruft:
Closes #311


isislovecruft:
Reminder to remove the legacy 3.x series support code before doing a 4.0 release! It should just mean reverting 
0da8f08, which added support for using the old basepoint table struct without having to do `use curve25519_dalek::traits::BasepointTable`.

PR #352, opened 2021-04-21
Implement Identity for MontgomeryPoint.

isislovecruft:
This can be used in checks for non-contributory behaviour in protocols
where that is a concern.


isislovecruft:
Merging because the change is trivial and I'd like it for a new x25519-dalek release with some other features.

PR #353, opened 2021-05-18
Add `TryFrom<[u8; 32]>` implementations for `Scalar` and `EdwardsPoint `

thomaseizinger:
The first patch unifies use of the `fmt` module across the codebase. This was motivated by the use of `fmt::Display` in the 2nd patch.

The 2nd patch adds `TryFrom<[u8; 32]>` implementations for `Scalar` and `EdwardsPoint`.

I made a call in introducing a dedicated error type to make the `TryFrom` implementations more idiomatic.

My motivation for adding these impls is to directly use `EdwardsPoint` and `Scalar` within structs that implement `serde::Deserialize` via a dedicated `serde_with` attribute here: https://github.com/comit-network/monero-epee-bin-serde/blob/64b44c22db605aad03447b5c0885828a4b9c5292/tests/deserialize_monerod_rpc_responses.rs#L95-L96

The generated implementation enforces `TryFrom::Error: std::error::Error`. Hence, if extending the public API with dedicated errors types is a NACK, then at least from my perspective, we might as well not add the impls at all because I wouldn't able to use them.

Curious to hear your thoughts!


thomaseizinger:
Closing due to inactivity.

PR #357, opened 2021-06-20
Add inverse Elligator2 mapping to Curve25519.

lealanko:
This PR adds `montgomery_elligator_decode`, an inverse to `montgomery_elligator_encode`. This is my first contribution, so I'm very interested in advice on how to best accommodate Dalek's conventions.

Some points of note:
* I'm not fond of the naming these two functions. The `encode` function is clearly named from the perspective of hash-to-curve operations, where pre-existing uniform data is "encoded" into a curve point. However, the inverse mapping can be used to map an existing curve point invertibly into uniform-looking data. Calling this operation "decode" is confusing. I would suggest more neutral naming like `map_to_field` and `map_from_field`.
* The implementation is not constant-time, since the expected usage is to call the function repeatedly with randomly generated points until one is found that is in the mapping's domain. Still, in that case the timing leaks some information about the random generator. Should a constant-time version be provided, even if that would make rejection sampling more expensive?
* As part of its input checking, the function has to ensure that the point is on the curve. This is expensive, and seems to be in conflict with Dalek's stated goal of "making illegal states unrepresentable". There seems to be need for a type that contains a u-coordinate that is guaranteed to belong to a point on the curve.
* I test some randomly generated points to check that functions are inverses of each other, but I have no idea what's a reasonable number of iterations. Also, I'm not fond of using `OsRng` since that means that test runs are unpredictable..
* `montgomery_elligator_encode` does compute the sign of the v-coordinate (as `eps_is_sq`) but it does not return it. This is an unfortunate asymmetry with the inverse function (whose output depends on the sign), and it may be critical for applications that require a full curve point, not just a u-coordinate.

I can try to address some of these issues already within this PR if that's a reasonable thing to do.

### DIFF
--- a/src/montgomery.rs
+++ b/src/montgomery.rs
@@ -224,7 +224,7 @@ pub(crate) fn elligator_encode(r_0: &FieldElement) -> MontgomeryPoint {
 /// # Returns
 ///
 /// Either `None`, if the point couldn't be mapped, or `Some(fe)` such
-/// that `elligator_encode(&fe) == point`.
+/// that `fe` is nonnegative and `elligator_encode(&fe) == point`.
 ///
 /// [elligator] https://elligator.cr.yp.to/elligator-20130828.pdf
 ///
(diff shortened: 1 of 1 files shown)


src/montgomery.rs
```diff
@@ -193,6 +193,81 @@ pub(crate) fn elligator_encode(r_0: &FieldElement) -> MontgomeryPoint {
     MontgomeryPoint(u.to_bytes())
 }
 
+/// Perform the inverse Elligator2 mapping from a Montgomery point to
+/// field element. This algorithm is based on [Elligator:
```
koba-e964, line 197: nit: the indefinite article
```suggestion
/// a field element. This algorithm is based on [Elligator:
```

src/montgomery.rs
```diff
@@ -193,6 +193,81 @@ pub(crate) fn elligator_encode(r_0: &FieldElement) -> MontgomeryPoint {
...
+
+    // if !v_is_negative: r = sqrt(-u / 2(u + a)) = root * u
+    // if  v_is_negative: r = sqrt(-(u+A) / 2u)   = root * (u + A)
+    let add = FieldElement::conditional_select(&u, &u_plus_A, v_is_negative);
+    let r = &root * &add;
+
+    // Both r and -r are valid results. Pick the nonnegative one.
+    let result = FieldElement::conditional_select(&r, &-&r, r.is_negative());
+
+    return Some(result);
```
koba-e964, line 268: You may omit `return` here in the last line of a function, as in other functions in this file.
```suggestion
    Some(result)
```

PR #359, opened 2021-07-09
update rand to 0.6

trinity-1686a:
This seems to be wanted by [Arti's developers](https://gitlab.torproject.org/tpo/core/arti/-/blob/main/WANT_FROM_OTHER_CRATES#L6-7)


thibault-martinez:
We would appreciate this to be merged as well, for some features-related reasons, updating `rand_core` to `0.6` allows this crate to actually be compiled on `no_std` environments.

isislovecruft:
This is a breaking change and so it has been included in the `4.0.0-pre.0` release.

PR #361, opened 2021-07-19
re-added deletions of 1491f0d

iquerejeta:
Seems that the [`Update copyright years`](https://github.com/dalek-cryptography/curve25519-dalek/commit/1491f0db364f8d8fc1072caff88967248c555ef5) commit took away the elligator2 functionality for Edwards curves (and some zeroize implementation), which was included in #336 . 

Just copied the code that was deleted in that commit. No additions or modifications.

### DIFF
--- a/src/edwards.rs
+++ b/src/edwards.rs
@@ -101,10 +101,13 @@ use core::ops::{Add, Neg, Sub};
 use core::ops::{AddAssign, SubAssign};
 use core::ops::{Mul, MulAssign};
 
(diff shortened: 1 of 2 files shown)


isislovecruft:
Hi @iquerejeta! Thanks for finding this and pointing it out. It was indeed a merge mistake. I've gone ahead and fixed it by reverting the commit with the mistake and redoing it, so as to preserved authorship history.

PR #369, opened 2021-09-14
Updated deps

rozbb:
**This requires a major version bump**, as it is a breaking change. Because `curve25519-dalek` exposes its `rand` dependency, if another crate previously used an `OsRng` from v0.7 to generate a scalar (which e.g. `ed25519-dalek` does), it would not compile anymore. This seemed awkward to me but it looks like `merlin` [does this too](https://github.com/zkcrypto/merlin/blob/main/CHANGELOG.md).

Notes on dependency updates:
* `byteorder = "1"` is the line that's used in byteorder's own [README](https://github.com/BurntSushi/byteorder/blob/master/README.md)
* `subtle` is semantic-versioned (check me on this)
* I didn't switch `fiat_crypto` to `0.1` because I have no idea how it's versioned. Thankfully it's bug-free by definition, so that's not an issue >.>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,25 +33,25 @@ travis-ci = { repository = "dalek-cryptography/curve25519-dalek", branch = "mast
 [dev-dependencies]
 sha2 = { version = "0.9", default-features = false }
 bincode = "1"
-criterion = { version = "0.3.0", features = ["html_reports"] }
-hex = "0.4.2"
-rand = "0.7"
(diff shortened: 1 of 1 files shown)


paolobarbolini:
I guess this PR should target the [`release/4.0` branch](https://github.com/dalek-cryptography/curve25519-dalek/tree/release/4.0)

rozbb:

PR #377, opened 2021-11-11
Hash to curve as defined in the standard

iquerejeta:
The current implementation is not compatible with the current definition of the standard. This PR provides a hash-to-curve implementation as defined in draft-irtf-cfrg-hash-to-curve-12.

* Implementation of `hash_to_field` as defined in the standard
* Implementation of `hash_to_curve` as defined in the standard, by changing the mechanism over which we chose the sign.
* For the point above, had to change the `elligator_encode` to return whether `eps` is a square or not (required for `hash_to_curve`).
* Included test vectors of the draft.
* Included `FieldElement::from_bytes_wide(bytes: &u8; 64])` to reduce integers encoded in 64 bytes.


iquerejeta:
I see in #438 that the function `hash_to_bytes` was deprecated. Is there any chance to include the `hash_to_curve` as defined in the [standard](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/)? 
cc: @rozbb

rozbb:
Yes, thank u for following up. I do want to merge this but I haven't gotten a chance to do a proper review. We really need to get to a 4.0 release, so I think this might have to wait until 4.1. Thank you so much for this contribution. I promise it will get the attention it deserves.

tarcieri:
It would also be interesting if someone could attempt to impl the [`hash2curve` traits from the `elliptic-curve` crate](https://docs.rs/elliptic-curve/latest/elliptic_curve/hash2curve/index.html) i.e. as an optional dependency

iquerejeta:
Happy to do it 👍  given that there is no rush, I should be able to dedicate a few cycles in a reasonable amount of time.

tarcieri:
Well for starters, it's full of merge conflicts

paweljakubas:
Thank you very much @iquerejeta (I have also started rebasing your PR on Fri but you was quicker :smile: )
@tarcieri are we fine now? Thanks in advance for your help and time spent here and cheers :wave:

paweljakubas:
@tarcieri @rozbb @oleganza  can we do with @iquerejeta something more ?
Thanks in advance for your time and please do not be angry with me for my pinging :pray: 

We are working on VRF library based on dalek-cryptography and basing atm on this branch and we would like to have just master as dependency rather than branch. 

Cheers!

abailly:
Apologies for pestering you @tarcieri but we ([Amaru team](https://github.com/pragma-org/amaru)) ultimately depend on this PR to be included in a proper version of curve255159-dalek in order to simplify our dependencies on [VRF](https://github.com/txpipe/vrf) library and also to be able to request a proper audit on a stable code base. 
Is there anything we can help on to move this forward and cut a release in the near future?

tarcieri:
@abailly you should probably ask @rozbb. I'm working on a lot of prerequisites to the next releases and don't have time to review this, sorry

abailly:
Thanks a lot @tarcieri, waiting for @rozbb's feedback then 🙏

tarcieri:
I'll also note this looks fine to me at first glance, but I'd prefer one of us go through it in detail

curve25519-dalek/src/backend/serial/u64/constants.rs
```diff
@@ -31,6 +31,16 @@ pub(crate) const MINUS_ONE: FieldElement51 = FieldElement51::from_limbs([
     2251799813685247,
 ]);
 
+/// sqrt(-486664)
+#[cfg(feature = "digest")]
+pub(crate) const ED25519_SQRTAM2: FieldElement51 = FieldElement51::from_limbs([
```
rozbb, line 36: Nit: can you add the following test to the end of `src/constants.rs`?

```rust
/// Test that ED25519_SQRTAM2 squared is MONTGOMERY_A_NEG - 2
#[test]
fn test_sqrt_a_minus_2() {
    let one = FieldElement::ONE;
    let a_minus_two = &(&constants::MONTGOMERY_A_NEG - &one) - &one;

    assert_eq!(constants::ED25519_SQRTAM2.square(), a_minus_two)
}
```

curve25519-dalek/src/edwards.rs
```diff
@@ -574,12 +576,37 @@ impl EdwardsPoint {
         CompressedEdwardsY(s)
     }
 
+    #[cfg(feature = "digest")]
+    /// Perform hashing to curve, with explicit hash function and DST using the suite
+    /// edwards25519_XMD:SHA-512_ELL2_NU_
+    ///
+    /// See https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-12#section-6.8.2
+    pub fn hash_to_curve_dst<D>(bytes: &[u8], dst: &[u8]) -> EdwardsPoint
```
rozbb, line 584: Can we rename `dst` to `domain_sep` or something similar, and update the docstring? `dst` reads too much like "destination" imo

curve25519-dalek/src/backend/serial/u64/field.rs
```diff
@@ -362,6 +362,24 @@ impl FieldElement51 {
         ])
     }
 
+    /// Load a `FieldElement51` from 64 bytes, by reducing modulo q.
+    pub fn from_bytes_wide(hash: &[u8; 64]) -> FieldElement51 {
```
rozbb, line 366: Question: this seems to only be implemented for the serial backends. Do the vector backends not need something like this?

curve25519-dalek/src/backend/serial/u64/field.rs
```diff
@@ -362,6 +362,24 @@ impl FieldElement51 {
...
+        let mut fl = [0u8; 32];
+        let mut gl = [0u8; 32];
+        fl.copy_from_slice(&hash[..32]);
+        gl.copy_from_slice(&hash[32..]);
+        fl[31] &= 0x7f;
+        gl[31] &= 0x7f;
+
+        let mut fe_f = Self::from_bytes(&fl);
+        let fe_g = Self::from_bytes(&gl);
+        fe_f.0[0] = fe_f.0[0] + (hash[31] >> 7) as u64 * 19 + (hash[63] >> 7) as u64 * 722;
```
rozbb, line 376: Where does this math come from?

rozbb (changes requested):
Hi hi! Thank you so much for this. This overlapped with #612, which seems like a superset of this, and I was hoping it'd go faster. But it's not and that's fine!

I left a few comments here and there, but overall looks great!

curve25519-dalek/src/backend/serial/u64/field.rs
iquerejeta: I added them only in serial backends to follow how the library is structured. These functions `from_bytes_wide` only exist for serial in the library.

iquerejeta (commented):
Trying to refresh my memory with the first question!

curve25519-dalek/src/edwards.rs
iquerejeta: We could also simply cal it `hash_to_curve`, given that that's how the spec is defined.

curve25519-dalek/src/edwards.rs
```diff
@@ -574,12 +576,37 @@ impl EdwardsPoint {
...
+        D: Digest<OutputSize = U64> + Default,
+    {
+        let fe = FieldElement::hash_to_field::<D>(bytes, dst);
+        let (M1, is_sq) = crate::montgomery::elligator_encode(&fe);
+        let mut E1_opt = M1
+            .to_edwards(0)
+            .expect("Montgomery conversion to Edwards point in Elligator failed");
+
+        // Now we recover v, to ensure that we got the sign right.
+        let mont_v =
```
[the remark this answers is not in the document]
iquerejeta: There are a few ways this is done. In libsodium, for example, they compute `v` via a square root, pick the sign accordingly, and then convert to Edwards using that sign ([see here](https://github.com/jedisct1/libsodium/blob/d537f3010a4e8e4ab5afdbf782799c48d8f09272/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c#L2739)).

What we do is different: note that `is_square` tells us the sign of the Montgomery `v` coordinate, not the Edwards `X`. However, the `to_edwards` function only lets us specify the sign of `X`, so we:

- Call `to_edwards` with an arbitrary sign (in our case, 0),
- Recompute `v` from the `u` and the Edwards `X` using the mapping formula (recall that `X = ED25519_SQRTAM2 * u / v`)
- Then flip the sign of `X` if the resulting sign of `v` doesn't match `is_sq ^ v.is_negative()`.

This avoids using square roots and works with our existing conversion interface (the current interface would _not_ allow us to do it the square root way, because there is no `Montgomery -> Edwards` conversion that takes both `(u, v)` as input). We could add code that enables it—happy to benchmark it.

abailly:
@rozbb Would it be possible to get approval to run workflows so that we can check CI is passing?

curve25519-dalek/src/field.rs
```diff
@@ -303,6 +309,51 @@ impl FieldElement {
...
+
+        hasher = hasher.chain_update(l_i_b_str).chain_update([0u8]);
+
+        let mut domain_sep_len = 0usize;
+        for slice in domain_sep {
+            hasher = hasher.chain_update(slice);
+            domain_sep_len = domain_sep_len
+                .checked_add(slice.len())
+                .expect("Unexpected overflow from domain separator's size.");
+        }
```
daxpedda, line 336: Checking for overflow on slice length with a `usize` is unnecessary, you would run out of memory before it overflows.

curve25519-dalek/src/field.rs
```diff
@@ -303,6 +306,39 @@ impl FieldElement {
...
+
+    #[cfg(feature = "digest")]
+    /// Hash_to_field as described in hash_to_curve standard
+    pub fn hash_to_field<D>(bytes: &[u8], dst: &[u8]) -> Self
+    where
+        D: Digest<OutputSize = U64> + Default,
+    {
+        let len_in_bytes = 48;
+        let l_i_b_str = [0u8, len_in_bytes];
+        let z_pad = [0u8; 128];
```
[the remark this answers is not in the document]
daxpedda: Now that we do handle `BlockSizeUser` correctly, we also need to enforce [further requirements from the spec](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#section-5.4.1-4):
> For correctness, H requires b <= s.

This can be done via:
```rust
BlockSizeUser<BlockSize: Greater<D::OutputSize, Output = True>>
```

daxpedda (approved):

daxpedda:
Just a quick summary on what could be addressed in future improvements:
- Return a `Result` instead of panicking with invalid DSTs.
- Support hashes not with an output size of 64 bytes.
- Support [DSTs larger than 255 bytes](https://www.rfc-editor.org/rfc/rfc9380.html#name-using-dsts-longer-than-255-).
- Support `expand_message_xof`. ([`expand_message_xmd` is the suggested suite](https://www.rfc-editor.org/rfc/rfc9380.html#name-suites-for-curve25519-and-e) so it is the right default)

curve25519-dalek/src/edwards.rs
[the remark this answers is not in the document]
rozbb: Sorry, my brain is a bit fried but I'll try to understand better. To be clear, my absolute top concern is maintainability. It would be way easier for me if all the formulas in this PR used notation directly from the spec. Do you think that'd be possible? I don't mind some added code (eg the conversion method you mentioned) to get there. I'll even accept a performance hit if necessary. I think @tarcieri would agree

curve25519-dalek/src/backend/serial/u64/field.rs
rozbb: I added some comments that explain it all, but I have one last question: why does the `38 * ` expression below guaranteed to not overflow? Is there a property in libsodium that guarantees this? Do we have the same?

iquerejeta: `FieldElement51` represents an element of the field `2^255 - 19`. In other words, it represents a number of size 255 bits. With 4 limbs of `u64`, we would have enough for representing these numbers. However, `FieldElement51` represents the number with 5 limbs of 51 bits, and as [documented](https://github.com/dalek-cryptography/curve25519-dalek/blob/main/curve25519-dalek/src/backend/serial/u64/field.rs#L30): 
```
the coefficients are allowed to grow up to 2^54 between reductions modulo p
```

So even if the type is u64, we can do these operations here with the guarantee that we won't have an overflow.

curve25519-dalek/src/edwards.rs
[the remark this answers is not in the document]
iquerejeta: I'd be happy to do that. I'm wondering if it makes sense to change the `elligator_encode` function directly, and perform elligator as specified in the standard. What I have in mind is that we implement the straight line implementation they present in the standard ([here](https://www.rfc-editor.org/rfc/rfc9380.html#appendix-F.3). And then we have two functions, an elligator2 for montgomery, and an elligator2 for edwards. WDYT?

curve25519-dalek/src/backend/serial/u64/field.rs
rozbb: If `from_bytes` produces an element with a limb x = 2^51, then isn’t it the case that x * 38 > x * 32 = 2^56 > 2^54?

curve25519-dalek/src/edwards.rs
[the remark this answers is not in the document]
iquerejeta: I started trying to do this, and there's some functionality missing. The only `pow` implementation is `pow2k`. Following the inline implementation we need to compute 
```
y11 = tv2^c4
```
where
```
c4 = (q - 5) / 8
```

with c4 not being a power of 2 (see step 16 [here](https://datatracker.ietf.org/doc/html/rfc9380#name-curve25519-q-5-mod-8-k-1)). Don't know if I'm missing a trick here. I think at this point it would be easier to properly document, rather than implementing new functions. If you are happy with that path I can add proper documentation to the function, and leave it as is.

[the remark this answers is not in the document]
iquerejeta: I've added a comment. I decided to link [this](https://datatracker.ietf.org/doc/html/rfc9380#name-elligator-2-method) section, instead of the inline section, as I believe its easier to follow along. Let me know what you think.

curve25519-dalek/src/backend/serial/u64/field.rs
rozbb: Ah it looks like you indeed can pass limbs up to 2^64

https://github.com/dalek-cryptography/curve25519-dalek/blob/dd5bd108d6985491acb3de25497fb082a29c9fb7/curve25519-dalek/src/backend/serial/u64/field.rs#L293

So the current impl should work. Could you add a comment saying this when you get a chance?

paweljakubas:
hi @iquerejeta @rozbb @tarcieri  great work on both implementation and review side! Are we fine now to merge this PR ?

rozbb:
I believe we’re still waiting on a refactor to write the Elligator functions in the way the spec writes them, right?

iquerejeta:
I think I misunderstood then. I thought we agreed to simply document the function properly, as following the steps as they are presented in the draft required additional functionality. The discussion was [here](https://github.com/dalek-cryptography/curve25519-dalek/pull/377#discussion_r2111469297)

abailly:
Is there anything we (@paweljakubas , @iquerejeta or I) can do to help merging this PR?

tarcieri:
@paweljakubas @iquerejeta @abailly can you resolve the merge conflicts?

@rozbb I would suggest merging after that, and leaving updating the Elligator implementation as a followup (there are also several open PRs to do that, although I'm not sure if any actually follow the spec)

abailly:
Thanks, that would be awesome!

iquerejeta:
Rebase done 👍 
ccL @rozbb

tarcieri:
@iquerejeta there are some deprecations you'll need to fix

iquerejeta:
My bad 🙏  forgot about those in the rebase.

rozbb:
Thank you, and sorry for the delay. Will make time for this tomorrow

curve25519-dalek/src/field.rs
```diff
@@ -492,4 +556,113 @@ mod test {
     fn batch_invert_empty() {
         FieldElement::batch_invert(&mut []);
     }
+
+    #[test]
+    #[cfg(feature = "digest")]
+    fn from_hash_wide() {
```
rozbb: One last question I promise: Where are these test vectors from?

rozbb:
Ok, just cleaned up docs and clarified the origin of some KATs. I left just 1 question, namely where the from_bytes_wide test vectors came from. Other than that, I think we're good to go!

curve25519-dalek/src/field.rs
iquerejeta: I must admit that I do not remember. It's been a while. I was looking for those test vectors in the libsodium implementation or some drafts (which were the most likely sources) but couldn't find any. 

I'm happy to regenerate them from any source you'd like and document it accordingly.

rozbb:
This is confusing to me. It seems the `fiat` backend doesn't have a `FieldElement::from_bytes_wide` function, which is necessary for `hash_to_field`. For some reason, this compiled in the past. Why?

iquerejeta:
What should we do about the fiat backend? The logic of the from_bytes_wide is not from fiat crypto 🤔

rozbb:
@iquerejeta Well my question is: what was it doing when the build was succeeding? Shouldn’t the fiat build have always been failing?

iquerejeta:
Yeah, that's a good question. I have no idea why. I believe we never had an implementation of from_bytes_wide for fiat.

tarcieri:
Regarding `fiat-crypto` and wide reductions, this might be helpful: https://words.filippo.io/wide-reduction/

rozbb:
Thank you @tarcieri, that gave me the idea to make `from_bytes_wide` generic over any field element, rather than over specific implementations. This simplifies the implementation, makes it work for `fiat`, and also does not detectibly slow down `hash_to_curve` on my machine.

Do you wanna do one last look-over? I had to bump the `fiat-crypto` dependency because we needed `fiat_25519_from_bytes` to be a const fn, which it previously wasn't. I assume that's fine, since we don't export anything from there.

tarcieri (approved):
LGTM

rozbb:
Thank you so much @iquerejeta @tarcieri and @daxpedda !!

abailly:
🎉 Thank you everyone! Looking forward to next release so that we can release our stuff too :)

iquerejeta:
🎉  Thanks everyone! 🙏

PR #382, opened 2022-02-02
Use ok_or_else instead of ok_or in serde decoding

jrose-signal:
Serde errors are not simple enums; they format a full error string from their arguments. It's worth not doing that up front.


jrose-signal:
Bumping this, it's free performance! (Also caught by `cargo clippy --all-features`, in addition to a less-interesting suggestion to use `enumerate()` for deserializing the 32-tuples representing scalars and points.)

tarcieri:
Thanks!

PR #384, opened 2022-02-25
Making Scalar canonical checks constant time

elichai:
Some higher-level protocols might need to serialize and deserialize sensitive scalars, for that you'd want a constant time deserialization.
Currently to achieve that you need to:
1. Make sure the highest bit is off
2. Call `Scalar::from_bits`, and then call `is_canonical`.
and combine the result of both in constant time.

Instead, I've implemented a constant time `is_canonical` check, which is actually *faster*(~20%) than the current check (as it doesn't require a full reduction), and also made `from_canonical_bytes` also be constant time.

Caveats:
1. Both `is_canonical` and `from_canonical_bytes` now return `Choice` and `CtOption` which means that `subtle` is part of the public API (until now it was only true via subtle's traits).
2. `from_canonical_bytes` became much slower *if and only if* the most significant bit is on (as I've removed the `(bytes[31] >> 7) != 0u8` check)

Benchmarks from my machine (i9-9980HK @ 2.4GHz turbo off)
Before:
```
Max scalar from_canonical_bytes                                                                                                                                                                                                                                                
                        time:   [28.101 ns 28.151 ns 28.216 ns]                                                                                                                                                                                                                
                        change: [-0.4743% +0.0795% +0.6925%] (p = 0.82 > 0.05)                                                         
                        No change in performance detected.                                                                             
Found 11 outliers among 100 measurements (11.00%)                  
  6 (6.00%) high mild                                                                                                                  
  5 (5.00%) high severe                                            
                                                                   
Zero scalar from_canonical_bytes                                                                                                       
                        time:   [199.10 ns 199.31 ns 199.58 ns]                                                                        
                        change: [-0.6711% -0.1692% +0.1507%] (p = 0.57 > 0.05)                                                         
                        No change in performance detected.                                                                             
Found 9 outliers among 100 measurements (9.00%)                    
  5 (5.00%) high mild                                                                                                                  
  4 (4.00%) high severe                                            
                                                                   
Rand scalar from_canonical_bytes                                                                                                       
                        time:   [98.983 ns 99.066 ns 99.162 ns]
                        change: [-4.6623% -2.0323% -0.1946%] (p = 0.07 > 0.05)
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)                                                                                        
  4 (4.00%) high mild
```
After:
```
Max scalar from_canonical_bytes                                                                                                                                                                                                                                                
                        time:   [158.61 ns 159.17 ns 159.97 ns]                                                                                                                                                                                                                
                        change: [+460.89% +464.15% +466.74%] (p = 0.00 < 0.05)                                                                                                                                                                                                 
                        Performance has regressed.                                                                                                                                                                                                                             
Found 8 outliers among 100 measurements (8.00%)                                                                                                                                                                                                                                
  2 (2.00%) high mild                                                                                                                                                                                                                                                          
  6 (6.00%) high severe                                                                                                                                                                                                                                                        
                                                                                                                                       
Zero scalar from_canonical_bytes                                                                                                       
                        time:   [158.71 ns 158.92 ns 159.19 ns]    
                        change: [-20.347% -20.187% -19.984%] (p = 0.00 < 0.05)                                                         
                        Performance has improved.                  
Found 10 outliers among 100 measurements (10.00%)                  
  4 (4.00%) high mild                                                                                                                  
  6 (6.00%) high severe                                                                                                                
                                                                                                                                       
Rand scalar from_canonical_bytes                                                                                                       
                        time:   [138.34 ns 138.43 ns 138.54 ns]    
                        change: [+39.316% +39.533% +39.747%] (p = 0.00 < 0.05)                                                         
                        Performance has regressed.                 
Found 13 outliers among 100 measurements (13.00%)                  
  5 (5.00%) high mild                                                                                                                  
  8 (8.00%) high severe                                        
```


elichai:
Note:
Using the following implementation still produces CT code but is much more performant: (https://godbolt.org/z/Y6KvfoMvn)
```rust
pub fn is_canonical(&self) -> Choice {
    let mut over = 0;
    let mut under = 0;
    for (this, l) in self.unpack().0.iter().zip(&constants::L.0).rev() {
        let gt = (this < l) as u8;
        let eq = (this == l) as u8;
        under |= (!gt & !eq) &!over;
        over |= gt;
    }
    Choice::from(under)
}
```

```
Max scalar from_canonical_bytes                                                                             
                        time:   [46.327 ns 46.615 ns 46.978 ns]
                        change: [-70.884% -70.734% -70.581%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) high mild
  8 (8.00%) high severe

Zero scalar from_canonical_bytes                                                                             
                        time:   [46.194 ns 46.294 ns 46.413 ns]
                        change: [-70.940% -70.853% -70.771%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  6 (6.00%) high mild
  7 (7.00%) high severe

Rand scalar from_canonical_bytes                                                                            
                        time:   [23.529 ns 23.947 ns 24.533 ns]
                        change: [-83.142% -82.922% -82.652%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe
```

elichai:
Yet another option, is using https://github.com/veorq/cryptocoding for the eq/gt code, and doing it like this:
```rust
  pub fn is_canonical(&self) -> Choice {
      let mut over = 0;
      let mut under = 0;
      for (&this, &l) in self.unpack().0.iter().zip(&constants::L.0).rev() {
          let gt = (l ^ ((l ^ this) | (l.wrapping_sub(this) ^ this))) >> 63;
          let eq = 1 ^ ((this.wrapping_sub(l) | l.wrapping_sub(this)) >> 63);
          under |= (!gt & !eq) & !over;
          over |= gt;
      }
      Choice::from((under & 1) as u8)
  }
```
With the following ASM: https://godbolt.org/z/r4zvrG9ex
And these perfs:
```
Max scalar from_canonical_bytes
                        time:   [46.914 ns 47.635 ns 48.383 ns]
                        change: [-71.013% -70.465% -69.927%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 18 outliers among 100 measurements (18.00%)
  3 (3.00%) high mild
  15 (15.00%) high severe

Zero scalar from_canonical_bytes
                        time:   [46.148 ns 46.674 ns 47.403 ns]
                        change: [-70.320% -69.843% -69.337%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 17 outliers among 100 measurements (17.00%)
  3 (3.00%) high mild
  14 (14.00%) high severe

Rand scalar from_canonical_bytes
                        time:   [25.928 ns 26.083 ns 26.288 ns]
                        change: [-81.376% -81.078% -80.819%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe
```

elichai:
I rebased the branch on top of #472 and updated the benchmarks, description and topic
@tarcieri Would appreciate a review whenever you have time :), although it's less important now that #472 added a constant time check

Cargo.toml
```diff
@@ -51,7 +51,7 @@ required-features = ["rand_core"]
 cfg-if = "1"
 rand_core = { version = "0.6.4", default-features = false, optional = true }
 digest = { version = "0.10", default-features = false, optional = true }
-subtle = { version = "2.3.0", default-features = false }
+subtle = { version = "^2.4", default-features = false }
```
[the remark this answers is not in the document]
elichai: You're right. Fixed.

src/scalar.rs
```diff
@@ -1138,7 +1136,15 @@ impl Scalar {
...
+        let mut over = Choice::from(0);
+        let mut under = Choice::from(0);
+        for (this, l) in self.unpack().0.iter().zip(&constants::L.0).rev() {
+            let gt = this.ct_gt(l);
+            let eq = this.ct_eq(l);
+            under |= (!gt & !eq) & !over;
+            over |= gt;
+        }
+        under
     }
```
tarcieri, line 1148: This seems like it should be fine.

Another way to do this check is to subtract the modulus and check for underflow, although I don't think that's going to be straightforward with `UnpackedScalar::sub`.

elichai: That's an interesting idea!

I hacked your idea right now, and it seems to be the fastest yet :)

it drops to ~40ns 

I'll verify the code by hand and figure out what's the best place/naming for it but that's the logic I have in mind (and it passes the tests) https://godbolt.org/z/7jbqTxo6r

(sadly this would need to be added to the backends)

elichai:
Rebased ontop of `main`

PR #394, opened 2022-05-17
Relax Rng trait bounds to allow `?Sized` Rngs

garbageslam:
This allows the code to compile if you pass it `&mut dyn RngType`,
since trait objects are unsized.

See here for an example of caller code that is simplified by this
change:

https://github.com/mobilecoinfoundation/mobilecoin/pull/1977#discussion_r872906913

### DIFF
--- a/src/ristretto.rs
+++ b/src/ristretto.rs
@@ -651,7 +651,7 @@ impl RistrettoPoint {
     /// discrete log of the output point with respect to any other
     /// point should be unknown.  The map is applied twice and the
     /// results are added, to ensure a uniform distribution.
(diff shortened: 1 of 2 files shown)


rozbb:
Of all the options, I think `R: ... + ?Sized` is the most Rustic, and less surprising than `&mut dyn`, as Artyom mentioned in the linked thread. Happy to merge. @tarcieri sounds good?

tarcieri:
Yeah that’s where we netted out too I think. The only difference being we switched to the new `CryptoRngCore` trait. But this is fine for now.

PR #386, opened 2022-03-21
feat: remove limitation on zeroize

poliorcetics:
Helps with compatibility with recent versions of other crypto crates

PR #397, opened 2022-06-20
Update zeroize and MSRV accordingly

### LINKED ISSUE #387: Old version of zeroize
curve25519-dalek requires zeroize < 1.4, but a lot of crates already depend on 1.4.1 etc.
Any plans to bump zeroize deps?

nazar-pc:
Turns out similar to #386, but updates other relevant things.

I need this because it causes issues deep in the dependencies, my chain looks like this:
```
my-crate -> substrate -> libp2p -> snow -> curve25519-dalek
```

So this crate deep in dependency tree causes too many issues. I think bumping it in pre-release will not cause any issues for anyone in the ecosystem.

Closes #386, fixes #387

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ name = "curve25519-dalek"
 # - update html_root_url
 # - update README if required by semver
 # - if README was updated, also update module documentation in src/lib.rs
-version = "4.0.0-pre.2"
+version = "4.0.0-pre.3"
 authors = ["Isis Lovecruft <isis@patternsinthevoid.net>",
            "Henry de Valence <hdevalence@hdevalence.ca>"]
 readme = "README.md"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -50,7 +50,7 @@ serde = { version = "1.0", default-features = false, optional = true, features =
 # The original packed_simd package was orphaned, see
 # https://github.com/rust-lang/packed_simd/issues/303#issuecomment-701361161
 packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_bits"], optional = true }
-zeroize = { version = ">=1, <1.4", default-features = false }
+zeroize = { version = "1.5.5", default-features = false }
```
tarcieri, line 53: There's no need for it to be this restrictive, and making it restrictive creates interoperability problems.

```suggestion
zeroize = { version = "1", default-features = false }
```

nazar-pc: Makes sense, updated

fabioberger:
@tarcieri any reason this hasn't been merged yet?

tarcieri:
I’m not a maintainer of this crate

kimpers:
@isislovecruft would love to see this merged as this issue is blocking us from updating our dependencies

fabioberger:
@isislovecruft is there a reason why this isn't getting merged?

rozbb:

Issue #398, opened 2022-06-23
Batch double-and-compress (on ristretto255) fails on the identity point


pornin:
`RistrettoPoint::double_and_compress_batch()` may fail if one of the input points is the Ristretto identity. The function computes, for each input point, the values `e`, `f`, `g`, and `h` (in a `BatchCompressState` structure) then inverts the product of all four values. All the inversions are performed in a call to `FieldElement::batch_invert()`. However, that function does not tolerate zeros as input; there is even an explicit `assert!()` to verify that all inputs are non-zero.

Values `e` and `f` can be zero for some points which all represent the Ristretto identity (namely, when the representative element is, on the twisted Edwards curve, a point of order 1, 2 or 8). In particular, you get that situation by using the point returned by `RistrettoPoint::identity()`. In that case, the corresponding input to `batch_invert()` is zero, and the assertion fails, which triggers a panic.

To fix this issue, it suffices to make `batch_invert()` tolerate zeros, returning zero as an "inverse" of itself (while zero is not formally invertible, returning zero as its own inverse works with the double-and-compress formulas). This can be done by "skipping" zeros (in a constant-time way, of course) in both passes of the batch invert function.

pornin:
PR #399 contains a fix.

rozbb:
Closed in #399

PR #399, opened 2022-06-23
Fix double-and-compress on Ristretto identity (issue #398).

pornin:
This makes `FieldElement::batch_invert()` able to process inputs of value zero (the "inverse" of zero is formally defined to be zero; this is usually the most useful value to return in that case, and it works with the Ristretto double-and-compress formulas). I also modified the unit tests to check that functionality explicitly, both in `batch_invert()` itself, and in Ristretto double-and-compress.


rozbb:
Thank you for this! Good changes and good regression test. Would you be able to rebase on `release/4.0`?

PR #403, opened 2022-07-15
Update Dependencies

npmccallum:
In particular, we're having problems with dependency trees in the reference implementation for the wasi-crypto specification.


npmccallum:
FYI @tarcieri

gitmalong:
What does it take to get it merged @isislovecruft?

PR #404, opened 2022-08-13
Expose function for directly encoding bytes to curve, without hashing.

AmitPr:
This PR:
* Adds a `bytes_to_curve` method that allows one to encode arbitrary bytes to the Edwards curve. This function is identical to `hash_from_bytes`, without the hashing.
* For convinience, exposes an `AsRef<[u8]>` implementation for `CompressedEdwardsY`.
* Relaxes `zeroize` dependency to allow newer versions, as the existing version was outdated & incompatible with some newer libraries, for some reason.


AmitPr:
Note, seems that there's many open PRs that address the incompatibility issues that arise from the outdated `zeroize` dependency. Would be happy to remove this from this PR if one of those PRs is addressed.

PR #408, opened 2022-09-07
Update dependencies, move to 2021 edition

rickwebiii:
Update crate to 2021 edition (just had to fix crate paths) and update dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ exclude = [
     ".gitignore",
     ".travis.yml",
 ]
+edition = "2021"
 
 [package.metadata.docs.rs]
 # Disabled for now since this is borked; tracking https://github.com/rust-lang/docs.rs/issues/302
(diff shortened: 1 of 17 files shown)


ethyi (approved):
No changes to tests, logic, just better and more flexible versioning, especially for zeroize.

All tests pass.

Please merge, inflexible versioning leads to difficult versioning conflicts down the line, and Rust currently doesn't have the right tooling to work around it.

rickwebiii:
@ethyi I need someone to click "Approve and Run" as this is my first contribution here.

ethyi (commented):
It could be better to specify zeroize with "1.0.0" since it will be compatible with anything above "1.0", including "1.5.7". Same applies for other major release crates.

rickwebiii:
@ethyi I did some investigation as to why they were using zeroize >=1.0, <1.4 and it appears to be to maintain compatibility with Rust toolchain 1.41 for some downstream dependencies. I'm not sure if this is still a requirement or not, @isislovecruft could provide some insight as they made the change.

rozbb:

PR #412, opened 2022-10-17
Relax zeroize dependency

rozbb:
This PR:
* Relaxes `zeroize` to `^1`
* Improves documentation about this crate's MSRV
* Fixes the MSRV tests. Previously they were using the newest deps. But the MSRV should be calculated with respect to the oldest possible deps.

I would like to cut a release soon. According to the MSRV test, our v4.0 will not require an MSRV bump.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -105,6 +105,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    # First run `cargo +nightly -Z minimal-verisons check` in order to get a
+    # Cargo.lock with the oldest possible deps
     - uses: actions-rs/toolchain@v1
       with:
         profile: minimal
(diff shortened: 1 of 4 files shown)


rozbb:
Would it be better actually to do `cargo hack +MSRV build --powerset`? This is what the RustCrypto [`minimum-versions`](https://github.com/RustCrypto/actions/blob/0d6a4dcc99de309e48d7bec49b9e8870a0d29e1f/.github/workflows/minimal-versions.yml#L31) script does, which is pulled in in `zeroize`.

tarcieri:
I would still suggest bumping MSRV. Though the minimum versions will work, the default cargo resolution will not and will throw compile errors if the `zeroize` feature is enabled.

At least 1.56 introduces `rust-version` so it can surface a useful message about future MSRV incompatibilities.

tarcieri (approved):

PR #413, opened 2022-10-17
Updated to edition 2021

rozbb:
Not sure what the pro/cons are here. Any opinions?

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ '*' ]
   pull_request:
-    branches: [ main, develop ]
+    branches: [ '*' ]
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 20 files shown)


tarcieri:
@rozbb if you're MSRV 1.56 already, there are no downsides to bumping editions.

Bumping editions lets you remove `TryFrom`/`TryInto` imports, since they're already in the prelude. It looks like there's only one in `montgomery.rs` though.

rozbb:
I guess another benefit is that the modern `use crate::` is more familiar to me and likely others. That's enough to warrant the bump imo. I'll remove the `TryInto` and redundant `extern crate`s and it should be good to go.

rozbb:
Should be good now. See if you can't find anything overtly 2015ish. Had to do a few things by hand, since `cargo fix` didn't do much, and some doctests needed edits

tarcieri:
@rozbb one other thing it might be good to do is enable the `warn(rust_2018_idioms)` lint. It looks like there are a few implicit elided lifetimes.

You can add the following to CI to reject warnings there:

```yaml
env:
  RUSTFLAGS: "-D warnings"
```

Issue #392, opened 2022-05-02
Remove dependencies


xpepermint:
I believe that lower-level libraries such as this one should be dependency-free. This is especially true for libraries that provide "delicate" features such as encryption. I see lots of dependencies listed in the `Cargo.toml` file which IMHO aren't crucial and should be wrapped into optional features or even extracted into smaller specialized crates.

@isislovecruft @cathieyun

k0nserv:
per @piegamesde's comment I got rid of `byteorder` in https://github.com/dalek-cryptography/curve25519-dalek/pull/418

rozbb:
Fixed!

PR #418, opened 2022-10-21
Remove byteorder

k0nserv:
Instead of having a dependency on `byteorder` use methods from the
standard library(`{to,from}_le_bytes`).

This addresses the concern in #392

### DIFF
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -138,6 +138,7 @@
 
 use core::borrow::Borrow;
 use core::cmp::{Eq, PartialEq};
+use core::convert::TryInto;
 use core::fmt::Debug;
 use core::iter::{Product, Sum};
 use core::ops::Index;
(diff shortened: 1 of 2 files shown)


tarcieri (approved):

tarcieri:
@k0nserv actually I think you need to specify `release/4.0` as the merge target, rather than `main`? /cc @rozbb

k0nserv:
Ah yes, of course 😅

src/scalar.rs
```diff
@@ -1186,6 +1183,26 @@ impl UnpackedScalar {
...
+/// ## Panics
+/// Panics if `src.len() != 8 * dst.len()`.
+fn read_le_u64_into(src: &[u8], dst: &mut [u64]) {
+    assert!(
+        src.len() == 8 * dst.len(),
+        "src.len() = {}, dst.len() = {}",
+        src.len(),
+        dst.len()
+    );
+    for (bytes, val) in src.chunks(core::mem::size_of::<u64>()).zip(dst.iter_mut()) {
```
rozbb, line 1197: Small thing: you can use `8` here instead of `core::mem::size-of::<u64>()`, since you use `8` in the `assert` above and the `expect` below.

k0nserv: Yeah I was thinking it was inconsistent now. Is there an argument for using `size_of` everywhere? It's more "correct" but I don't think there's ever a case(now or in the future) when it will return anything other than 8

rozbb: It's actually fixed at 8 according to the docs https://doc.rust-lang.org/std/mem/fn.size_of.html

But don't worry about it I'm about to push this change, plus 1 extra unit test, and then I'll merge :)

k0nserv: Aight, Thanks 🙏

rozbb:
Thank you for this PR!

PR #419, opened 2022-10-22
Fix AVX2 and AVX-512 builds

rozbb:
It looks like #413 broke the build for AVX2 and AVX-512 backends by not sprinkling enough `crate::` prefixes around. This PR fixes the build. I tested it on a machine with both CPU features available.

It's not great that this breakage was able to happen. As far as I know, though, Github Actions runners do not have AVX2. I'll just have to manually run these tests locally until we have another solution.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -26,13 +26,19 @@ jobs:
     - uses: dtolnay/rust-toolchain@stable
     - run: cargo test --no-default-features --features "std u64_backend"
 
-  test-simd:
-    name: Test simd backend (nightly)
+  build-simd:
+    name: Build simd backend (nightly)
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
     - uses: dtolnay/rust-toolchain@nightly
-    - run: cargo test --no-default-features --features "std simd_backend"
+    # Build with AVX2 features, then with AVX512 features
+    - env:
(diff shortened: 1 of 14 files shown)


tarcieri:
@rozbb AVX2 should work on GH Actions runners. They use DS2V2 Azure VMs:

https://learn.microsoft.com/en-us/azure/virtual-machines/dv2-dsv2-series

> Dv2-series sizes run on the 3rd Generation Intel® Xeon® Platinum 8370C (Ice Lake), Intel® Xeon® Platinum 8272CL (Cascade Lake), Intel® Xeon® 8171M 2.1GHz (Skylake), or the the Intel® Xeon® E5-2673 v4 2.3 GHz (Broadwell), or the Intel® Xeon® E5-2673 v3 2.4 GHz (Haswell) processors with Intel Turbo Boost Technology 2.0.

So, Haswell and above, which are all AVX2.

However, no guarantees of AVX-512. It might be worth integrating CPU feature detection with fallback, and it could be opportunistically tested if the runner has it available.

tarcieri (approved):

rozbb:
Turns out you can `cargo build` with AVX2, so long as you don't try to run it. Better than nothing. I'll merge the other CI changes and add this to this PR. https://github.com/rozbb/curve25519-dalek/commit/2b9849b3879355e9b824d4d0f2c36aa9b7013560

rozbb:
Force push was due to rebase on #422

PR #420, opened 2022-10-23
Rustfmt

### LINKED ISSUE #416: Run rustfmt
A lot of this crate uses nonstandard formatting. If you edit files in an IDE with rustfmt-on-save, then you end up getting diffs that are way bigger than necessary. This crate should either have a `rustfmt.toml` or use the default formatting rules, and then apply them crate-wide.

robjtede:
closes #416 

reviewer note: this is much easier to review [with "hide whitespace changes" enabled](https://github.com/dalek-cryptography/curve25519-dalek/pull/420/files?w=1) to skip over indentation-only changes


tarcieri:
This codebase makes extensive use of stylized whitespace which I think in many places assists in understanding how the code works. So I think the hardest part of a good rustfmt PR is identifying these places and annotating them with `#[rustfmt::skip]` or thereabouts.

I can point out a few of them.

src/backend/serial/u64/scalar.rs
```diff
@@ -193,53 +193,47 @@ impl Scalar52 {
...
-        z[8] =                                                             m(a[4],b[4]);
+        z[0] = m(a[0], b[0]);
+        z[1] = m(a[0], b[1]) + m(a[1], b[0]);
+        z[2] = m(a[0], b[2]) + m(a[1], b[1]) + m(a[2], b[0]);
+        z[3] = m(a[0], b[3]) + m(a[1], b[2]) + m(a[2], b[1]) + m(a[3], b[0]);
+        z[4] = m(a[0], b[4]) + m(a[1], b[3]) + m(a[2], b[2]) + m(a[3], b[1]) + m(a[4], b[0]);
+        z[5] = m(a[1], b[4]) + m(a[2], b[3]) + m(a[3], b[2]) + m(a[4], b[1]);
+        z[6] = m(a[2], b[4]) + m(a[3], b[3]) + m(a[4], b[2]);
+        z[7] = m(a[3], b[4]) + m(a[4], b[3]);
+        z[8] = m(a[4], b[4]);
```
tarcieri, line 207: Here is a location where the whitespace assists in understanding how the code works.

The "stair-stepping" approach shows how the indices line up, which is absent from the rustfmt'd version (and looks quite a bit like the original `ref10` C implementation.

This is a detail I think it would be a pity to lose.

src/backend/vector/avx2/field.rs
```diff
@@ -539,13 +536,18 @@ impl FieldElement2625x4 {
...
+        carry(&mut z, 5);
+        carry(&mut z, 2);
+        carry(&mut z, 6);
+        carry(&mut z, 3);
+        carry(&mut z, 7);
         // Since z[3] < 2^64, c < 2^(64-25) = 2^39,
         // so    z[4] < 2^26 + 2^39 < 2^39.0002
-        carry(&mut z, 4); carry(&mut z, 8);
+        carry(&mut z, 4);
+        carry(&mut z, 8);
```
tarcieri, line 550: The original formatting illustrates how two halves of the carry chain are being computed in parallel, and it's very clear which half is which prior to rustfmt, which loses that information.

src/backend/vector/avx2/field.rs
```diff
@@ -618,31 +620,55 @@ impl FieldElement2625x4 {
...
+            + m(x8, x8_19)
+            + ((m(x7_2, x9_19)) << 1);
+        let mut z7 = m(x0_2, x7) + m(x1_2, x6) + m(x2_2, x5) + m(x3_2, x4) + ((m(x8, x9_19)) << 1);
+        let mut z8 = m(x0_2, x8)
+            + m(x1_2, x7_2)
+            + m(x2_2, x6)
+            + m(x3_2, x5_2)
+            + m(x4, x4)
+            + ((m(x9, x9_19)) << 1);
+        let mut z9 = m(x0_2, x9) + m(x1_2, x8) + m(x2_2, x7) + m(x3_2, x6) + m(x4_2, x5);
```
tarcieri, line 671: This looks awful compared to the original formatting.

src/backend/vector/avx2/field.rs
```diff
@@ -821,16 +848,106 @@ impl<'a, 'b> Mul<&'b FieldElement2625x4> for &'a FieldElement2625x4 {
...
+        let z9 = m(x0, y9)
+            + m(x1, y8)
+            + m(x2, y7)
+            + m(x3, y6)
+            + m(x4, y5)
+            + m(x5, y4)
+            + m(x6, y3)
+            + m(x7, y2)
+            + m(x8, y1)
+            + m(x9, y0);
```
tarcieri, line 950: The original formatting was much better here.

src/field.rs
```diff
@@ -124,26 +124,26 @@ impl FieldElement {
...
+        let t10 = t9.pow2k(20); // 39..20
+        let t11 = &t10 * &t9; // 39..0
+        let t12 = t11.pow2k(10); // 49..10
+        let t13 = &t12 * &t7; // 49..0
+        let t14 = t13.pow2k(50); // 99..50
+        let t15 = &t14 * &t13; // 99..0
+        let t16 = t15.pow2k(100); // 199..100
+        let t17 = &t16 * &t15; // 199..0
+        let t18 = t17.pow2k(50); // 249..50
+        let t19 = &t18 * &t13; // 249..0
```
tarcieri, line 146: I prefer the original formatting here

src/scalar.rs
```diff
@@ -1163,32 +1176,32 @@ impl UnpackedScalar {
...
+        square_multiply(&mut y, 3 + 2, &_11);
+        square_multiply(&mut y, 3 + 2, &_11);
+        square_multiply(&mut y, 1 + 4, &_1001);
+        square_multiply(&mut y, 1 + 3, &_111);
+        square_multiply(&mut y, 2 + 4, &_1111);
+        square_multiply(&mut y, 1 + 4, &_1011);
+        square_multiply(&mut y, 3, &_101);
+        square_multiply(&mut y, 2 + 4, &_1111);
+        square_multiply(&mut y, 3, &_101);
+        square_multiply(&mut y, 1 + 2, &_11);
```
tarcieri, line 1204: Another case where the original formatting was debatably better

src/backend/serial/curve_models/mod.rs
```diff
@@ -373,20 +373,21 @@ impl CompletedPoint {
 
 impl ProjectivePoint {
     /// Double this point: return self + self
-    pub fn double(&self) -> CompletedPoint { // Double()
-        let XX          = self.X.square();
-        let YY          = self.Y.square();
-        let ZZ2         = self.Z.square2();
-        let X_plus_Y    = &self.X + &self.Y;
+    pub fn double(&self) -> CompletedPoint {
+        // Double()
```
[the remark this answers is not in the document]
robjtede: there's a couple others too

robjtede:
@tarcieri combed through and think i got all the stair-stepping and comment alignments covered now

src/backend/serial/u32/field.rs
```diff
@@ -339,28 +344,33 @@ impl FieldElement2625 {
...
+        carry(&mut z, 5);
+        carry(&mut z, 2);
+        carry(&mut z, 6);
+        carry(&mut z, 3);
+        carry(&mut z, 7);
         // Since z[3] < 2^64, c < 2^(64-25) = 2^39,
         // so    z[4] < 2^26 + 2^39 < 2^39.0002
-        carry(&mut z, 4); carry(&mut z, 8);
+        carry(&mut z, 4);
+        carry(&mut z, 8);
```
tarcieri, line 368: Would be nice to do something to preserve this still

src/backend/serial/curve_models/mod.rs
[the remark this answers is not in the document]
tarcieri: Not sure offhand

tarcieri (approved):

tarcieri (changes requested):
Hmm, this seemed to cause a number of test failures.

@robjtede can you split this into two PRs, one that adds the `#[rustfmt::skip]` annotations, and one that runs `rustfmt`? That would make it easier to review, as the latter would be 100% automated changes.

robjtede:
So... after rebasing onto `release/4.0`, which I expect is the one that will be yaknow released as 4.0, it seems that a blind rustfmt run has already been done on the repo in commit 5758b8cce17ab64c09f9ea2751e41997e440bf30. Meaning this PR is turning into "restore the original stair-stepping etc formatting.

tarcieri:
@robjtede aha, so it is. Well awesome, thanks for doing that.

robjtede:
Err okay... wasn't as bad as I initially expected. Ready for another review I think.

rozbb:
I'll second what Tony said: could we have two commits? One where you add `rustfmt::skip`, and one where you run rustfmt? There can be a third commit with all the small spacing changes for the skipped functions too. The main thing is I don't trust myself to review this much code for this sensitive a crate. What I'd rather do is just run `rustfmt` and compare the diff to the next commit's diff.

robjtede:
@tarcieri @rozbb split the changes up how you asked

tarcieri (approved):
Looks good to me now although I haven't personally tried to reproduce 96e2d4513dcb3deb7c41cef6db99d1eef5f102cf locally

rozbb:
Thank you for this again! I wasn't sure last week if we had asked for too much from you so I had gone and done the splitting myself in #424, using a lot of your `[rustfmt::skip]` points. I added a few more, and did some formatting changes slightly differently. Since it's a particularly sensitive crate, I also documented the steps to verify that no semantic changes were made. I also credited you in the commit message.

I'm gonna close this PR unless you think I missed something important. Again, thank you!

PR #422, opened 2022-10-26
Update GitHub actions

nox:
actions/checkout@v2 uses Node 12 which will be deprecated by Summer 2023.

Actions from actions-rs also use Node 12 but they are unmaintained, so actions-rs/toolchain is replaced by dtolnay/rust-toolchain and actions-rs/cargo is replaced by manual cargo commands.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,118 +14,62 @@ jobs:
     name: Test u32 backend
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions-rs/toolchain@v1
-      with:
-        profile: minimal
-        toolchain: stable
-        override: true
-    - uses: actions-rs/cargo@v1
-      with:
-        command: test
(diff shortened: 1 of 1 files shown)


ptondereau (approved):

rozbb:
Looks good to me!

cc @tarcieri . After this I'm gonna add extra CI runs to #419

.github/workflows/rust.yml
```diff
@@ -14,118 +14,62 @@ jobs:
...
-      with:
-        profile: minimal
-        toolchain: stable
-        override: true
-    - uses: actions-rs/cargo@v1
-      with:
-        command: test
-        args: --no-default-features --features "std u32_backend"
+    - uses: actions/checkout@v3
+    - uses: dtolnay/rust-toolchain@stable
```
tarcieri, line 18: Whoa, neat

tarcieri (approved):
Re: bumping things like `actions/checkout@2`, it might be worth considering adding a Dependabot config

Issue #388, opened 2022-04-13
Increase MSRV to allow for Zeroize > 1.4


hariria:
Hi, is there any chance we can increase the minimum supported rust version and zeroize to > 1.4? It conflicts with a lot of other crates. Thanks!

Based on what i'm seeing here: https://github.com/Devolutions/picky-rs/issues/140#issuecomment-1093662134

samuelvanderwaal:
I, too, am curious about this.

rcolomina:
What is the problem? Why are setting <1.4 restriction on Zeroize?

k0nserv:
We are hitting this in the webrtc-dtls crate too. We have both `elliptic-curve` and `curve25519-dalek` in our tree. elliptic-curve@v0.12.3 requires `zeroize > 1.5` and `curve25519-dalek` requires `zeroize < 1.4`

robjtede:
Hitting a similar issue to @k0nserv at work with the aws-sdk crates which require zeroize ^1.4.

It's very common in the Cargo ecosystem for MSRV bumps to exist in minor versions. If users of the crate wish to remain on an outdated compiler, they are able to solve MSRV problems themselves by pinning deps (even deep ones) in their own application's manifest.

juliotpaez:
Same problem here with aws-sdk crates and ethereum web3 crates

brson:
Just linking to the issue that established this pinned version: https://github.com/dalek-cryptography/curve25519-dalek/issues/362

rozbb:
This is now resolved in the `release/4.0` branch. Next release (or prerelease) will have a laxer zeroize dependency.

k0nserv:
@rozbb When can we expect a release here? I think for many folks this problem remains unresolved until a release is made(pointing to GitHub will not work well in libraries I think)

tarcieri:
@k0nserv see open items under https://github.com/dalek-cryptography/curve25519-dalek/issues/416

PR #425, opened 2022-11-08
Loosen zeroize semver

drozdziak1:
Cargo does not allow conflicting versions that are semver compatible. I have a dependency that wants `zeroize ^1.5` and the hard limit on `<1.4` causes such conflict. Is there a reason why `1.4` is frowned upon?


tarcieri:
Dup of #388

drozdziak1:
This is clearly not a duplicate of the issue and the reason why this bump is not possible is never mentioned in the issue.

tarcieri:
The issue is already fixed on the `release/4.0` branch: https://github.com/dalek-cryptography/curve25519-dalek/blob/release/4.0/Cargo.toml#L53

drozdziak1:
That makes sense. Thanks for the context!

Issue #414, opened 2022-10-18
Simplifying backend selection


tarcieri:
`curve25519-dalek` currently defines the following backends, modeled as crate features:

- `u32_backend`
- `u64_backend`
- `fiat_u32_backend`
- `fiat_u64_backend`
- `simd_backend`
- `avx2_backend`

First, `avx2_backend` can be removed: it's a deprecated legacy alias for `simd_backend`.

Second, I think `u32_backend` and `u64_backend` could be selected automatically by gating on e.g. `cfg(target_pointer_width = "64")`. If this selection were automatic, I think these features could be removed completely, with `u32_backend`/`u64_backend` selected and used by default unless another backend has been explicitly enabled.

Likewise, `fiat_u64_backend` and `fiat_u32_backend` can be collapsed down to `fiat_backend` using similar `target_pointer_width`-based gating.

If all of the above were done, the list above could be simplified down to just `fiat_backend` and `simd_backend`.

nox:
Any chance to make the crate use additive features?

tarcieri:
The best way to do that would probably be to get rid of feature-gating for `simd_backend`, and instead use runtime autodetection to choose when to enable it.

ryankurte:
> If this selection were automatic, I think these features could be removed completely, with u32_backend/u64_backend selected and used by default unless another backend has been explicitly enabled.

so long as there's still an override / way to disable the detection! sometimes it's useful to be able to run tests for 32-bit platforms under x86_64 with the u32_backend, and where target detection is involved, `cpuid` tends to break in miri and enclave builds...

> Any chance to make the crate use additive features?

i was wondering whether y'all had considered refactoring the backends to traits w/ implementations so you could build the library with any / all enabled, then default based on runtime detection or availability? similar to the way allocators are implemented now. i looked at starting / PRing this briefly but, it would be a lot of work without knowing it'd be useful / desirable.

tarcieri:
> so long as there's still an override / way to disable the detection! sometimes it's useful to be able to run tests for 32-bit platforms under x86_64 with the u32_backend

FWIW, we used to have explicit features for this in the https://github.com/RustCrypto/elliptic-curves crates, but got rid of the because you can test 32-bit code on 64-bit Linux by compiling for `--target i686-unknown-linux-gnu`, or using `cross`.

> and where target detection is involved, cpuid tends to break in miri and enclave builds...

[...]

PR #428, opened 2022-11-11
Implement simplified backend selection

tarcieri:
As proposed in #414, this commit changes the backend selection approach, introspecting `target_pointer_width` to select `u32_backend` vs `u64_backend` (or `fiat_u32_backend`/`fiat_u64_backend` if the `fiat_backend` feature is enabled).

This helps eliminate the use of non-additive features, and also the rather confusing errors that happen if multiple backends are selected (i.e. thousands of lines of rustc errors).

The selection logic checks if `target_pointer_width = "64"` and uses the 64-bit backend, or falls back to the 32-bit backend otherwise. This means the crate will always have a valid backend regardless of the pointer width, although there may be odd edge cases for exotic platforms which would optimally use the 64-bit backend but have a non-"64" target pointer width for whatever reason. We can handle those cases as they come up.


tarcieri:
Note: I deliberately punted on trying to change `fiat_backend` and `simd_backend` into a `cfg` attribute to keep this PR relatively simple

Cargo.toml
```diff
@@ -43,6 +43,7 @@ name = "dalek_benchmarks"
 harness = false
 
 [dependencies]
+cfg-if = "1"
```
tarcieri, line 46: Note this is already a transitive hard dependency via the `sha2` crate which uses it in a similar capacity.

It's maintained by the core team, so nothing to worry about as far as supply chain attacks go.

README.md
```diff
@@ -99,18 +99,6 @@ Curve arithmetic is implemented using one of the following backends:
...
-```sh
-cargo build --no-default-features --features "std u32_backend"
-cargo build --no-default-features --features "std u64_backend"
-# Requires nightly, RUSTFLAGS="-C target_feature=+avx2" to use avx2
-cargo build --no-default-features --features "std simd_backend"
-# Requires nightly, RUSTFLAGS="-C target_feature=+avx512ifma" to use ifma
-cargo build --no-default-features --features "std simd_backend"
-```
-Crates using `curve25519-dalek` can either select a backend on behalf of their
-users, or expose feature flags that control the `curve25519-dalek` backend.
```
tarcieri, line 112: Could probably use a description of the new backend selection logic here.

README.md
```diff
@@ -166,8 +154,7 @@ compiled with appropriate `target_feature`s, so this cannot occur.
 Benchmarks are run using [`criterion.rs`][criterion]:
 
 ```sh
-cargo bench --no-default-features --features "std u32_backend"
-cargo bench --no-default-features --features "std u64_backend"
+cargo bench --no-default-features --features std
 # Uses avx2 or ifma only if compiled for an appropriate target.
 export RUSTFLAGS="-C target_cpu=native"
 cargo bench --no-default-features --features "std simd_backend"
```
rozbb, line 160: `cargo bench --features "simd_backend"` will suffice

src/lib.rs
```diff
@@ -105,18 +105,6 @@
 //! * an `avx2` backend using [parallel formulas][parallel_doc] and `avx2` instructions (sets speed records);
```
rozbb, line 105: All this is killed by #429

rozbb (changes requested):
Looking good! And nice changes to the CI. Left some small comments here and there. After that, and a conflict resolution, this is good to go.

README.md
tarcieri: FYI, running that I notice that backend no longer compiles, but it's unrelated to this PR.

README.md
```diff
@@ -99,18 +99,6 @@ Curve arithmetic is implemented using one of the following backends:
 * an `avx2` backend using [parallel formulas][parallel_doc] and `avx2` instructions (sets speed records);
 * an `ifma` backend using [parallel formulas][parallel_doc] and `ifma` instructions (sets speed records);
```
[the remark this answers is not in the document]
tarcieri: Added in f0bff96

src/lib.rs
tarcieri: Removed in rebase

README.md
tarcieri: Added in f0bff96

README.md
```diff
@@ -166,8 +154,7 @@ compiled with appropriate `target_feature`s, so this cannot occur.
 Benchmarks are run using [`criterion.rs`][criterion]:
 
 ```sh
-cargo bench --no-default-features --features "std u32_backend"
-cargo bench --no-default-features --features "std u64_backend"
+cargo bench --no-default-features --features std
```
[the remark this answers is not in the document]
tarcieri: Fixed in f0bff96

README.md
tarcieri: Fixed in f0bff96

PR #433, opened 2022-11-13
Fix `simd_backend` + `--no-default-features`

tarcieri:
...and one other `alloc` cleanup.

`simd_backend` requires `alloc` and so this combination previously caused a compile error.

Also one additional test that works with an `alloc`-only profile wasn't enabled, causing a warning in this case.

Finally, changes CI to test this configuration.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,7 +27,8 @@
 // External dependencies:
 //------------------------------------------------------------------------
 
-#[cfg(all(feature = "alloc", not(feature = "std")))]
+#[cfg(feature = "alloc")]
+#[allow(unused_imports)]
 #[macro_use]
 extern crate alloc;
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,4 +63,4 @@ alloc = ["zeroize/alloc"]
 # fiat-crypto backend with formally-verified field arithmetic
 fiat_backend = ["fiat-crypto"]
 # The SIMD backend uses parallel formulas, using either AVX2 or AVX512-IFMA.
-simd_backend = ["nightly", "packed_simd"]
+simd_backend = ["alloc", "nightly", "packed_simd"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,4 +63,4 @@ alloc = ["zeroize/alloc"]
 # fiat-crypto backend with formally-verified field arithmetic
 fiat_backend = ["fiat-crypto"]
 # The SIMD backend uses parallel formulas, using either AVX2 or AVX512-IFMA.
-simd_backend = ["alloc", "nightly", "packed_simd"]
+simd_backend = ["nightly", "packed_simd"]
--- a/src/edwards.rs
+++ b/src/edwards.rs
@@ -1389,6 +1389,7 @@ mod test {
     }
 
     #[test]
+    #[cfg(feature = "alloc")]
     fn impl_sum() {
         // Test that sum works for non-empty iterators
         let BASE = constants::ED25519_BASEPOINT_POINT;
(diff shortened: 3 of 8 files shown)


tarcieri:
Well that's fun...

https://github.com/dalek-cryptography/curve25519-dalek/actions/runs/3456308070/jobs/5768944920

```
error: failed to run custom build command for `typenum v1.15.0`

Caused by:
  process didn't exit successfully: `/home/runner/work/curve25519-dalek/curve25519-dalek/target/debug/build/typenum-e88f9767296c4856/build-script-main` (signal: 4, SIGILL: illegal instruction)
  --- stdout
  cargo:rustc-env=TYPENUM_BUILD_CONSTS=/home/runner/work/curve25519-dalek/curve25519-dalek/target/debug/build/typenum-11848c0af93[14](https://github.com/dalek-cryptography/curve25519-dalek/actions/runs/3456308070/jobs/5768944920#step:5:15)bcc/out/consts.rs
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.
```

It seems the `RUSTFLAGS` to enable the SIMD target features are being applied to a build script, which is then executed in CI, and dies due to SIGILL.

I'm not sure if there's a way to selectively enable RUSTFLAGS only for the target executable and not for the build scripts.

rozbb:
Ugggh I've fought this before. I don't think there's a way to do crate-specific RUSTFLAGS, but I might be wrong.

Is the only alternative the feature flag route?

tarcieri:
@rozbb my best guess is it's a change in a recent nightly which is auto-vectorizing something which wasn't being vectorized before. See #434 as a trivial code change which is also breaking in CI now.

I tried pinning the nightly version in #435 and that seems to provide a band-aid until we can figure out a better solution.

tarcieri:
@rozbb this is green now

src/lib.rs
```diff
@@ -27,7 +27,8 @@
 // External dependencies:
 //------------------------------------------------------------------------
 
-#[cfg(all(feature = "alloc", not(feature = "std")))]
+#[cfg(feature = "alloc")]
```
rozbb, line 30: Not for this PR, but shouldn't we have a thing where `std` is sufficient to use allocating functions? I've seen the `crate::prelude::Vec` pattern before for this

rozbb:
Actually, it seems that only tests require `alloc`. `cargo +nightly build --no-default-features --features "simd_backend"` works for me

rozbb:
> Finally, changes CI to test this configuration.

I don't see the changes anymore

src/lib.rs
tarcieri: `std` just re-exports what's in `alloc`, and the `std` crate feature also activates `alloc`.

This means that types defined in `alloc` are always available via `alloc`, regardless of whether or not the `std` feature is enabled.

The only historical reason to go through convolutions to access those types was when `alloc` wasn't stable.

tarcieri:
> Actually, it seems that only tests require alloc. cargo +nightly build --no-default-features --features "simd_backend" works for me

Aah hmmm, ok. Maybe the tests should be gated on `alloc`.

> I don't see the changes anymore

Was lost in a rebase unfortunately

src/lib.rs
rozbb: Ah ok that's good

rozbb:
If it's a pain to feature-gate the tests, I don't think it matters that much. I don't think there's an AVX2-capable machine that also has no RAM

tarcieri:
Gated the tests in `a94629c`.

> If it's a pain to feature-gate the tests, I don't think it matters that much. I don't think there's an AVX2-capable machine that also has no RAM

I'm a little curious why this only pops up with `simd_backend` enabled. But also, I think it's generally good to ensure all combinations of features work unless you explicitly bail out with `compile_error!` for invalid combinations.

PR #435, opened 2022-11-13
CI: pin nightly version to 2022-11-01

tarcieri:
A pinned nightly may potentially work around the build failures we're experiencing due to build scripts using SIMD features:

https://github.com/dalek-cryptography/curve25519-dalek/pull/433#issuecomment-1312787275

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,7 +34,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: dtolnay/rust-toolchain@nightly
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: nightly-2022-11-01
     # Build with AVX2 features, then with AVX512 features
     - env:
         RUSTFLAGS: "-C target_feature=+avx2"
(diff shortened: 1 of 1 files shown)


tarcieri:
This seems to have recently popped up, so it seems that LLVM might be auto-vectorizing some code in the build script it previously didn't used to.

Hopefully pinning the version provides a band-aid and also keeps the build reproducibly green.

rozbb:
I can't seem to replicate the build error on my x86-64 machine. On `release/4.0` I'm running `cargo +nightly test --features "simd_backend"` Should that produce an error?

tarcieri:
It appears to be dying when these `RUSTFLAGS` are used:

```
RUSTFLAGS="-C target_feature=+avx512ifma"
```

tarcieri:
Hmm, it seems while this nightly version "fixed" the issue in CI, it's still failing locally for me

rozbb:
Same, even after a `cargo clean`

tarcieri:
Oddly enough, enabling the `std` feature seems to fix it.

rozbb:
Not for me 😕

tarcieri:
Might've found a workaround: https://github.com/dalek-cryptography/curve25519-dalek/pull/436

tarcieri:
I think #436 is working. Closing this.

PR #436, opened 2022-11-14
CI: fix `build-simd` job

tarcieri:
The `RUSTFLAGS` were getting applied to build scripts, which caused them to crash with SIGILL.

According to this issue, RUSTFLAGS won't be applied to build scripts when cross-compiling by passing the `--target` attribute:

https://github.com/rust-lang/cargo/issues/4423

This attempts to work around the problem by explicitly passing:

    --target x86_64-unknown-linux-gnu

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -38,10 +38,10 @@ jobs:
     # Build with AVX2 features, then with AVX512 features
     - env:
         RUSTFLAGS: "-C target_feature=+avx2"
-      run: cargo build --no-default-features --features "std simd_backend"
+      run: cargo build --target x86_64-unknown-linux-gnu --features simd_backend
     - env:
         RUSTFLAGS: "-C target_feature=+avx512ifma"
-      run: cargo build --no-default-features --features "std simd_backend"
+      run: cargo build --target x86_64-unknown-linux-gnu --features simd_backend
 
   test-defaults-serde:
     name: Test default feature selection and serde


tarcieri:
Hmm, why aren't the CI jobs running? 😕

rozbb:
This fix works for me, but it kinda sucks that the build is broken this bad. Do you know if anyone upstream knows about this? I guess we can just document in the README that this is how you have to compile with SIMD for now (since it already requires a few steps)

tarcieri:
@rozbb I think a less brittle and more permanent fix would be to use runtime CPU feature detection rather than `RUSTFLAGS`. It would even be possible to run the tests with `simd_backend` enabled in that case

rozbb:
Don't you need the `RUSTFLAGS` part for codegen? If we want runtime detection, then every build needs to include the AVX2/AVX512 code, no?

tarcieri:
For ideal codegen you need `RUSTFLAGS`, but you can still achieve baseline functionality without it.

Instead you can annotate functions with `#[target_feature(enable = "avx...")]` and LLVM will still inline across those with the assumption the given features are available.

Such functions must also be annotated `unsafe` and have the potential to SIGILL if you don't do runtime feature detection.

rozbb:
Ah I see. I assumed the `RUSTFLAGS` were necessary but I suppose they're not. Just generate the code for _everything_ if `simd_backend` is enabled. It's only a redundancy of 1 backend, at most, so not bad. Is that right?

tarcieri:
Yeah, you include code for all backends for a particular target architecture, and detect which are available at runtime.

That eliminates the need for most users to enable `RUSTFLAGS` to take advantage of the backend, and also makes the binary relocatable.

PR #438, opened 2022-11-21
Deprecate `EdwardsPoint::hash_from_bytes`

### LINKED ISSUE #400: Misleading description for hash-to-Edwards point
The function `EdwardsPoint::hash_from_bytes()` is described as "performing hashing to the group" and explicitly references draft-irtf-cfrg-hash-to-curve:

https://github.com/dalek-cryptography/curve25519-dalek/blob/8abb22bcafe30ac2dbad372d0444fdba9e63bc0e/src/edwards.rs#L529-L532

This is a rather misleading description because:

  - This function does _not_ implement the hash-to-curve process of draft-irtf-cfrg-hash-to-curve: the hash from input to a field element is different, and the handling of the sign bit is also different.
  - Only a single Elligator2 map is applied, yielding something that [draft-irtf-cfrg-hash-to-curve](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-16#section-3) calls `encode_to_curve`, not `hash_to_curve` (the draft reserves the latter name for the construction that invokes the map twice, and adds together the two output points, to yield a quasi-uniform output distribution).

The first point is mostly about interoperability; while the map is not the same as in the draft, it is still safe (in particular, this code just uses the first 255 bits of a hash output as a field element, while the draft would generate 129 extra bits and perform a modular reduction; but in Curve25519, the field modulus is very close to 2^255, so the bias from using just 255 input bits is negligible). The second point is, arguably, more important for security, because the current `hash_from_bytes()` produces points with a non-uniform distribution that is easy to distinguish from a uniform random generation. Most protocols (and in particular the Signal stuff that this implementation follows) don't mind, but some of them require a uniform distribution, and the lack of uniformity can have consequences ranging from invalid security proofs to actual information leaks.

I recommend, at least, modifying the documentation to make it explicit that the implemented functionality is not compatible with draft-irtf-cfrg-hash-to-curve, and that it has a non-uniform output distribution, which may be troublesome for some (admittedly not many) protocols.

Looking at the existing pull requests, I see PR #377, which not only modifies the description of `hash_from_bytes()`, but also implements (in new, additional functions) the draft-irtf-hash-to-curve process. I have not looked in details at that PR, but the idea seems good.

rozbb:
Fixes #400 

Changes:
* Renamed `EdwardsPoint::hash_from_bytes` to `EdwardsPoint::hash_and_encode`, and deprecates it
* Added test vectors for the Ristretto one-way map function

Unresolved:
* I think 4.1 should remove `hash_and_encode` entirely. It's not necessarily wrong (or at least I haven't tested it on known-answers yet), but that doesn't mean it should be exposed.
* I would like to add `EdwardsPoint::hash_from_bytes` back in the future, as the real Elligator2 impl. The question is whether or not that would cause breakage for someone moving from 3.x to 4.1. Part of the answer might be in knowing who uses this function, but I don't know the answer to that, and don't really know how to find out
* Neither the `EdwardsPoint` nor the `RistrettoPoint` (nor the `Scalar` for that matter) version of `hash_from_bytes` do any domain separation. This is not strictly required for `RistrettoPoint` ("should" rather than "MUST"), but it is [strictly required](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-16.html#section-3.1) for `EdwardsPoint`. So  I guess it's fine to leave `RistrettoPoint` be, and just do it correctly in the future for `EdwardsPoint`? Feels icky.

cc @pornin if you have opinions on the rename or use of the functions


pornin:
As usual I commented on the wrong issue...

The ristretto code is fine; it follows the ristretto255 draft spec. Domain-separation would be added into the process that produces the 64 bytes on which `RistrettoPoint::hash_from_bytes()` feeds; this part is not implemented by curve25519-dalek, and arguably has no need to be, so I would say that there is nothing to change here (though adding test vector checks, as this PR does, is a nice touch).

For `EdwardsPoint::hash_from_bytes()`, the important part, in my opinion, is to have correct documentation, i.e. stating explicitly that the function does not actually follow the referenced Elligator2 map (it does something similar but subtly different), and that it is not "a hash" in that its output is not uniform. Renaming and/or deprecating the function is a different issue. Renaming it will break things for whoever may be calling that function, which is in general undesirable, but might be actually the point here (you'd like to know whether that function is actually used anywhere). If the name is changed then I think that the new name should not contain the word "hash" (to avoid conveying the idea that it has a uniform output, as is normally expected from "hash functions"); maybe "map_to_point"?

rozbb:
@tarcieri ready for review. Only thing that bothers me is what to name the new, correctly implemented function once it's written.

Looks like the last remaining `do-for-4.0` after this is docs-related. We're close!

tarcieri (approved):

PR #443, opened 2022-11-28
Add SemVer Policy for Public API

### LINKED ISSUE #442: SemVer policy for pre-1.0 dependencies
Something I think should be documented prior to the 4.0 release is the SemVer policy for pre-1.0 dependencies which are part of the public API, namely `digest` and `rand_core`.

I would suggest exempting these APIs from SemVer guarantees around major versions, allowing them to be updated with a minor version bump only.

It might be worth considering making both of those dependencies optional and noting if they're activated, you're opting into something which isn't stable around major versions. That would mean the feature profile one receives with the default features activated *is* stable around major versions.

pinkforest:
Closes #442 

This currently does not consider that `digest` / `rand_core` are optional features as suggested.

Will change the doc here if the status of those dependencies changes for 4.0.0.


README.md
```diff
@@ -125,11 +125,33 @@ cross-compiling will work on an x86-64 Linux machine:
...
+| 4.x      | Dependencies `digest` and `rand_core` | Minor SemVer bump   |
+
+* In the future exempted `digest` and `rand_core` may be changed to optional features.
+* Any changes to these exemptions will be accompanied by a major version bump.
+
+## SemVer
+
+* All on-by-default features of this library are covered by SemVer
+* MSRV is considered exempt from SemVer as noted
+* Breaking changes on dependencies exposed via Public API are also considered exempt as noted
```
tarcieri, line 154: Maybe open with this?

README.md
```diff
@@ -125,11 +125,33 @@ cross-compiling will work on an x86-64 Linux machine:
...
+## Public API SemVer Exemptions
+
+Breaking changes on SemVer exempted components affecting the Public API will also be accompanied by;
+
+| Releases | Public API Component(s)               | Policy              |
+| :---     | :---                                  | :---                |
+| 4.x      | Dependencies `digest` and `rand_core` | Minor SemVer bump   |
+
+* In the future exempted `digest` and `rand_core` may be changed to optional features.
+* Any changes to these exemptions will be accompanied by a major version bump.
```
tarcieri, line 148: I think it would be OK to stabilize the associated version of `digest` and/or `rand_core` if they drop a 1.0 release without bumping `curve25519-dalek` to 5.0, e.g. hypothetically "all versions of `curve25519-dalek` v4.3.0 and newer use `rand_core` v1"

README.md
pinkforest: Simplied a bit

README.md
pinkforest: Ok. Simplified this too.

rozbb:
I'm a bit torn. On one hand it makes sense to follow strict semver on the default state of our repo. But on the other, `random()` methods seem like too essential a thing to put behind a feature gate. I am leaning towards the strict semver option though.

pinkforest:
Subsumed by #453 :+1:

PR #444, opened 2022-11-30
Minor documentation fixes

pinkforest:
- Also fixup the documentation re: internal docs


rozbb:
Could we directly use the katex jsdelivr repo rather than our own? The [tutorial](https://katex.org/docs/browser.html) seems to do it

PR #445, opened 2022-11-30
Move katex docs assets to repo

pinkforest:
- Moves the required katex assets to repo so these can be included via #444


rozbb:
Decided not to vendor katex in #444

PR #447, opened 2022-11-30
Make `rand_core` optional

pinkforest:
As proposed in #442 moves `rand_core` as optional feature

This still gets used via dev-dependencies and depending on feature selection tests uses API impl by proxy or mock.


src/ristretto.rs
```diff
@@ -113,7 +113,7 @@
 //! used to implement
 //!
 //! * `RistrettoPoint::random()`, which generates random points from an
-//! RNG;
+//! RNG - enabled by rand_core feature;
```
rozbb, line 116: Typsetting: should be `rand_core`

src/lib.rs
```diff
@@ -83,3 +83,6 @@ pub(crate) mod prelude;
 
 // Generic code for window lookups
 pub(crate) mod window;
+
+#[cfg(test)]
+pub mod mocks;
```
rozbb, line 88: I think this can be `pub(crate)`

src/mocks.rs
```diff
@@ -0,0 +1,42 @@
+//! This is used mocking / proxying the below for tests:
```
rozbb, line 1: This is great!

Cargo.toml
```diff
@@ -36,14 +36,15 @@ bincode = "1"
 criterion = { version = "0.4.0", features = ["html_reports"] }
 hex = "0.4.2"
 rand = "0.8"
+rand_core = { version = "0.6", default-features = false }
```
burdges, line 39: I suppose not being default is justified by saying most dependencies already specify their own usages of the lower level methods.

rozbb:
This is nearly ready. An annoying issue is that `rand_core` is actually enabled by default. This is because `std` enables `rand_core/std`. @tarcieri  I don't think there's any way around this other than:
1. Kill `rand_core/std` and don't let anyone use it. The thing we lose here is [only](https://github.com/rust-random/rand/tree/7aa25d577e2df84a5156f824077bb7f6bdf28d97/rand_core#crate-features) `impl<R: RngCore> RngCore for Box<R>`.
2. Make `rand_core` a default dependency, breaking our opt-in semver-breakage policy

I think (1) is the right choice. I don't think anyone will miss it.

rozbb:
@tarcieri thoughts on killing `rand_core/std`? If we do this, we get the extra benefit of killing `std` entirely actually.

tarcieri:
AFAIK nothing in the crate itself actually needs or uses `rand_core/std`.

It provides two things:

- transitive activation of `rand_core/getrandom` feature which provides `OsRng`
- `std::error::Error` impl on `rand_core::Error`, which is not found in `curve25519-dalek`'s public API

`OsRng` is used for testing in various places. It seems we can activate `rand_core/getrandom` in `dev-dependencies` and call it a day.

tarcieri:
#458 removes activation of `rand_core/std`

Issue #448, opened 2022-12-02
Scalar::bits


burdges:
We're mildly worried about the [`Scalar::bits`](https://github.com/dalek-cryptography/curve25519-dalek/blob/main/src/scalar.rs#L825) method.  It's always used with another loop in https://github.com/dalek-cryptography/curve25519-dalek/blob/main/src/montgomery.rs#L319-L322 so maybe the compiler unifies the loops, and removes the intermediate 256 byte array, and maybe this even helps behind the scenes optimizations, but..

A priori it feels risky to do this, like did anyone check if the compiler leaves secret key material on the stack?   An iterator should unify the loops for sure.

rozbb:
Resolved in #451

PR #451, opened 2022-12-05
Made Scalar::bits return an iterator rather than an array

rozbb:
This should address #448 

cc @tarcieri 
cc @burdges

### DIFF
--- a/src/montgomery.rs
+++ b/src/montgomery.rs
@@ -335,17 +335,21 @@ impl<'a, 'b> Mul<&'b Scalar> for &'a MontgomeryPoint {
             W: FieldElement::one(),
         };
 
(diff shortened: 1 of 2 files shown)


tarcieri (approved):

rozbb:
Sorry for the botther. I added an explicit `zeroize()` statement for the last bit on the stack. It seems like overkill (frankly I don't know how often we care about leaking stack data) but if you think it's okay I'll merge and move on.

burdges:
Yeah I'm not sure what the original code built to, just looked odd to me.  :)  I didn't know if it originated from SIMD thing or what, or just the original code style.  Anyways I guess you've regression tests so you'll see what happens here.

PR #453, opened 2022-12-07
Updates to README

rozbb:
[README Rendered](https://github.com/dalek-cryptography/curve25519-dalek/blob/4a687c9f8fd2549f7aa0f882e291038b23366069/README.md)

[CHANGELOG Rendered](https://github.com/dalek-cryptography/curve25519-dalek/blob/4a687c9f8fd2549f7aa0f882e291038b23366069/CHANGELOG.md)

This subsumes #443 (will add @pinkforest as coauthor). This:

* Reorders some things in the README
* Explains semver policy
* Specifies feature flags and backends more explicitly
* Removes `nightly` from the CI bc that didn't belong there
* Adds @pinkforest to thankyou list (@pinkforest plz tell me if there is a specfic name you want there)

Issue #449, opened 2022-12-02
u32_backend is slower than u64_backend on armv7 Android


jrose-signal:
I noticed the upcoming 4.0 release has removed the u32_backend and u64_backend features in favor of checking the target pointer size, which seems very sensible.* We know the u32_backend can be drastically faster than the u64_backend in some configurations:

> On my Xperia 10 with SailfishOS (armv7hl user space on aarch64 kernel, yes I know, rustflags = "-C target-feature=+v7,+neon"):
> - u32_backend + std: [644.25 us 646.59 us 649.28 us]
> - u64_backend + std: [6.4522 ms 6.4685 ms 6.4880 ms]

…but not all:

> On a Cortex-M3 (Texas Instruments CC2538):
> - u32_backend without alloc: 470507 cycles
> - u32_backend with alloc (wut): 878507 cycles
> - u64_backend without alloc: 494078 cycles
> - u64_backend with alloc: 470502 cycles (the fastest)

(https://github.com/signalapp/libsignal/issues/453#issuecomment-1068184262)

Today I finally got around to running some 32-bit microbenchmarks on a *new* Android phone (with help from a coworker), a Pixel 6a, which I feel should nonetheless still produce interesting results because, well, running armv7 code on an aarch64 device can't magically use 64-bit registers. We consistently found that the u32_backend was 5-15% *slower* than the u64_backend when compiling for that particular armv7.

_EDIT: Later, I tested on a 32-bit Moto X and got approximately the same results._

All of this is differences of at most a tenth of a millisecond on any particular operation (we tested key agreement and signing + verifying), so it's not like it's going to make or break the crate. But it did make me think twice about "u32_backend is for 32-bit CPUs, u64_backend is for 64-bit CPUs". Is that worth restoring feature flags to *override* the default inference?

\* Especially given that the previous config made it easy for a downstream crate to depend on u64_backend implicitly as a default feature, thus removing the chance for the final program to choose the u32_backend instead.

jrose-signal:
Hypothesis about @rubdos' drastic Xperia results: the compiler is doing a terrible job vectorizing the u64 backend for armv7 (target feature `+neon`). We don't get that because some 32-bit Android phones really are that low-end.

tarcieri:
> running armv7 code on an aarch64 device can't magically use 64-bit registers. We consistently found that the u32_backend was 5-15% slower than the u64_backend when compiling for that particular armv7.

[...]

PR #454, opened 2022-12-07
Add target u32/u64 backend overrides

pinkforest:
Related before-PR-doc (rendered): https://github.com/dalek-cryptography/curve25519-dalek/pull/453#discussion_r1042294155

Thou shall ask: https://github.com/dalek-cryptography/curve25519-dalek/issues/449#issuecomment-1340670829

Thou shall be giveth:

# cfg
Test cfg matrix / expectation

All selected backends are mutually exclusive.

## fiat_backend 
Auto select `fiat_u32` / `fiat_u64` based on `cfg(target_pointer_width)` - default `fiat_u32` on non-32/64 bits

## serial (default)
Auto select `u32` / `u64` based on `cfg(target_pointer_width)` - default `u32` on non-32/64 bits

## fiat_backend w/ `cfg(curve25519_dalek_bits = "32")`
Selects `fiat_u32` regardless of target_pointer on fiat_backend

## fiat_backend w/ `cfg(curve25519_dalek_bits = "64")`
Selects `fiat_u64` regardless of target_pointer

## serial (default) w/ `cfg(curve25519_dalek_bits = "32")`
Selects `u32` (serial) regardless of target_pointer

## serial (default) w/ `cfg(curve25519_dalek_bits = "64")`
Selects `u64` (serial) regardless of target_pointer

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,15 @@
+//! This selects the curve25519_dalek_bits either by default from target_pointer_width or explicitly set
+
+fn main() {
+    #[cfg(any(
+        all(not(target_pointer_width = "64"), not(curve25519_dalek_bits = "64")),
+        curve25519_dalek_bits = "32"
+    ))]
+    println!("cargo:rustc-cfg=curve25519_dalek_bits=\"32\"");
+
+    #[cfg(any(
+        all(target_pointer_width = "64", not(curve25519_dalek_bits = "32")),
+        curve25519_dalek_bits = "64"
+    ))]
+    println!("cargo:rustc-cfg=curve25519_dalek_bits=\"64\"");
+}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -115,4 +115,9 @@ jobs:
     - uses: actions/checkout@v3
     - uses: dtolnay/rust-toolchain@stable
     # This filter selects no benchmarks, so we don't run any, only build them.
-    - run: cargo bench "nonexistentbenchmark"
+    - name: Build u32 bench
+      run: env RUSTFLAGS="--cfg curve25519_dalek_bits=\"32\"" cargo bench "nonexistentbenchmark"
+    - name: Build u64 bench
+      run: env RUSTFLAGS="--cfg curve25519_dalek_bits=\"64\"" cargo bench "nonexistentbenchmark"
+    - name: Build default (host native) bench
+      run: cargo bench "nonexistentbenchmark"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ major series.
 
 ## 4.x series
 
+* Add target u32/u64 backend overrides
 * Migrate documentation to docs.rs hosted
 * Fix backend documentation generation
 * Deprecate `EdwardsPoint::hash_from_bytes` and rename it `EdwardsPoint::nonspect_map_to_curve`
--- a/src/backend/serial/mod.rs
+++ b/src/backend/serial/mod.rs
@@ -21,18 +21,23 @@
 use cfg_if::cfg_if;
 
 cfg_if! {
+
     if #[cfg(feature = "fiat_backend")] {
-        #[cfg(not(target_pointer_width = "64"))]
+
+        #[cfg(curve25519_dalek_bits = "32")]
         pub mod fiat_u32;
 
-        #[cfg(target_pointer_width = "64")]
+        #[cfg(curve25519_dalek_bits = "64")]
         pub mod fiat_u64;
+
     } else {
-        #[cfg(not(target_pointer_width = "64"))]
+
+        #[cfg(curve25519_dalek_bits = "32")]
         pub mod u32;
 
-        #[cfg(target_pointer_width = "64")]
+        #[cfg(curve25519_dalek_bits = "64")]
         pub mod u64;
+
     }
 }
 
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -38,14 +38,14 @@ use crate::scalar::Scalar;
 
 cfg_if! {
     if #[cfg(feature = "fiat_backend")] {
-        #[cfg(not(target_pointer_width = "64"))]
+        #[cfg(curve25519_dalek_bits = "32")]
         pub use crate::backend::serial::fiat_u32::constants::*;
-        #[cfg(target_pointer_width = "64")]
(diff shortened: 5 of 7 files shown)


tarcieri:
FWIW I just pushed up a PR to switch to `cfg`-based backend selection: #455

pinkforest:
thx - I'll just send you a .patch then or re-base - my initial attempt was overly naive I see now :)

tarcieri:
@pinkforest I'm missing how these are "overrides"... it seems like they would be mandatory?

pinkforest:
> @pinkforest I'm missing how these are "overrides"... it seems like they would be mandatory?

It's in the lib.rs with global cfg_attr that sets it if it's not set - for some reason criterion doesn't set it ? unless I'm mistaken.

src/lib.rs
```diff
@@ -22,6 +22,23 @@
 )]
 #![doc(html_root_url = "https://docs.rs/curve25519-dalek/4.0.0-pre.2")]
 #![doc = include_str!("../README.md")]
+//------------------------------------------------------------------------
+// curve25519-dalek arithmetric overrides
+//------------------------------------------------------------------------
+#![cfg_attr(
```
pinkforest, line 28: @tarcieri this has the global default and override mechanism

tarcieri: Aha!

src/lib.rs
```diff
@@ -22,6 +22,23 @@
...
+    ),
+    cfg(dalek_bits = "32")
+)]
+#![cfg_attr(
+    any(
+        all(target_pointer_width = "64", not(dalek_bits = "32")),
+        dalek_bits = "64"
+    ),
+    cfg(dalek_bits = "64")
+)]
```
tarcieri, line 41: Perhaps this could be addressed in a followup PR, but it seems there are a few `target_arch`es wwhere `target_pointer_width = "32"` where we might want to override the default 32-backend, namely `wasm`

src/lib.rs
pinkforest: Oh you're right for some reason this doesn't set the cfg() here :see_no_evil:

pinkforest:
Ok I moved the thing into via build.rs and now the default detection w/ overrides works properly :tada:

src/lib.rs
pinkforest: This default / override logic is now in build.rs as this does not work via cfg_attr - naive / tired I again was :woman_shrugging: 

Nonetheless yeah definitely it would be nice to map those overrides out

pinkforest:
> One nit from https://github.com/dalek-cryptography/curve25519-dalek/pull/455 was whether the cfg names should be curve25519_dalek_* instead of just dalek_*

Will be waiting direction on this and rename after :+1:

tarcieri:
I think `curve25519_dalek_*` is fine and nicely unambiguous, if a bit long.

But people will probably be pasting examples of the `cfg` directives from the docs anyway, so it doesn't really matter

pinkforest:
Problem with that may be the ergonomics -

If there is dalek "family of crate"s and and one wants to switch `dalek_bits` it for all dalek inter-dependency crates this then requires remembering to add cfg flags for all of them separately ?

I mean if we have similar switch for other dalek crates then it might be feasible to just have dalek_* for some things - 

.. or at least for `dalek_bits` considering backend is mostly `curve25519_dalek_backend` specific thing ?

... maybe both `dalek_bits` and `curve25519_dalek_bits` where `curve25519_dalek_bits` overrides `dalek_bits` within curve25519-dalek crate but otherwise it is `dalek_bits` on other dalek family of crates - if there is a use-case for it ?

tarcieri:
The "family of crates" is a good point. Many users will use `curve25519-dalek` transitively through `ed25519-dalek` or `x25519-dalek`, where `curve25519_*` is an implementation detail.

pinkforest:
Déjà vu 574217694e952ad5193f273a541c633ea32207c2 :sweat_smile:
```
commit 574217694e952ad5193f273a541c633ea32207c2
Date:   Fri Oct 11 17:10:57 2019 -0700

    Remove build.rs.
    
    This was more useful at the time when we were determining, e.g., optimal lookup
    table sizes and could regenerate them more easily, but it came at a massive
    complexity cost.  It also meant that we were unable to implement backend
    autoselection.  This commit removes the `build.rs` entirely.  In the future, a
    different `build.rs` could be added that auto-selects a backend, but it seems
    like the current default-u64 setup has been working fine.

commit 4dc8073330f63fb23040f7aa7dd0159aac7a9923 (tag: 1.2.3)
```

`git diff 4dc8073330f63fb23040f7aa7dd0159aac7a9923 574217694e952ad5193f273a541c633ea32207c2 build.rs`

build.rs
```diff
@@ -0,0 +1,15 @@
+//! This selects the dalek_bits either by default from target_pointer_width or explicitly set
```
pinkforest, line 1: Déjà vu 574217694e952ad5193f273a541c633ea32207c2 :sweat_smile:
```
commit 574217694e952ad5193f273a541c633ea32207c2
Date:   Fri Oct 11 17:10:57 2019 -0700

    Remove build.rs.
    
    This was more useful at the time when we were determining, e.g., optimal lookup
    table sizes and could regenerate them more easily, but it came at a massive
    complexity cost.  It also meant that we were unable to implement backend
    autoselection.  This commit removes the `build.rs` entirely.  In the future, a
    different `build.rs` could be added that auto-selects a backend, but it seems
    like the current default-u64 setup has been working fine.

commit 4dc8073330f63fb23040f7aa7dd0159aac7a9923 (tag: 1.2.3)
```

`git diff 4dc8073330f63fb23040f7aa7dd0159aac7a9923 574217694e952ad5193f273a541c633ea32207c2 build.rs`

rozbb:
This looks great! I think we can safely just use `curve25519_dalek_bits` for now.

If we want to add an umbrella cfg later, we can do that as a non-breaking change. The converse isn't true: if we make a separately low-level library then `dalek_bits` will either become an umbrella (breaking the package-specific use for `curve-`), or remains specific (making the naming bad).

Are we good to merge?

pinkforest:
> Are we good to merge?

:+1: I've re-named it to `curve25519_dalek_bits` - this should be good to merge now

src/lib.rs
pinkforest: Raised an issue: https://github.com/dalek-cryptography/curve25519-dalek/issues/456

tarcieri (approved):

PR #458, opened 2022-12-08
Remove activation of `rand_core/std`

tarcieri:
This feature provides two things:

1. Transitive activation of the `rand_core/getrandom` feature
2. Impl of `std::error::Error` for `rand_core::Error`

`rand_core::Error` is not part of `curve25519-dalek`'s public API and is irrelevant.

The `rand_core/getrandom` feature provides `OsRng` which is used in tests only, so we can activate it in `dev-dependencies`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ bincode = "1"
 criterion = { version = "0.4.0", features = ["html_reports"] }
 hex = "0.4.2"
 rand = "0.8"
+rand_core = { version = "0.6", features = ["getrandom"] }
 
 [[bench]]
 name = "dalek_benchmarks"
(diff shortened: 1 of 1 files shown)


rozbb:
`rand_core/std` does one thing: `impl<R: RngCore> RngCore for Box<R>`. Theoretically a user might care about this trait.

But if we remove `rand_core/std` anyways, then we can remove `std` entirely, since `subtle/std` does nothing.

rozbb:
Actually that's wrong. The README says that, but in reality you [only need alloc](https://github.com/rust-random/rand/blob/0dddc2c5593fae7dc473b9f1d5774f1caa5e61c2/rand_core/src/lib.rs#L457). So I guess we can remove std altogether

tarcieri:
Oh hey, you're right. Let me open a PR based on this one that removes the `std` feature entirely

PR #459, opened 2022-12-08
Remove `std` feature

tarcieri:
All of the existing usages of `std` can be replaced with `alloc`.

They are legacy usages from before when liballoc was stabilized.


rozbb:
This is odd. The benches now build, but `cargo bench` doesn't perform any benchmarks anymore 🤔

Cargo.toml
```diff
@@ -56,8 +57,7 @@ fiat-crypto = { version = "0.1.6", optional = true}
 
 [features]
 nightly = ["subtle/nightly"]
```
rozbb, line 59: Unrelated, but I thought `nightly` was removed at some point? The flag does nothing since subtle v2.2.

rozbb: nvm that's in my README PR

Cargo.toml
```diff
@@ -36,11 +36,12 @@ bincode = "1"
 criterion = { version = "0.4.0", features = ["html_reports"] }
 hex = "0.4.2"
 rand = "0.8"
-rand_core = { version = "0.6", default-features = false }
+rand_core = { version = "0.6", default-features = false, features = ["getrandom"] }
 
 [[bench]]
 name = "dalek_benchmarks"
 harness = false
+required-features = ["rand_core"]
```
tarcieri, line 44: @rozbb this is why it doesn't run benchmarks by default.

It's unrelated to this PR but I was getting a [build failure](https://github.com/dalek-cryptography/curve25519-dalek/actions/runs/3651302919/jobs/6168360815) without these changes.

Seems related to #447?

Cargo.toml
tarcieri: Can remove it in a followup.

Cargo.toml
rozbb: #447 doesn't cause this in itself. `cargo bench` runs benches currently on `release/4.0`. Hmm

tarcieri: Previously `std` was activated by default, and `std` activated `rand_core/std`.

The benchmarks [clearly do require `rand_core`](https://github.com/dalek-cryptography/curve25519-dalek/actions/runs/3651302919/jobs/6168360815), it's just it used to be "accidentally" enabled by default.

rozbb: Ah ok. Then I'll just make a note in the README for now

PR #460, opened 2022-12-08
Remove mocks

rozbb:
If you gate `Scalar::random()` and `RistrettoPoint::random()` on `any(cfg(test), feature = "rand_core")`, then you don't need mocks anymore.


tarcieri (approved):
Oh nice, I just touched these in #461 and was wondering what they were for, finding few uses. Seems you came to the same conclusion!

tarcieri:
@rozbb if you want to merge this first I can rebase #461

PR #461, opened 2022-12-08
Fix `--no-default-features` test warnings; consolidate CI jobs

tarcieri:
Previously `cargo test --no-default-features` would succeed but with warnings.

This commit fixes all of those warnings and tests `--no-default-features` in CI to ensure that in perpetuity.

It moves the `tests-default-serde` into the `test` job which reduces the total number of job runners needed and reuses intermediate artifacts which should improve build times.

It removes `test-alloc-u32` as this case is already covered by the `test` job when it runs `cargo test --target i686-unknown-linux-gnu` as the `alloc` feature alone is now enabled by default.


rozbb (approved):

tarcieri:
It’s `-D warnings` which I now see we don’t have enabled in CI. Let me go ahead and add that.

.github/workflows/rust.yml
```diff
@@ -8,6 +8,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  RUSTFLAGS: '-D warnings'
```
tarcieri, line 11: Added `-D warnings`.

FWIW ed25519-dalek already has it

tarcieri:
@rozbb added `-D warnings` if you want to re-approve / merge

PR #463, opened 2022-12-09
Allow the fiat serial backend work with the simd backend

### LINKED ISSUE #437: Allow `fiat_backend` and `simd_backend` to be activated at the same time
Right now `fiat_backend` and `simd_backend` are mutually exclusive: activating both crate features generates a compile error.

In #414 we discussed various alternatives including something like `--cfg dalek_backend="simd"`. However, since the backends map to optional crate dependencies, features seem like a more natural fit.

As an alternative to a compile error, another option would be to handle the case of both features being activated by choosing a "winner" backend which is selected when both options are enabled.

[In #414 I suggested that given various tradeoffs, it should probably be `fiat_backend`](https://github.com/dalek-cryptography/curve25519-dalek/issues/414#issuecomment-1316167153) because it would be bad to enable a non-verified backend when someone has asked for a formally verified one (even though `simd_backend` is superior in pretty much all other respects).

This isn't a blocker for a v4.0 release though, as it can easily be added afterward as it changes a case which is currently a compile error into one that works.

pinkforest:
- Fixes #437 
 
As discussed above it is feasible to allow the fiat serial backend to work with the simd backend.

This change makes it possible to use the fiat serial backend with the simd backend.

Introduces `cfg(curve25519_dalek_serial = "fiat")` to work alongside `cfg(curve25519_dalek_backend = "simd")`.

For ergonomics `cfg(curve25519_dalek_backend = "fiat")` work as-is like `curve25519_dalek_serial` alone.

It was also noted that the fiat serial backend requires MSRV 1.57.

The backend selection logic for this is now in build.rs.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,8 +34,8 @@ jobs:
           RUSTFLAGS: '--cfg curve25519_dalek_backend="fiat"'
         run: cargo test --target ${{ matrix.target }}
 
-  build-simd:
-    name: Build simd backend (nightly)
+  build-simd-default:
+    name: Build simd backend without fiat (nightly)
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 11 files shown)


pinkforest:
There is one test that fails blockign this: https://github.com/dalek-cryptography/curve25519-dalek/issues/464

pinkforest:
This is now ready :partying_face: 
Except just a blocking test on for Fiat with simd: 
 - https://github.com/dalek-cryptography/curve25519-dalek/issues/464

pinkforest:
Sure yeah, well I've tried that now FWIW :)

And as confused I can the only worthwhile course of action is for me to close this considering the above:)

PR #465, opened 2022-12-09
Fixes `curve25519_dalek_bits` defaults for cross and wasm

### LINKED ISSUE #456: Well known defaults for `curve25519_dalek_bits`
Origin issue:
 - https://github.com/dalek-cryptography/curve25519-dalek/issues/449#issuecomment-1340670829

Our MVP became to support "override" between `[fiat_]u32` and `[fiat_]u64` regardless of `cfg(target_pointer_width)`

We implemented the MVP - cfg(curve25519_dalek_bits = "32"|"64") w/ default set by `cfg(target_pointer_width)` here: 
 - https://github.com/dalek-cryptography/curve25519-dalek/pull/454

Meanwhile the default is based on `cfg(target_pointer_width)`:
 - `[fiat_]_u32` on 32 bit targets
 - `[fiat_]_u64` on 64 bit targets
 - `[fiat_]u_32` on all other targets

Question was raised to provide well known "target defaults" e.g. wasm family etc. : https://github.com/dalek-cryptography/curve25519-dalek/pull/454#discussion_r1042643559

Could we perhaps discuss / gather a list for MVP for this and then can perhaps throw PR at this ?

pinkforest:
As discussed in #456 ~~this sets well known defaults for:~~
 ~~`cfg(target_family = "wasm")`~~ and ~~`cfg(target_arch = "arm")`~~

~~By setting the 64 bit serial arithmetric via `cfg(curve25519_dalek_bits = "64")`~~

EDIT: I've used platforms crate via env TARGET to determine the actual cross target and this now fixes build.rs only.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -48,6 +48,33 @@ jobs:
         RUSTFLAGS: '--cfg curve25519_dalek_backend="simd" -C target_feature=+avx512ifma'
       run: cargo build --target x86_64-unknown-linux-gnu
 
+  cross:
+    strategy:
+      matrix:
+        include:
+          # ARM32
+          - target: armv7-unknown-linux-gnueabihf
+            rust: stable
+
+          # ARM64
+          - target: aarch64-unknown-linux-gnu
+            rust: stable
+
+          # PPC32
+          - target: powerpc-unknown-linux-gnu
+            rust: stable
+
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: ${{ matrix.deps }}
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.rust }}
+          targets: ${{ matrix.target }}
+      - run: cargo install cross
+      - run: cross test --release --target ${{ matrix.target }}
+
   nightly:
     name: Test nightly compiler
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ major series.
 
 #### Other changes
 
+* Set well known defaults for wasm and arm
 * Update Maintenance Policies for SemVer
 * Migrate documentation to docs.rs hosted
 * Fix backend documentation generation
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,9 @@ hex = "0.4.2"
 rand = "0.8"
 rand_core = { version = "0.6", default-features = false, features = ["getrandom"] }
 
+[build-dependencies]
+platforms = "3.0.2"
+
 [[bench]]
 name = "dalek_benchmarks"
 harness = false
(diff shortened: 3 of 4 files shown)


tarcieri (approved):

rozbb:
Awesome! One thing: could we have documentation somewhere (can even be in the comments of build.rs) that explains what's going on and why? A link to something about WASM benchmarks would be nice to include too if you have it

pinkforest:
I'm fixing the build script for cross.. this is WIP for an hour or so. sorry

tarcieri:
Sidebar: we should probably add some cross tests, particularly for big-endian architectures.

Here's an example of the ones we use for the @RustCrypto elliptic curve crates (although testing on both MSRV and stable is probably overkill):

https://github.com/RustCrypto/elliptic-curves/blob/0337e2a/.github/workflows/p256.yml#L89-L109

pinkforest:
K. I'll add those here

tarcieri:
I'm not super familiar with WASM benchmarking but this seems like an option:

https://docs.rs/easybench-wasm/latest/easybench_wasm/

@burdges you brought this up originally... any thoughts on how to do WASM benchmarks?

pinkforest:
Should we also build / test on windows and mac runners ?
I have a portable builder that uses diff runners https://github.com/rinse-repeat/rust-auditable-builder here
zig allows easy cross but then qemu could be used to test if we need (as you pointed elsewhere I think)

tarcieri:
@pinkforest that's something I think is helpful when using `std` APIs with different `std::os`-specific backends, but this is a purely computational crate now which doesn't link `std` at all so I think it's probably unnecessary in this case

pinkforest:
Yeah that's a very good point

pinkforest:
This is now documented here alongside other backend selection changes:
- https://github.com/dalek-cryptography/curve25519-dalek/pull/467

tarcieri:
It would be good to confirm via a benchmark that WASM benefits from this, which is something I can poke at this weekend unless someone else wants to.

pinkforest:
With criterion could have to wrap the wasm blob with a runtime - e .g. with lunatic -  system oriented - we do:
https://github.com/lunatic-solutions/lunatic/blob/main/benches/benchmark.rs

pinkforest:
For browser few pop up:
https://github.com/originjs/WASM-benchmark

https://github.com/PSPDFKit-labs/pspdfkit-webassembly-benchmark
that provides this:
http://iswebassemblyfastyet.com/

rozbb:
I think sticking with browser benches (easybench looks okay) is probably enough for now. For this crate, I'd guess  the most common place people will try to deploy their WASM is Chrome.

pinkforest:
FWIW - wasmer seems to benchmark sha1 crate via their runtime test
https://github.com/wasmerio/wasmer-bench/blob/master/benchmarks/src/sha.rs

pinkforest:
I've set this to only fix the build.rs for cross-compile and wasm targets for now as keen to get this merged.
This also gives the structure for well known targets -
Another follow-up PR perhaps can be instead used to comment out the armv7 and wasm potentially when we have more test / bench coverage on those overrides to justify the defaults.

burdges:
> you brought this up originally... any thoughts on how to do WASM benchmarks?

I donno, but I'll forward the question to people who do.

PR #467, opened 2022-12-09
Document `curve25519_dalek_` `backend` and `bits` selection

pinkforest:
[Rendered](https://github.com/pinkforest/curve25519-dalek/tree/docs-backend-cfgs)

Documenting
- #456  
- #465 
- #455 
- #454 

Note `fiat` cannot be used in combination with `simd` as I attempted :hankey: and found out :blush: 
- #463

I also thought it is best to document the cross-re-use between target backends between `avx` and `u64`

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,12 @@ rustflags = ['--cfg=curve25519_dalek_backend="BACKEND"']
 ```
 More info [here](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags).
 
-**NOTE:** The `simd` backend requires extra configuration. See [the SIMD
+The `simd` backend requires extra configuration. See [the SIMD
 section](#simd-target-backends).
+
+Note for contributors: The target backends are not entirely independent of each
+other. The `simd` backend directly depends on parts of the the `u64` backend to
+function.
 
 ## Word size for serial backends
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -188,10 +222,10 @@ compiled with appropriate `target_feature`s, so this cannot occur.
 Benchmarks are run using [`criterion.rs`][criterion]:
 
 ```sh
-cargo bench --features "alloc"
+cargo bench --features "rand_core"
 # Uses avx2 or ifma only if compiled for an appropriate target.
-export RUSTFLAGS="-C target_cpu=native"
-cargo +nightly bench --features "alloc simd_backend"
+export RUSTFLAGS='--cfg curve25519_dalek_backend="simd" -C target_cpu=native'
```
rozbb, line 227: Why is `-C target_cpu=native` in here? I remember looking into what this actually does, and it sets avx2 but it doesn't set avx512ifma even on a machine with that feature. What should the recommendation be?

tarcieri:
FYI: typo in the title of this PR: `curve24419_*` => `curve25519_*`

README.md
tarcieri: Previous benchmarking showed it provided the best optimizations at the cost of non-relocatability

rozbb: Hmm ok. Maybe we should recommend that as a "I don't know what to do" option for the simd backend. I can do that tn or tomorrow.

rozbb:
@pinkforest re mentioning that simd depends on u64: I agree it's an important detail but it felt distracting in the README, since it's more an impl detail. What do you think about relocating it to the top-level `mod vector` documentation?

pinkforest:
There was some confusion:
https://github.com/dalek-cryptography/curve25519-dalek/pull/455#issuecomment-1344615689

And this is also a question people will ask I think -

My motivation with that was to dispel confusion around current state that backends are separated and de-coupled where as they are really not. 

If we can point out the fact in some other way to ensure people understand how these beackends link to each other that would be nice maybe too ? :woman_shrugging:

rozbb:
Is the misconception something our crate users need to know or is it an implementation detail for contributors? I assumed it was the latter (and therefore can go in docs) but idk. If it's the former tho, I agree it should be in the README.

PR #470, opened 2022-12-11
Use inherent constants for `ZERO`, `ONE`, and `MINUS_ONE`

tarcieri:
For the field element types `FieldElement` and `Scalar`, use inherent constants instead of (non-const) functions to return these constant values.

It's likely the original functions predate support for inherent constants, but now that they're available, they're a better fit for these sort of constant values.


str4d (commented):
Post-hoc utACK

str4d:
The change should be document in the changelog though.

rozbb:
Was updated in https://github.com/dalek-cryptography/curve25519-dalek/commit/f38d677759cc69e6f389291ec99260e2ac6260d4

PR #471, opened 2022-12-11
[WIP] Impl `ff::Field` for `Scalar`

tarcieri:
Note: depends on #470 which needs to be merged first.

Adds https://github.com/zkcrypto/ff as an optional dependency, which provides traits for finite fields.

### DIFF
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -36,7 +36,7 @@
 //! ```
 //! use curve25519_dalek::scalar::Scalar;
 //!
-//! let one_as_bytes: [u8; 32] = Scalar::one().to_bytes();
+//! let one_as_bytes: [u8; 32] = Scalar::ONE.to_bytes();
 //! let a: Option<Scalar> = Scalar::from_canonical_bytes(one_as_bytes);
 //!
 //! assert!(a.is_some());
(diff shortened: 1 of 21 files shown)


src/scalar.rs
```diff
@@ -568,6 +571,36 @@ impl Zeroize for Scalar {
     }
 }
 
+#[cfg(all(feature = "ff", feature = "rand_core"))]
```
tarcieri, line 574: Would be nice to use namespaced features so an `ff` feature could activate `dep:ff` and `rand_core`

str4d: We will be using this in `ff` itself once its MSRV is 1.60 (https://github.com/zkcrypto/ff/issues/78), but I haven't seen a strong push for that MSRV yet in the wider ecosystem.

src/scalar.rs
```diff
@@ -568,7 +571,48 @@ impl Zeroize for Scalar {
...
+    fn double(&self) -> Self {
+        self + self
+    }
+
+    fn invert(&self) -> CtOption<Self> {
+        CtOption::new(self.invert(), Choice::from(1u8))
+    }
+
+    fn sqrt_ratio(_num: &Self, _div: &Self) -> (Choice, Self) {
+        // TODO: can't use `ff::helpers::sqrt_ratio_generic` because `Scalar` isn't a `PrimeField`
```
str4d, line 599: The documentation of `Scalar` indicates that its arithmetic is modulo `ell`, the prime order of both Ristretto and the Ed25519 basepoint. So we can `impl PrimeField for Scalar`.

tarcieri:
Closing in favor of #473

PR #472, opened 2022-12-12
Change `Scalar::from_canonical_bytes` to return `CtOption`

tarcieri:
~~*NOTE: includes #470 which should be merged first*~~ Merged!

This is helpful for implementing `ff::PrimeField::from_repr`.

Also changes `Scalar::is_canonical` to return `Choice`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ required-features = ["rand_core"]
 cfg-if = "1"
 rand_core = { version = "0.6.4", default-features = false, optional = true }
 digest = { version = "0.10", default-features = false, optional = true }
-subtle = { version = "^2.2.1", default-features = false }
+subtle = { version = "2.3.0", default-features = false }
 serde = { version = "1.0", default-features = false, optional = true, features = ["derive"] }
 zeroize = { version = "1", default-features = false }
 
@@ -65,3 +65,6 @@ packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_
 [features]
 default = ["alloc"]
 alloc = ["zeroize/alloc"]
+
+[profile.dev]
+opt-level = 2
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -37,7 +37,7 @@
 //! use curve25519_dalek::scalar::Scalar;
 //!
 //! let one_as_bytes: [u8; 32] = Scalar::ONE.to_bytes();
-//! let a: Option<Scalar> = Scalar::from_canonical_bytes(one_as_bytes);
+//! let a: Option<Scalar> = Scalar::from_canonical_bytes(one_as_bytes).into();
 //!
 //! assert!(a.is_some());
 //! ```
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -50,7 +50,7 @@ required-features = ["rand_core"]
 cfg-if = "1"
 rand_core = { version = "0.6.4", default-features = false, optional = true }
 digest = { version = "0.10", default-features = false, optional = true }
-subtle = { version = "^2.2.1", default-features = false }
+subtle = { version = "2.3.0", default-features = false }
```
tarcieri, line 53: Note: needed for `impl From<CtOption<T>> for Option<T>`

Cargo.toml
```diff
@@ -65,3 +65,6 @@ packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_
 [features]
 default = ["alloc"]
 alloc = ["zeroize/alloc"]
+
+[profile.dev]
+opt-level = 2
```
tarcieri, line 70: This makes the tests run 5 times faster for me with `cargo test`, with compile times effectively the same

src/scalar.rs
```diff
@@ -256,18 +257,10 @@ impl Scalar {
     /// - `Some(s)`, where `s` is the `Scalar` corresponding to `bytes`,
     ///   if `bytes` is a canonical byte representation;
     /// - `None` if `bytes` is not a canonical byte representation.
-    pub fn from_canonical_bytes(bytes: [u8; 32]) -> Option<Scalar> {
-        // Check that the high bit is not set
-        if (bytes[31] >> 7) != 0u8 {
-            return None;
-        }
+    pub fn from_canonical_bytes(bytes: [u8; 32]) -> CtOption<Scalar> {
```
str4d, line 260: This change should be documented in the changelog.

src/scalar.rs
```diff
@@ -1133,22 +1125,20 @@ impl Scalar {
...
+    /// assert!(! bool::from(_2_255_minus_1.is_canonical()));
     ///
     /// let reduced = _2_255_minus_1.reduce();
-    /// assert!(reduced.is_canonical());
+    /// assert!(bool::from(reduced.is_canonical()));
     /// # }
     /// ```
-    pub fn is_canonical(&self) -> bool {
-        *self == self.reduce()
+    pub fn is_canonical(&self) -> Choice {
```
str4d, line 1140: This change should be documented in the changelog.

str4d (approved):
I'm fine with these API changes, though they are not technically necessary to implement `ff::PrimeField` (we can inline these changed methods into that impl if API stability is preferred here).

src/scalar.rs
tarcieri: Added in c51ba24

src/scalar.rs
tarcieri: Added in c51ba24

tarcieri:
If we want to keep the `Scalar` APIs the same, it's easy enough to add this constant-time version in the `ff::PrimeField::from_repr` method, since it's 3 lines

elichai:
I opened a similar PR a few months ago: https://github.com/dalek-cryptography/curve25519-dalek/pull/384

tarcieri:
@elichai oh nice! maybe we can get this straightforward version in then you can rebase your PRs which add the performance improvements?

elichai (approved):
utACK, reviewed the changes and they seem correct to my understanding of the library

src/scalar.rs
```diff
@@ -1133,22 +1125,20 @@ impl Scalar {
...
+        self.ct_eq(&self.reduce())
```
rozbb, line 1141: Just making certain: is `reduce()` constant time itself?

tarcieri: It's supposed to be, and if it weren't, it would be a general security issue.

Issue #373, opened 2021-10-12
Implement elliptric-curve traits


daxpedda:
I propose implementing some of the traits defined in [elliptic-curve](https://crates.io/crates/elliptic-curve) to improve compatibility with downstream users.

Specifically [`Group`](https://docs.rs/elliptic-curve/0.10.6/elliptic_curve/trait.Group.html), [`GroupEncoding`](https://docs.rs/group/0.11.0/group/trait.GroupEncoding.html) and [`Field`](https://docs.rs/elliptic-curve/0.10.6/elliptic_curve/trait.Field.html) is what I had in mind.

I'm working on a PR as we speak.

kayabaNerve:
Is there any chance this will be done for EdwardsPoint as well (perhaps in another PR)? I don't believe group has the necessary functions to handle torsion, which would explain why it's not being done in that PR, yet it should at least be possible to provide group bindings with a main subgroup requirement. That may be a higher level decision than desired (see discussion on https://github.com/dalek-cryptography/curve25519-dalek/issues/380), yet my current workaround is my own library stubbing out parts of ff/group as needed (with a main subgroup check when decoding). I'd really love to be able to delete it in favor of this PR (or a relative), yet unfortunately do need to work with Ed25519.

Even if no one's willing to step up to the plate to handle this right now, I would be curious if such a PR would be accepted.

daxpedda:
I believe this can be closed now that https://github.com/dalek-cryptography/curve25519-dalek/pull/562 is merged?

rozbb:
Yup!

PR #473, opened 2022-12-12
Implement `ff` and `group` traits

str4d:
These are placed behind the new `group` feature flag.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ required-features = ["rand_core"]
 
 [dependencies]
 cfg-if = "1"
+ff_crate = { version = "0.13", default-features = false, optional = true, package = "ff" }
 rand_core = { version = "0.6.4", default-features = false, optional = true }
 digest = { version = "0.10", default-features = false, optional = true }
 subtle = { version = "2.3.0", default-features = false }
@@ -65,6 +66,7 @@ packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_
 [features]
 default = ["alloc"]
 alloc = ["zeroize/alloc"]
+ff = ["ff_crate", "rand_core"]
 
 [profile.dev]
 opt-level = 2
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ required-features = ["rand_core"]
 
 [dependencies]
 cfg-if = "1"
-ff_crate = { version = "0.13", default-features = false, optional = true, package = "ff" }
+group_crate = { version = "0.13", default-features = false, optional = true, package = "group" }
 rand_core = { version = "0.6.4", default-features = false, optional = true }
 digest = { version = "0.10", default-features = false, optional = true }
 subtle = { version = "2.3.0", default-features = false }
@@ -66,7 +66,7 @@ packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_
 [features]
 default = ["alloc"]
 alloc = ["zeroize/alloc"]
-ff = ["ff_crate", "rand_core"]
+group = ["group_crate", "rand_core"]
 
 [profile.dev]
 opt-level = 2
--- a/src/edwards.rs
+++ b/src/edwards.rs
@@ -195,23 +195,50 @@ impl CompressedEdwardsY {
     ///
     /// Returns `None` if the input is not the \\(y\\)-coordinate of a
     /// curve point.
-    #[rustfmt::skip] // keep alignment of explanatory comments
     pub fn decompress(&self) -> Option<EdwardsPoint> {
-        let Y = FieldElement::from_bytes(self.as_bytes());
+        let (is_valid_y_coord, X, Y, Z) = decompress::step_1(self);
+        if is_valid_y_coord.unwrap_u8() != 1u8 {
+            return None;
+        }
+        Some(decompress::step_2(self, X, Y, Z))
+    }
+}
+
+mod decompress {
+    use super::*;
+
+    #[rustfmt::skip] // keep alignment of explanatory comments
+    pub(super) fn step_1(
+        repr: &CompressedEdwardsY,
+    ) -> (Choice, FieldElement, FieldElement, FieldElement) {
+        let Y = FieldElement::from_bytes(repr.as_bytes());
         let Z = FieldElement::ONE;
         let YY = Y.square();
         let u = &YY - &Z;                            // u =  y²-1
         let v = &(&YY * &constants::EDWARDS_D) + &Z; // v = dy²+1
-        let (is_valid_y_coord, mut X) = FieldElement::sqrt_ratio_i(&u, &v);
+        let (is_valid_y_coord, X) = FieldElement::sqrt_ratio_i(&u, &v);
 
-        if is_valid_y_coord.unwrap_u8() != 1u8 { return None; }
+        (is_valid_y_coord, X, Y, Z)
+    }
 
+    #[rustfmt::skip]
+    pub(super) fn step_2(
+        repr: &CompressedEdwardsY,
+        mut X: FieldElement,
+        Y: FieldElement,
+        Z: FieldElement,
+    ) -> EdwardsPoint {
          // FieldElement::sqrt_ratio_i always returns the nonnegative square root,
          // so we negate according to the supplied sign bit.
-        let compressed_sign_bit = Choice::from(self.as_bytes()[31] >> 7);
+        let compressed_sign_bit = Choice::from(repr.as_bytes()[31] >> 7);
         X.conditional_negate(compressed_sign_bit);
 
-        Some(EdwardsPoint{ X, Y, Z, T: &X * &Y })
+        EdwardsPoint {
+            X,
+            Y,
+            Z,
+            T: &X * &Y,
+        }
     }
 }
 
@@ -1158,30 +1185,9 @@ impl GroupEncoding for EdwardsPoint {
     type Repr = [u8; 32];
 
     fn from_bytes(bytes: &Self::Repr) -> CtOption<Self> {
-        // NOTE: this is duplicated from CompressedEdwardsY::decompress due to the
-        // constant time requirement.
-
-        let Y = FieldElement::from_bytes(bytes);
-        let Z = FieldElement::ONE;
-        let YY = Y.square();
-        let u = &YY - &Z; // u =  y²-1
-        let v = &(&YY * &constants::EDWARDS_D) + &Z; // v = dy²+1
-        let (is_valid_y_coord, mut X) = FieldElement::sqrt_ratio_i(&u, &v);
-
-        // FieldElement::sqrt_ratio_i always returns the nonnegative square root,
-        // so we negate according to the supplied sign bit.
-        let compressed_sign_bit = Choice::from(bytes[31] >> 7);
-        X.conditional_negate(compressed_sign_bit);
-
-        CtOption::new(
-            EdwardsPoint {
-                X,
-                Y,
-                Z,
-                T: &X * &Y,
-            },
-            is_valid_y_coord,
-        )
+        let repr = CompressedEdwardsY(*bytes);
+        let (is_valid_y_coord, X, Y, Z) = decompress::step_1(&repr);
+        CtOption::new(decompress::step_2(&repr, X, Y, Z), is_valid_y_coord)
     }
 
     fn from_bytes_unchecked(bytes: &Self::Repr) -> CtOption<Self> {
--- a/src/edwards.rs
+++ b/src/edwards.rs
@@ -104,6 +104,13 @@ use core::ops::{Mul, MulAssign};
 #[cfg(feature = "digest")]
 use digest::{generic_array::typenum::U64, Digest};
 
+#[cfg(feature = "group")]
+use {
+    group::{cofactor::CofactorGroup, prime::PrimeGroup, GroupEncoding},
+    rand_core::RngCore,
+    subtle::CtOption,
+};
+
 use subtle::Choice;
 use subtle::ConditionallyNegatable;
 use subtle::ConditionallySelectable;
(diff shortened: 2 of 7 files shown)


str4d:
I'll rebase this.

str4d:
Rebased on `release/4.0`.

daxpedda:
Cc https://github.com/dalek-cryptography/curve25519-dalek/issues/373.

str4d:
I haven't added `group` trait impls in this PR because there are upcoming breaking changes to some of those traits that I'm hoping to make soon, and we can avoid needing to bring them in as part of a 5.0 if we wait until after that.

tarcieri:
@str4d you add `ff` and `group` to these SemVer exceptions, which allow for upgrades around minor versions rather than major versions: https://github.com/dalek-cryptography/curve25519-dalek#public-api-semver-exemptions

str4d:
If that's a strategy that the users of this crate (specifically the users of the `ff` trait impls) are happy with, then that's fine with me. It's definitely easier to set this up from the start, and doesn't affect anyone not enabling the `ff` feature. It does mean that `cargo update` doesn't work without specific care, so I'd like the README or somewhere else to give a bit of guidance on how crate users should handle these minor version bumps if they can't follow the update yet.

tarcieri:
@str4d if you had the full traits wired up, you could just have a `group` feature which pulls `ff` in as `group::ff`.

I expect most users who care about the `ff` impls also want `group` (here's [one anecdote](https://github.com/dalek-cryptography/curve25519-dalek/issues/373#issuecomment-1105778108)) and it would make for one fewer knob.

Re: making `cargo update` work, it's fine if you use this sort of SemVer predicate:

```toml
curve25519-dalek = ">= 5, < 5.1"
```

...which will still allow for patch-level updates, though users would need to keep an eye out for minor version updates.

These ranges can also be expanded once it's been verified that a minor version bump doesn't affect functionality a given downstream dependent crate happens to be using, e.g.

```toml
curve25519-dalek = ">= 5.1, < 5.4"
```

str4d:
> @str4d if you had the full traits wired up, you could just have a `group` feature which pulls `ff` in as `group::ff`.
>
> I expect most users who care about the `ff` impls also want `group` (here's https://github.com/dalek-cryptography/curve25519-dalek/issues/373#issuecomment-1105778108) and it would make for one fewer knob.

Given that `ff` would be optional, we get an implicit knob for it anyway (until MSRV is at least 1.60, at which point [the implicit feature for an optional crate can be disabled](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies)). So we may as well have it behave correctly (i.e. have it enable `rand_core`).

> ...which will still allow for patch-level updates, though users would need to keep an eye out for minor version updates.
>
> These ranges can also be expanded once it's been verified that a minor version bump doesn't affect functionality a given downstream dependent crate happens to be using

If I understand you correctly, the recommended workflow for using these "breaking changes in minor version bumps" feature flags is:
- When you enable the feature flag, bound your crate's dependency to `< X.{Y+1}` where `X.Y` is the latest release.
- When `X.{Y+1}.0` is released, check that it works for you, and then cut a backwards-compatible point release of your crate with the bound increased to `< X.{Y+2}`.

It does require proactivity, but it seems fine as long as it is documented.

tarcieri:
> Given that `ff` would be optional, we get an implicit knob for it anyway (until MSRV is at least 1.60, at which point [the implicit feature for an optional crate can be disabled](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies)). So we may as well have it behave correctly (i.e. have it enable `rand_core`).

What you're suggesting would work just fine with namespaced features in 1.60 if you did something like this:

```toml
[dependencies]
ff = { version = "0.13", optional = true }
group = { version = "0.13", optional = true }
rand_core = { version = "0.6", optional = true }

[features]
ff = ["dep:ff", "dep:rand_core"]
group = ["ff", "dep:group"]
```

However prior to 1.60 you can't rely on feature unification to activate other features in the same crate, i.e. if you *just* have this:

```toml
[dependencies]
ff = { version = "0.13", optional = true }
group = { version = "0.13", optional = true }
rand_core = { version = "0.6", optional = true }
```

Activating the `ff` feature will *not* automatically activate `rand_core` as a dependency, nor will activating the `group` feature automatically activate `ff` and `rand_core`, even though `ff` has `rand_core` as a dependency and `group` has `ff` as a dependency.

str4d:
I know, which is why the current PR uses the crate renaming trick to change the name of the implicit feature, so we can instead add an explicit feature that does have the correct setup. I was just arguing that we may as well preserve this because we're going to have an `ff` feature in the simple case. An alternative would be to use the renaming trick to "hide" the implicit feature and not add an explicit one, but that feels like busywork (as the renamed implicit feature is still an implicit feature).

str4d:
Oh wait, I'm not tired now and see what you meant. Accessing `ff` as `group::ff` means we don't need an explicit dependency on `ff`, which means we don't have the implicit feature flag. Sure, that works fine for me, and we can always add the `ff` feature flag in a backwards-compatible way if users say they want it.

str4d:
This PR now includes implementations of the `group` traits. A required `crate::edwards::SubgroupPoint` struct is introduced, to enable `impl CofactorGroup for EdwardsPoint`. I have not implemented `group::Curve` etc. because `curve25519-dalek` does not (currently) expose an affine representation of an Edwards point.

str4d:
[Force-pushed](https://github.com/dalek-cryptography/curve25519-dalek/compare/a8cc9ecd014c0c28bd6972fdbe1685c8286de3b9..0f25c70ceb55f23a3fff88d73cf6431cdd905696) to fix test imports.

src/edwards.rs
```diff
@@ -104,6 +104,13 @@ use core::ops::{Mul, MulAssign};
 #[cfg(feature = "digest")]
 use digest::{generic_array::typenum::U64, Digest};
 
+#[cfg(feature = "group")]
+use {
+    group_crate::{cofactor::CofactorGroup, prime::PrimeGroup, GroupEncoding},
```
tarcieri, line 109: If you add `extern crate group_crate as group`, you can `use group::` for imports.

It can have a `TODO` comment to remove that directive after a migration to namespaced features (which we might consider prior to a final 4.0 release, especially as it would be helpful for #481 as well)

src/edwards.rs
```diff
@@ -1107,6 +1114,339 @@ impl Debug for EdwardsPoint {
...
+        self.0 -= rhs.0;
+    }
+}
+
+#[cfg(feature = "group")]
+define_sub_assign_variants!(LHS = SubgroupPoint, RHS = SubgroupPoint);
+
+#[cfg(feature = "group")]
+impl<'b> SubAssign<&'b SubgroupPoint> for EdwardsPoint {
+    fn sub_assign(&mut self, rhs: &'b SubgroupPoint) {
```
tarcieri, line 1311: Lifetime elision should also work here

src/ristretto.rs
```diff
@@ -1129,6 +1136,133 @@ impl Debug for RistrettoPoint {
...
+    }
+}
+
+#[cfg(feature = "group")]
+impl GroupEncoding for RistrettoPoint {
+    type Repr = [u8; 32];
+
+    fn from_bytes(bytes: &Self::Repr) -> CtOption<Self> {
+        // NOTE: this is duplicated from CompressedRistretto::decompress due to the
+        // constant time requirement.
```
tarcieri, line 1179: Perhaps both could be composed in terms of a private free function?

str4d: I don't think so, because in the non-constant-time case you want the early returns.

str4d: Ah, but it can be done in terms of _two_ private free functions.

str4d:
[Force-pushed](https://github.com/dalek-cryptography/curve25519-dalek/compare/0f25c70ceb55f23a3fff88d73cf6431cdd905696..715e438e622c97d423ba78d8b1c04d4f2d9bd252) to address most of @tarcieri's comments. I'll add a follow-up commit with the de-duplication of decompression.

tarcieri (approved):

src/ristretto.rs
```diff
@@ -260,6 +267,27 @@ impl CompressedRistretto {
     ///
     /// - `None` if `self` was not the canonical encoding of a point.
     pub fn decompress(&self) -> Option<RistrettoPoint> {
+        let (s_encoding_is_canonical, s_is_negative, s) = decompress::step_1(self);
+
+        if s_encoding_is_canonical.unwrap_u8() == 0u8 || s_is_negative.unwrap_u8() == 1u8 {
```
rozbb, line 272: Same thing: `unwrap_u8 -> bool::from`

src/edwards.rs
```diff
@@ -188,23 +195,50 @@ impl CompressedEdwardsY {
     ///
     /// Returns `None` if the input is not the \\(y\\)-coordinate of a
     /// curve point.
-    #[rustfmt::skip] // keep alignment of explanatory comments
     pub fn decompress(&self) -> Option<EdwardsPoint> {
-        let Y = FieldElement::from_bytes(self.as_bytes());
+        let (is_valid_y_coord, X, Y, Z) = decompress::step_1(self);
+        if is_valid_y_coord.unwrap_u8() != 1u8 {
```
rozbb, line 200: I'd use the `bool::from` impl here rather than `unwrap_u8()`

src/scalar.rs
```diff
@@ -1214,6 +1220,126 @@ impl UnpackedScalar {
...
+
+    fn sqrt_ratio(num: &Self, div: &Self) -> (Choice, Self) {
+        group::ff::helpers::sqrt_ratio_generic(num, div)
+    }
+
+    fn sqrt(&self) -> CtOption<Self> {
+        group::ff::helpers::sqrt_tonelli_shanks(
+            self,
+            [
+                0xcb02_4c63_4b9e_ba7d,
```
rozbb, line 1255: What is this constant? Put a comment here

src/scalar.rs
```diff
@@ -1214,6 +1220,126 @@ impl UnpackedScalar {
...
+            0x00, 0x00, 0x00, 0x08,
+        ],
+    };
+    const MULTIPLICATIVE_GENERATOR: Self = Self {
+        bytes: [
+            2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0,
+        ],
+    };
+    const S: u32 = 2;
```
rozbb, line 1313: What is S?

src/scalar.rs
```diff
@@ -1214,6 +1220,126 @@ impl UnpackedScalar {
...
+        ],
+    };
+    const ROOT_OF_UNITY_INV: Self = Self {
+        bytes: [
+            0x19, 0xcc, 0x37, 0x71, 0x3a, 0xed, 0x8a, 0x99, 0xd7, 0x18, 0x29, 0x60, 0x8b, 0xa3,
+            0xee, 0x05, 0x86, 0x3d, 0x3e, 0x54, 0x9f, 0x92, 0xc2, 0x82, 0x18, 0x7e, 0x86, 0x1f,
+            0xef, 0x8c, 0xb5, 0x06,
+        ],
+    };
+    const DELTA: Self = Self {
```
rozbb, line 1328: What is Delta?

src/scalar.rs
```diff
@@ -1214,6 +1220,126 @@ impl UnpackedScalar {
...
+        ],
+    };
+    const MULTIPLICATIVE_GENERATOR: Self = Self {
+        bytes: [
+            2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0,
+        ],
+    };
+    const S: u32 = 2;
+    const ROOT_OF_UNITY: Self = Self {
```
rozbb, line 1314: Which root of unity is this?

src/scalar.rs
```diff
@@ -1214,6 +1220,126 @@ impl UnpackedScalar {
...
+    fn to_repr(&self) -> Self::Repr {
+        self.to_bytes()
+    }
+
+    fn is_odd(&self) -> Choice {
+        Choice::from(self.as_bytes()[0] & 1)
+    }
+
+    const MODULUS: &'static str =
+        "0x1000000000000000000000000000000014def9dea2f79cd65812631a5cf5d3ed";
```
rozbb, line 1296: Should cite this and/or add a test for it in the tests. This constant occurs in the `l()` function on page 52 of [RFC 8032](https://www.rfc-editor.org/rfc/rfc8032).

src/scalar.rs
```diff
@@ -1857,6 +1983,56 @@ mod test {
...
+        // ROOT_OF_UNITY^{2^s} mod m == 1
+        assert_eq!(
+            Scalar::ROOT_OF_UNITY.pow(&[1u64 << Scalar::S, 0, 0, 0]),
+            Scalar::ONE,
+        );
+
+        // DELTA^{t} mod m == 1
+        assert_eq!(
+            Scalar::DELTA.pow(&[
+                0x9604_98c6_973d_74fb,
```
rozbb, line 2005: What is `t`?

rozbb (commented):
Looks good, thank you! Left a few comments. Structurally, I think it also might be cleaner to put all the trait impls into a separate file. `edwards.rs` has almost 1/4 of its non-test code spent on impls.

src/scalar.rs
tarcieri: FWIW these constants are documented in the trait definition itself:

https://docs.rs/ff/latest/ff/trait.PrimeField.html#associatedconstant.S

Cargo.toml
```diff
@@ -65,6 +66,7 @@ packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_
 [features]
 default = ["alloc"]
 alloc = ["zeroize/alloc"]
+group = ["group_crate", "rand_core"]
```
tarcieri, line 69: With #485 landed this can now be:

```suggestion
group = ["dep:group", "rand_core"]
```

src/scalar.rs
tarcieri: It's `MULTIPLICATIVE_GENERATOR ^ t mod p` where `t = (modulus - 1) >> s`

src/scalar.rs
tarcieri: `MULTIPLICATIVE_GENERATOR ^ (2^s) mod p`, i.e. `s` squarings of `MULTIPLICATIVE_GENERATOR`

ycscaly:
What is preventing the merge?

rozbb:
This feature is taking a back seat to the other things we need before 4.0. This can likely come in 4.1.

ycscaly:
> This feature is taking a back seat to the other things we need before 4.0. This can likely come in 4.1.

Thanks, and is there any rough estimation on when 4.1 will come?

Also, is the code here already working or needs much more work?

tarcieri:
@str4d if you have some time, would you mind rebasing this?

I think @rozbb intends to cut a 4.0 release soon, and it'd be good to at least preflight this PR prior to a release to make sure it can be merged afterward without breaking changes.

dconnolly:
> @str4d if you have some time, would you mind rebasing this?
> 
> I think @rozbb intends to cut a 4.0 release soon, and it'd be good to at least preflight this PR prior to a release to make sure it can be merged afterward without breaking changes.

4.0 is out 🎉

tarcieri:
Would love to get this merged soon if possible. People are still asking me about it.

pinkforest:
Might just move this along:
- #562

rozbb:
Superceded by #562

PR #475, opened 2022-12-12
Fixed doc build

rozbb:
`make doc` wasn't building after #470 because of missing doc comments and also because `simd_backend` isn't a feature anymore. Fixed those.

Also I edited the docs CSS to display code at the normal size. It always appeared too small to me.


str4d (approved):
utACK

PR #477, opened 2022-12-13
Fixed docsrs flags in Cargo.toml

rozbb:
Current build [fails](https://docs.rs/crate/curve25519-dalek/4.0.0-pre.3/builds/700411) because `simd_backend` was accidentally listed as a feature. This should fix it, but I haven't tested it myself.


tarcieri (approved):

rozbb:
Ok seems to work locally

PR #481, opened 2022-12-18
Make `zeroize` an optional dependency

tarcieri:
The `zeroize` crate provides a defense against memory read oracles which typically arise from memory unsafety.

Pure Rust programs may not benefit from `zeroize`, and in certain cases the unsafe code used by `zeroize` may be more concerning.

This commit makes `zeroize` into an optional feature so users may elect to disable it if they so desire.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ rand_core = { version = "0.6.4", default-features = false, optional = true }
 digest = { version = "0.10", default-features = false, optional = true }
 subtle = { version = "2.3.0", default-features = false }
 serde = { version = "1.0", default-features = false, optional = true, features = ["derive"] }
-zeroize = { version = "1", default-features = false }
+zeroize = { version = "1", default-features = false, optional = true }
 
 [target.'cfg(curve25519_dalek_backend = "fiat")'.dependencies]
 fiat-crypto = "0.1.6"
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -68,8 +68,8 @@ fiat-crypto = "0.1.6"
 packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_bits"] }
 
 [features]
-default = ["alloc"]
-alloc = ["zeroize/alloc"]
+default = ["alloc", "zeroize"]
```
tarcieri, line 71: Note: still on-by-default

Cargo.toml
```diff
@@ -68,8 +68,8 @@ fiat-crypto = "0.1.6"
...
+alloc = ["zeroize/alloc"] # TODO: use weak feature activation
```
tarcieri, line 72: Another use case for weak feature activation

rozbb:
LGTM, but is this something people are asking for? `zeroize` seems like a very conservative dependency to me, plus none of the types of `ZeroizeOnDrop`, so it's not like the unsafe code is running by default anyway.

tarcieri:
Yes, several people have asked me about this.

IIRC @noncombatant requested it

PR #485, opened 2022-12-27
Weakly activate `zeroize?/alloc`; MSRV 1.60

tarcieri:
Previously `alloc` implicitly activated `zeroize` via `zeroize/alloc`.

This commit switches to weak feature activation as added in Rust 1.60, only activating `zeroize/alloc` if the `zeroize` dependency is explicitly activated (which it is by default).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -28,6 +28,7 @@ jobs:
       - run: rustup target add ${{ matrix.target }}
       - run: ${{ matrix.deps }}
       - run: cargo test --target ${{ matrix.target }} --no-default-features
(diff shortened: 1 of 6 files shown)


tarcieri:
This use case alone isn't a great argument for bumping MSRV in and of itself, although Rust 1.60 is nearly 8 months old at this point, and should be fairly widely adopted.

However, #473 would also benefit as it's currently doing:

```toml
[dependencies]
group_crate = { version = "0.13", default-features = false, optional = true, package = "group" }

[features]
group = ["group_crate", "rand_core"]
```

...which could instead be the following with namespaced features:

```toml
[features]
```

PR #488, opened 2022-12-27
Make basepoint table constants `&'static` references

### LINKED ISSUE #486: Should the tables be static instead of const?
I wonder if the fact that the generator tables are const could result in the table replicating across compilation units and if it should be a static instead.

I did not check binaries for this nor did I dive into how by-reference const access happens in rust

tarcieri:
All current accesses to the basepoint tables currently use explicit borrowing.

It's good borrowing is used to prevent them from being copied onto the stack, however instead of requiring the caller to borrow the constant, instead the constant can simply be defined as a `&'static` reference.

Closes #486 (I think? cc @elichai)


elichai:
According to the rust reference: https://doc.rust-lang.org/reference/items/constant-items.html
> References to the same constant are not necessarily guaranteed to refer to the same memory address.

So the reference doesn't guarantee that the current implementation won't end up duplicating the table

And I'm not longer sure that even a const reference is enough: https://github.com/rust-lang/rust/issues/72004

src/constants.rs
```diff
@@ -93,8 +93,8 @@ pub const BASEPOINT_ORDER: Scalar = Scalar {
 
 use crate::ristretto::RistrettoBasepointTable;
 /// The Ristretto basepoint, as a `RistrettoBasepointTable` for scalar multiplication.
-pub const RISTRETTO_BASEPOINT_TABLE: RistrettoBasepointTable =
-    RistrettoBasepointTable(ED25519_BASEPOINT_TABLE);
+pub const RISTRETTO_BASEPOINT_TABLE: &'static RistrettoBasepointTable =
+    &RistrettoBasepointTable(ED25519_BASEPOINT_TABLE_INNER_DOC_HIDDEN);
```
tarcieri, line 97: @elichai this in particular makes switching to `static` difficult... `RistrettoBasepointTable` is a newtype for an `EdwardsBasepointTable`.

At the very least it would need to have an added lifetime and refer to the `EdwardsBasepointTable` by reference. That's not entirely trivial as there are various APIs that expect ownership of the inner basepoint table, namely `RistrettoBasepointTable::create`

tarcieri: nm, got it to work in d69e67c

tarcieri:
Okay, I figured out how to make all of the basepoint tables `static` in d69e67c.

They're now all `static` as well as `&'static ref`. Not sure if that's overkill or not.

src/backend/serial/u32/constants.rs
```diff
@@ -234,11 +234,12 @@ pub const EIGHT_TORSION_INNER_DOC_HIDDEN: [EdwardsPoint; 8] = [
 ];
 
 /// Table containing precomputed multiples of the Ed25519 basepoint \\(B = (x, 4/5)\\).
-pub const ED25519_BASEPOINT_TABLE: EdwardsBasepointTable = ED25519_BASEPOINT_TABLE_INNER_DOC_HIDDEN;
+pub static ED25519_BASEPOINT_TABLE: &'static EdwardsBasepointTable =
```
elichai, line 237: I think this can be `const` considering that it points to an actual static allocation (`ED25519_BASEPOINT_TABLE_INNER_DOC_HIDDEN`)

src/backend/serial/u64/constants.rs
```diff
@@ -321,11 +321,12 @@ pub const EIGHT_TORSION_INNER_DOC_HIDDEN: [EdwardsPoint; 8] = [
 ];
 
 /// Table containing precomputed multiples of the Ed25519 basepoint \\(B = (x, 4/5)\\).
-pub const ED25519_BASEPOINT_TABLE: EdwardsBasepointTable = ED25519_BASEPOINT_TABLE_INNER_DOC_HIDDEN;
+pub static ED25519_BASEPOINT_TABLE: &'static EdwardsBasepointTable =
```
elichai, line 324: Also this

src/constants.rs
```diff
@@ -93,8 +93,8 @@ pub const BASEPOINT_ORDER: Scalar = Scalar {
 
 use crate::ristretto::RistrettoBasepointTable;
 /// The Ristretto basepoint, as a `RistrettoBasepointTable` for scalar multiplication.
-pub const RISTRETTO_BASEPOINT_TABLE: RistrettoBasepointTable =
-    RistrettoBasepointTable(ED25519_BASEPOINT_TABLE);
+pub static RISTRETTO_BASEPOINT_TABLE: &'static RistrettoBasepointTable =
```
elichai, line 96: I think that this has to stay `static` because it's a new type and doesn't point to the same table (notice that this means that the table might appear twice in the binary once for ed25519 and once for ristretto).

This can probably be bypassed(but sadly stay as static) with some `#[repr(transparent)]` and unsafe (https://play.rust-lang.org/?gist=c73adb96a26f2fbc90e67e250f1750de)

src/backend/serial/u32/constants.rs
tarcieri: Nope

```
error[E0013]: constants cannot refer to statics
   --> src/backend/serial/u64/constants.rs:325:6
    |
325 |     &ED25519_BASEPOINT_TABLE_INNER_DOC_HIDDEN;
    |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

src/constants.rs
tarcieri: Okay, 78d79bb uses "const crimes" to cast the static reference to `EdwardsBasepointTable` to a now `#[repr(transparent)]` `RistrettoBasepointTable`, which avoids making a copy of the table.

tarcieri:
Went ahead and squashed and split out the rustfmt changes into a separate commit.

There are <50 lines of actual changes in 59b4f6b73d70a4e05e0ec50464ef2fc12b811ca6 but thanks to rustfmt it's over 9000 total.

tarcieri:
Phew, green again.

Sidebar: we might consider feature-gating these basepoint tables.

rozbb:
Looks good, thank you! This is how I checked the fmt step:
```
/t/curve25519-dalek (main)$ git diff 3c6ed86^..3c6ed86 | shasum -a 256
    42953ce3fdbb4f7297d5dfe2b1647977325047b68a8cc72b10e68adf534d4b46  -
/t/curve25519-dalek (main)$ git checkout 3c6ed86^
/t/curve25519-dalek ((00df3786…))$ cargo fmt
/t/curve25519-dalek ((00df3786…))$ git diff | shasum -a 256
    42953ce3fdbb4f7297d5dfe2b1647977325047b68a8cc72b10e68adf534d4b46  -
```

Issue #355, opened 2021-05-28
Problem with large stack frames on BPF target


brson:
I recently tried to compile curve25519-dalek for a Solana program, which targets BPF. This platform has stack frames limited to 4KB, and when I build for it I see these errors:

```
Error: Function _ZN117_$LT$curve25519_dalek..edwards..EdwardsBasepointTableRadix128$u20$as$u20$curve25519_dalek..traits..BasepointTable$GT$6create17h27a4aab4
d6ab2304E Stack offset of -7680 exceeded max offset of -4096 by 3584 bytes, please minimize large stack variables
Error: Function _ZN117_$LT$curve25519_dalek..edwards..EdwardsBasepointTableRadix256$u20$as$u20$curve25519_dalek..traits..BasepointTable$GT$6create17h399b7a66
709f3a4bE Stack offset of -15360 exceeded max offset of -4096 by 11264 bytes, please minimize large stack variables
Error: Function _ZN16curve25519_dalek5field79_$LT$impl$u20$curve25519_dalek..backend..serial..u64..field..FieldElement51$GT$8pow2250117hba5efddbdd192654E Sta
ck offset of -4496 exceeded max offset of -4096 by 400 bytes, please minimize large stack variables
Error: Function _ZN209_$LT$curve25519_dalek..window..NafLookupTable8$LT$curve25519_dalek..backend..serial..curve_models..ProjectiveNielsPoint$GT$$u20$as$u20$
core..convert..From$LT$$RF$curve25519_dalek..edwards..EdwardsPoint$GT$$GT$4from17h77ae8748c9a7483fE Stack offset of -10920 exceeded max offset of -4096 by 68
24 bytes, please minimize large stack variables
Error: Function _ZN205_$LT$curve25519_dalek..window..NafLookupTable8$LT$curve25519_dalek..backend..serial..curve_models..AffineNielsPoint$GT$$u20$as$u20$core
..convert..From$LT$$RF$curve25519_dalek..edwards..EdwardsPoint$GT$$GT$4from17h24d4257e81f959d3E Stack offset of -8320 exceeded max offset of -4096 by 4224 by
tes, please minimize large stack variables
Error: Function _ZN212_$LT$curve25519_dalek..window..LookupTableRadix64$LT$curve25519_dalek..backend..serial..curve_models..ProjectiveNielsPoint$GT$$u20$as$u
[...]

PR #489, opened 2022-12-27
Add `basepoint-tables` crate feature

tarcieri:
~~*Note: depends on #488 which should be merged first.*~~

Feature-gates the inclusion of basepoint tables under a `basepoint-tables` feature, with the goal of reducing code size for e.g. embedded applications.

In its current for this probably isn't sufficient to address e.g. #355 but it's a start.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,8 +63,9 @@ fiat-crypto = "0.1.6"
 packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_bits"] }
 
 [features]
-default = ["alloc", "zeroize"]
+default = ["alloc", "basepoint-tables", "zeroize"]
 alloc = ["zeroize?/alloc"]
+basepoint-tables = []
 
 [profile.dev]
 opt-level = 2
--- a/src/backend/serial/u32/constants.rs
+++ b/src/backend/serial/u32/constants.rs
@@ -16,8 +16,11 @@
 use super::field::FieldElement2625;
 use super::scalar::Scalar29;
 use crate::backend::serial::curve_models::AffineNielsPoint;
-use crate::edwards::{EdwardsBasepointTable, EdwardsPoint};
-use crate::window::{LookupTable, NafLookupTable8};
+use crate::edwards::EdwardsPoint;
+use crate::window::NafLookupTable8;
+
+#[cfg(feature = "basepoint-tables")]
+use crate::{edwards::EdwardsBasepointTable, window::LookupTable};
 
 /// The value of minus one, equal to `-&FieldElement::ONE`
 pub(crate) const MINUS_ONE: FieldElement2625 = FieldElement2625([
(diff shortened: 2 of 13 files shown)


tarcieri:
Rebased and ready for review.

I imagine a question will be: "Won't the linker remove the basepoint tables from the resulting binary if they're unused?"

The short answer is yes (depending on how the linker is configured), however per #355 it's possible to hit code size limitations prior to linking depending on the target.

Being able to shut these constants off through conditional compilation eliminates these concerns entirely, and I would imagine that most users who reach for `default-features = false` (i.e. embedded users) will almost always want them off.

src/edwards.rs
```diff
@@ -702,6 +702,24 @@ impl<'a, 'b> Mul<&'b EdwardsPoint> for &'a Scalar {
...
+            scalar * constants::ED25519_BASEPOINT_POINT
+        }
+
+        #[cfg(feature = "basepoint-tables")]
+        {
+            scalar * constants::ED25519_BASEPOINT_TABLE
+        }
+    }
+}
+
```
rozbb, line 722: If pre-linker code size is a concern, then we could probably also stand to gate-out the `impl_basepoint_table!` and `impl_basepoint_table_conversions!` macros in this file.

src/backend/serial/u32/constants.rs
```diff
@@ -234,11 +237,13 @@ pub const EIGHT_TORSION_INNER_DOC_HIDDEN: [EdwardsPoint; 8] = [
 ];
 
 /// Table containing precomputed multiples of the Ed25519 basepoint \\(B = (x, 4/5)\\).
+#[cfg(feature = "basepoint-tables")]
 pub static ED25519_BASEPOINT_TABLE: &'static EdwardsBasepointTable =
```
rozbb, line 241: Should this even be `pub` anymore? Now that there's a `mul_base()` method, I think it would be best if this constant gets hidden away. This would also remove the need for the `traits::BasepointTable` trait.

rozbb (commented):
Left some comments and made 2 commits. The commits were just to make more tests run even when `basepoint-tables` isn't set.

Cargo.toml
```diff
@@ -63,8 +63,9 @@ fiat-crypto = "0.1.6"
 packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_bits"] }
 
 [features]
-default = ["alloc", "zeroize"]
+default = ["alloc", "basepoint-tables", "zeroize"]
```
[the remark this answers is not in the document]
tarcieri: Unfortunately there's no clear guidance on "snake" vs "kebab" case: https://github.com/rust-lang/api-guidelines/discussions/101 however that issue suggests that feature names should match the style of the crate name, where the dalek crates use hyphens (i.e. `curve25519-dalek` vs `curve25519_dalek`).

Per https://github.com/rust-lang/api-guidelines/discussions/29 I think there's something of a preference for kebab case

tarcieri:
@rozbb on both the points you raised I agree it would make sense to keep the basepoint table statics and the related functionality for defining them out of the public API, which among other things would permit making breaking changes to how they're structured.

rozbb:
Great. I'll make some commits tn to hide them all.

rozbb:
It looks like we expose a good amount of functionality here. I think we should keep both actually
* [`BasepointTable`](https://docs.rs/curve25519-dalek/4.0.0-pre.5/curve25519_dalek/traits/trait.BasepointTable.html) lets you create your own table. This seems pretty useful
* If someone writes code that's generic over the basepoint table, then they might want to use the default one. This means we have to expose `ED25519_BASEPOINT_TABLE`

Also a naming inconsistency: there's `BasepointTable::basepoint_mul` and `EdwardsPoint::mul_base`. I like the latter better. Shall I rename the trait method?

rozbb:
I put more code behind `basepoint-tables`. Not sure if the extra noise is worth it though. Is there an easy way to measure pre-linking code size?

tarcieri:
> Also a naming inconsistency: there's BasepointTable::basepoint_mul and EdwardsPoint::mul_base. I like the latter better. Shall I rename the trait method?

Yeah I was going to ask about that. Another candidate is `mul_by_generator`: https://github.com/zkcrypto/group/pull/44

Issue #497, opened 2023-01-07
Proposal to add `SecretSeed` trait(?) into `curve25519-dalek` crate as public service util.


SergeStrashko:
Proposal:
Implement utility `Seed` like following
```rust
pub trait SecretSeed {
    fn random() -> Scalar; // generates new random and does `clamp_bytes`
    fn from_bytes(&[u8]) -> Scalar; //does `clamp_bytes` on input and returns Scalar
}

fn clamp_bytes(&[u8]) -> Scalar;

```
- Repos [`ed25519-dalek`](https://github.com/dalek-cryptography/ed25519-dalek) and [`x25519-dalek`](https://github.com/dalek-cryptography/x25519-dalek) can use this trait to obtain always compliant secret bytes. 
- It will (*as far as I can see - should*) allow to drop dependency on `rand` and `rand-core` in those repos.

Rationale:
1. https://github.com/dalek-cryptography/ed25519-dalek/blob/7529d65506147b6cb24ca6d8f4fc062cac33b395/src/public.rs#L150
```rust
    fn mangle_scalar_bits_and_multiply_by_basepoint_to_produce_public_key(
        bits: &mut [u8; 32],
    ) -> PublicKey {
        bits[0] &= 248;
        bits[31] &= 127;
        bits[31] |= 64;

        let point = &Scalar::from_bits(*bits) * &constants::ED25519_BASEPOINT_TABLE;
        let compressed = point.compress();

        PublicKey(compressed, point)
    }
```
2. https://github.com/dalek-cryptography/ed25519-dalek/blob/7529d65506147b6cb24ca6d8f4fc062cac33b395/src/secret.rs#L286 
```rust
    fn from(secret_key: &'a SecretKey) -> ExpandedSecretKey {
        let mut h: Sha512 = Sha512::default();
        let mut hash:  [u8; 64] = [0u8; 64];
        let mut lower: [u8; 32] = [0u8; 32];
        let mut upper: [u8; 32] = [0u8; 32];

        h.update(secret_key.as_bytes());
        hash.copy_from_slice(h.finalize().as_slice());

        lower.copy_from_slice(&hash[00..32]);
        upper.copy_from_slice(&hash[32..64]);

        lower[0]  &= 248;
        lower[31] &=  63;
        lower[31] |=  64;

        ExpandedSecretKey{ key: Scalar::from_bits(lower), nonce: upper, }
    }
```
[...]

PR #498, opened 2023-01-08
Add `clamp` function to `Scalar` struct (#497)

SergeStrashko:
* test added
* Little change to `Scalar::from_bytes` to remove `mut Scalar` variable (I think this way it's a bit more clear)

### DIFF
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -144,7 +144,7 @@ use core::cmp::{Eq, PartialEq};
 use core::convert::TryInto;
 use core::fmt::Debug;
 use core::iter::{Product, Sum};
-use core::ops::{Index};
+use core::ops::Index;
 use core::ops::Neg;
 use core::ops::{Add, AddAssign};
 use core::ops::{Mul, MulAssign};
--- a/src/scalar.rs
+++ b/src/scalar.rs
@@ -144,7 +144,7 @@ use core::cmp::{Eq, PartialEq};
 use core::convert::TryInto;
 use core::fmt::Debug;
 use core::iter::{Product, Sum};
-use core::ops::Index;
+use core::ops::{Index};
 use core::ops::Neg;
 use core::ops::{Add, AddAssign};
 use core::ops::{Mul, MulAssign};
@@ -270,12 +270,33 @@ impl Scalar {
     /// require specific bit-patterns when performing scalar
     /// multiplication.
     pub const fn from_bits(bytes: [u8; 32]) -> Scalar {
-        let mut s = Scalar { bytes };
-        // Ensure that s < 2^255 by masking the high bit
-        s.bytes[31] &= 0b0111_1111;
-
-        s
+        let mut bytes:[u8; 32] = bytes;
+        bytes[31] &= 0b0111_1111;
+        Scalar { bytes }
     }
+
+    /// Construct a `Scalar` from the low 255 bits of a little-endian 256-bit integer
+    /// `clamping` it's value to be in range
+    ///
+    /// **n ∈ 2^254 + 8\*{0, 1, 2, 3, . . ., 2^251 − 1}**
+    pub const fn from_bits_clamped(bytes: [u8; 32]) -> Scalar {
+        Scalar{ bytes : Scalar::clamp(bytes) }
+    }
+
+    /// Constructs `[u8; 32]` bytes representing 255 bit little-endinan integer in range
+    ///
+    /// **n ∈ 2^254 + 8\*{0, 1, 2, 3, . . ., 2^251 − 1}**
+    ///
+    /// input value `bytes` is treated as little-endian 256-bit integer
+    pub const fn clamp(bytes: [u8; 32]) -> [u8; 32] {
+        let mut bytes : [u8; 32] = bytes;
+        bytes[0]  &= 0b1111_1000;
+        bytes[31] &= 0b0111_1111;
+        bytes[31] |= 0b0100_0000;
+
+        bytes
+    }
+
 }
 
 impl Debug for Scalar {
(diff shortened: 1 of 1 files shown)


tarcieri:
Just as a general impression: you're making a lot of superfluous copies of secret key material. IMO when dealing with key material you should try to make as few copies as possible.

LLVM will most likely optimize them all away, but it would be better if you wrote the code in such a way it didn't make copies in the first place.

src/scalar.rs
```diff
@@ -270,12 +270,33 @@ impl Scalar {
     /// require specific bit-patterns when performing scalar
     /// multiplication.
     pub const fn from_bits(bytes: [u8; 32]) -> Scalar {
-        let mut s = Scalar { bytes };
-        // Ensure that s < 2^255 by masking the high bit
-        s.bytes[31] &= 0b0111_1111;
+        let mut bytes:[u8; 32] = bytes;
+        bytes[31] &= 0b0111_1111;
+        Scalar { bytes }
```
tarcieri, line 275: This will require making the above `bytes` param `mut`, i.e. `mut bytes: [u8; 32]`.

```suggestion
        bytes[31] &= 0b0111_1111;
        Scalar { bytes }
```

src/scalar.rs
```diff
@@ -270,12 +270,33 @@ impl Scalar {
...
+    }
 
-        s
+    /// Construct a `Scalar` from the low 255 bits of a little-endian 256-bit integer
+    /// `clamping` it's value to be in range
+    ///
+    /// **n ∈ 2^254 + 8\*{0, 1, 2, 3, . . ., 2^251 − 1}**
+    pub const fn from_bits_clamped(bytes: [u8; 32]) -> Scalar {
```
tarcieri, line 282: `from_bits_clamped` is arguably a better name than `clamp`, but I don't see any point in having both functions

src/scalar.rs
```diff
@@ -270,12 +270,33 @@ impl Scalar {
...
+        Scalar{ bytes : Scalar::clamp(bytes) }
+    }
```
tarcieri, line 284: I think there's little value in splitting what you have as `clamp` off into its own function.

```suggestion
    pub const fn from_bits_clamped(mut bytes: [u8; 32]) -> Scalar {
        bytes[0]  &= 0b1111_1000;
        bytes[31] &= 0b0111_1111;
        bytes[31] |= 0b0100_0000;
        Scalar{ bytes }
    }
```

src/scalar.rs
```diff
@@ -270,12 +270,33 @@ impl Scalar {
...
+    ///
+    /// input value `bytes` is treated as little-endian 256-bit integer
+    pub const fn clamp(bytes: [u8; 32]) -> [u8; 32] {
+        let mut bytes : [u8; 32] = bytes;
+        bytes[0]  &= 0b1111_1000;
+        bytes[31] &= 0b0111_1111;
+        bytes[31] |= 0b0100_0000;
+
+        bytes
     }
```
tarcieri, line 298: Several things here:

1. This is an inherent method of `Scalar` but it doesn't actually deal with the `Scalar` type? That's weird
2. It's once again making an unnecessary copy of `bytes`
3. I don't think there's a use case for clamping the bytes without using them to construct a `Scalar`?

I'd suggest just getting rid of it and folding it into a function that clamps and constructs a `Scalar`.

src/scalar.rs
SergeStrashko: Reverted
I actually checked this on godbolt. creating `mut Scalar` consumes less stack.

src/scalar.rs
SergeStrashko: renamed ->`pub fn clamp(...) -> Scalar;`

src/scalar.rs
SergeStrashko: two functions merged.

src/scalar.rs
tarcieri: FWIW `from_bits_clamped` is arguably a more descriptive name

src/scalar.rs
SergeStrashko: removed.

src/scalar.rs
SergeStrashko: sorry. its back.

src/scalar.rs
```diff
@@ -276,6 +276,20 @@ impl Scalar {
 
         s
     }
+
+    /// Construct a `Scalar` from the low 255 bits of a little-endian 256-bit integer
+    /// `clamping` it's value to be in range
+    ///
+    /// **n ∈ 2^254 + 8\*{0, 1, 2, 3, . . ., 2^251 − 1}**
```
tarcieri, line 283: FWIW this is a good explanatory blog post on clamping it might be worth linking to:

https://neilmadden.blog/2020/05/28/whats-the-curve25519-clamping-all-about/

Or at the very least, it would be good to summarize what the three operations do like that post did:

> For Curve25519, h = 8, and multiplying by 8 is the same as a binary left-shift by 3 bits. If you take a secret scalar value between 2^251 and 2^252 – 1 and left-shift by 3 bits then you end up with a 255-bit number with the most significant bit set to 1 and the least-significant three bits set to 0. The Curve25519 clamping operation takes an arbitrary 256-bit random value and clears the most-significant bit (making it a 255-bit number), sets the next bit, and then clears the 3 least-significant bits. In other words, it directly creates a scalar value that is in the right form and pre-multiplied by the cofactor.

tarcieri (approved):

src/scalar.rs
SergeStrashko: Updated

src/scalar.rs
```diff
@@ -276,6 +276,34 @@ impl Scalar {
...
+    /// For Curve25519, h = 8, and multiplying by 8 is the same as a binary left-shift by 3 bits.
+    /// If you take a secret scalar value between 2^251 and 2^252 – 1 and left-shift by 3 bits
+    /// then you end up with a 255-bit number with the most significant bit set to 1 and
+    /// the least-significant three bits set to 0.
+    /// The Curve25519 clamping operation takes **an arbitrary 256-bit random value** and
+    /// clears the most-significant bit (making it a 255-bit number), sets the next bit, and then
+    /// clears the 3 least-significant bits. In other words, it directly creates a scalar value that is
+    /// in the right form and pre-multiplied by the cofactor.
+    ///
+    /// refer <https://neilmadden.blog/2020/05/28/whats-the-curve25519-clamping-all-about/> for details
```
tarcieri, line 296:
```suggestion
    /// See <https://neilmadden.blog/2020/05/28/whats-the-curve25519-clamping-all-about/> for details
```

SergeStrashko: fixed

src/scalar.rs
```diff
@@ -276,6 +276,34 @@ impl Scalar {
...
+    /// `clamping` it's value to be in range
+    ///
+    /// **n ∈ 2^254 + 8\*{0, 1, 2, 3, . . ., 2^251 − 1}**
+    ///
+    /// # Explanation of `clamping`
+    ///
+    /// For Curve25519, h = 8, and multiplying by 8 is the same as a binary left-shift by 3 bits.
+    /// If you take a secret scalar value between 2^251 and 2^252 – 1 and left-shift by 3 bits
+    /// then you end up with a 255-bit number with the most significant bit set to 1 and
+    /// the least-significant three bits set to 0.
```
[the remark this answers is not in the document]
SergeStrashko: fixed

tarcieri:
FWIW, this is a great writeup on clamping: https://www.reddit.com/r/crypto/comments/107i1ar/why_do_we_need_to_clamp_25519_keys_when_we/j3nj8y5/

PR #500, opened 2023-01-20
Feature-gated more precomputed tables

rozbb:
We still had some size-8 NAF tables lying around for the `aA + bB` vartime procedure. This puts those tables behind the `precomputed-tables` feature gate.

New bench results for `precomputed-tables` on my M1 machine, using u64 serial backend:
65% speedup on fixed-base mul, 9% speedup on `aA + bB` vartime mul. And the `.rlib` file went from 1.4MB to 758KB (54% of original).

Weirdly, there was a 9% _slowdown_ on vartime fixed-based multiscalar mul over 512 elems, but a 6% speedup on the same thing over 1024 elements. Go figure.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ platforms = "3.0.2"
 [[bench]]
 name = "dalek_benchmarks"
 harness = false
-required-features = ["rand_core"]
+required-features = ["alloc", "rand_core"]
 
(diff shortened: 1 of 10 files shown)


tarcieri (approved):
Nice! I think with this all of the large precomputed tables are now gated.

PR #503, opened 2023-01-29
Fixed-based Montgomery scalar multiplication

tarcieri:
Adds `MontgomeryPoint::mul_base` as an API for fixed-base scalar multiplication which allows for potential future optimizations.

As a baseline implementation, it uses the variable base scalar multiplication implementation.

This follows the existing `EdwardsPoint::mul_base` and `RistrettoPoint::mul_base` APIs.

### DIFF
--- a/src/montgomery.rs
+++ b/src/montgomery.rs
@@ -54,7 +54,7 @@ use core::{
     ops::{Mul, MulAssign},
 };
 
-use crate::constants::{APLUS2_OVER_FOUR, MONTGOMERY_A, MONTGOMERY_A_NEG, X25519_BASEPOINT};
+use crate::constants::{APLUS2_OVER_FOUR, MONTGOMERY_A, MONTGOMERY_A_NEG};
 use crate::edwards::{CompressedEdwardsY, EdwardsPoint};
 use crate::field::FieldElement;
 use crate::scalar::Scalar;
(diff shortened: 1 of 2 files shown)


src/montgomery.rs
```diff
@@ -118,6 +118,11 @@ impl Zeroize for MontgomeryPoint {
 }
 
 impl MontgomeryPoint {
+    /// Fixed-base scalar multiplication (i.e. multiplication by the base point).
+    pub fn mul_base(scalar: &Scalar) -> Self {
+        scalar * X25519_BASEPOINT
```
tarcieri, line 123: Curious if `EdwardsPoint::mul_base(scalar).to_montgomery()` might be faster when `precomputed-tables` is enabled, but that needs a benchmark to evaluate (which might be worth adding to this PR).

That performs another scalar multiplication and an inversion so it's definitely not cheap, but maybe with the basepoint tables it would still be a win.

rozbb: I can benchmark this. Currently x25519-dalek does `EdwardsPoint::mul_base` and converts

rozbb:
Ok that was unexpected. On my machine there's a 3.5% speedup using `EdwardsPoint::mul_base(scalar).to_montgomery()`  _without_ precomputed tables. And then ofc with precomputed tables it's another 65% speedup.

One thing that's weird is benchmark filtering doesn't work anymore. Lmk if you've dealt with that before.

tarcieri:
utACK

PR #505, opened 2023-01-30
Fix docsrs build

rozbb:
This adds docs build to the CI, and fixes the [recently failed](https://docs.rs/crate/curve25519-dalek/4.0.0-rc.0/builds/731542) rc.0 build on docsrs.

I fixed the rc.0 error, but I'm now getting an error about `packed_simd` I can't quite diagnose. @tarcieri would you be able to take a look?

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -69,6 +69,15 @@ jobs:
         RUSTFLAGS: '--cfg curve25519_dalek_backend="simd" -C target_feature=+avx512ifma'
       run: cargo build --target x86_64-unknown-linux-gnu
 
+  build-docs:
+    name: Build docs
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dtolnay/rust-toolchain@nightly
+    - run: make doc
+    - run: make doc-internal
+
   cross:
     strategy:
       matrix:
--- a/src/backend/vector/mod.rs
+++ b/src/backend/vector/mod.rs
@@ -50,8 +50,5 @@ pub mod scalar_mul;
 ))]
 pub(crate) use self::avx2::constants::BASEPOINT_ODD_LOOKUP_TABLE;
 
-#[cfg(any(
-    all(target_feature = "avx512ifma", feature = "precomputed-tables"),
-    all(docsrs, target_arch = "x86_64")
-))]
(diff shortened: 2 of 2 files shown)


tarcieri:
`packed_simd` is gated as follows:

https://github.com/dalek-cryptography/curve25519-dalek/blob/1a388f7/Cargo.toml#L62

```toml
[target.'cfg(curve25519_dalek_backend = "simd")'.dependencies]
```

So it seems like it's probably trying to document code that makes use of `packed_simd` when `--cfg curve25519_dalek_backend = "simd"` hasn't been set

rozbb:
> So it seems like it's probably trying to document code that makes use of packed_simd when --cfg curve25519_dalek_backend = "simd" hasn't been set

I don't think that's it. In `Makefile`:
```make
doc:
	cargo +nightly rustdoc --features "$(FEATURES)" -- --html-in-header docs/assets/rustdoc-include-katex-header.html --cfg docsrs --cfg=curve25519_dalek_backend=\"simd\"
```

tarcieri:
After playing around with this locally quite a bit, I *think* this works?

```
$ RUSTFLAGS='--cfg curve25519_dalek_backend="simd" -C target_feature=+avx2' RUSTDOCFLAGS='--cfg curve25519_dalek_backend="simd" -C target_feature=+avx2' cargo +nightly rustdoc --open
```

tarcieri:
Here's an updated version that should match what's in the current `Makefile`:

```
$ RUSTFLAGS='--cfg docsrs --cfg curve25519_dalek_backend="simd"' RUSTDOCFLAGS='--cfg docsrs --cfg curve25519_dalek_backend="simd"' cargo +nightly rustdoc --all-features --open -- --html-in-header docs/assets/rustdoc-include-katex-header.html
```

The important part is `RUSTFLAGS` and `RUSTDOCFLAGS` need to match (you could set one to the other in environment variables in advance). `RUSTFLAGS` is needed for the dependencies in Cargo.toml that are cfg-gated to be selected.

rozbb:
Awesome, that works. Thank you!

I see now that the reason docsrs previously succeeded was bc this distinction was already made in `Cargo.toml`:
```toml
[package.metadata.docs.rs]
rustdoc-args = [
    "--html-in-header", "docs/assets/rustdoc-include-katex-header.html",
    "--cfg", "docsrs",
]
rustc-args = ["--cfg", "curve25519_dalek_backend=\"simd\""]
```

PR #510, opened 2023-03-09
Fix error for custom targets

### LINKED ISSUE #508: Build script failure on platform detection
Automatic backend (`DalekBits`) detection in curve25519-dalek-4.0.0-rc.1 `build.rs` fails at unwrap `platforms::Platform::find(&target_triplet).unwrap()` while building a risc0 guest method using ed25519-dalek crate. The actual `target_triplet` is `riscv32im-risc0-zkvm-elf`. I was unable to configure `curve25519_dalek_bits` externally, risc0-build doesn't allow it yet, and `.cargo/config` didn't work either. As a workaround we could return `DalekBits::Dalek32` if the platform is not found since this is the default value returned for weird `target_pointer_width`.

pinkforest:
Fixes #508 
Supercedes #509 

Desired behaviour is to require explicit setting in case of custom targets and give our helpful error message.

Also clears out the unwrap for missing standard TARGET in case of being used in non-Cargo build system.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -10,54 +10,6 @@ enum DalekBits {
     Dalek64,
 }
 
-// Only when the curve25519_dalek_bits is not explicitly set.
-// Error handling when the bits setting cannot be determined
-#[cfg(all(not(curve25519_dalek_bits = "64"), not(curve25519_dalek_bits = "32")))]
-fn determine_curve25519_dalek_bits_error(error_code: i32, trigger_msg: &str) -> ! {
-    eprintln!("Error: {trigger_msg}");
-    eprintln!("Please set cfg(curve25519_dalek_bits) explicitly either as 32 or 64.");
-    std::process::exit(error_code)
-}
-
-// Only when the curve25519_dalek_bits is not explicitly set.
-#[cfg(all(not(curve25519_dalek_bits = "64"), not(curve25519_dalek_bits = "32")))]
-fn determine_curve25519_dalek_bits() -> DalekBits {
-    use platforms::target::PointerWidth;
-
-    // TARGET environment is supplied by Cargo
-    // https://doc.rust-lang.org/cargo/reference/environment-variables.html
-    let target_triplet = match std::env::var("TARGET") {
-        Ok(t) => t,
-        Err(_) => determine_curve25519_dalek_bits_error(
-            10,
-            "Standard Cargo TARGET environment variable is not set",
-        ),
-    };
-
-    // platforms crate is the source of truth used to determine the platform
-    // Custom target platforms require explicit settings.
-    let platform = match platforms::Platform::find(&target_triplet) {
-        Some(p) => p,
-        None => determine_curve25519_dalek_bits_error(20, "Unrecognised Rust target"),
-    };
-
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -8,13 +10,36 @@ enum DalekBits {
...
     use platforms::target::PointerWidth;
 
-    let target_triplet = std::env::var("TARGET").unwrap();
-    let platform = platforms::Platform::find(&target_triplet).unwrap();
+    // TARGET environment is supplied by Cargo
+    // https://doc.rust-lang.org/cargo/reference/environment-variables.html
+    let target_triplet = match std::env::var("TARGET") {
+        Ok(t) => t,
+        Err(_) => determine_curve25519_dalek_bits_error(
+            10,
```
rozbb, line 32: What's the purpose of the constants exactly? If it makes sense, I'd rather have them be `const`s up top

build.rs
```diff
@@ -8,13 +10,36 @@ enum DalekBits {
...
+            "Standard Cargo TARGET environment variable is not set",
+        ),
+    };
+
+    // platforms crate is the source of truth used to determine the platform
+    // Custom target platforms require explicit settings.
+    let platform = match platforms::Platform::find(&target_triplet) {
+        Some(p) => p,
+        None => determine_curve25519_dalek_bits_error(20, "Unrecognised Rust target"),
```
rozbb, line 41: Ditto re the constant. Also can it be "recognized" instead? We use american spelling elsewhere :)

build.rs
```diff
@@ -8,13 +10,36 @@ enum DalekBits {
...
-fn lotto_curve25519_dalek_bits() -> DalekBits {
+fn determine_curve25519_dalek_bits_error(error_code: i32, trigger_msg: &str) -> ! {
+    eprintln!("Error: {trigger_msg}");
+    eprintln!("Please set cfg(curve25519_dalek_bits) explicitly either as 32 or 64.");
+    std::process::exit(error_code)
+}
+
+// Only when the curve25519_dalek_bits is not explicitly set.
+#[cfg(all(not(curve25519_dalek_bits = "64"), not(curve25519_dalek_bits = "32")))]
+fn determine_curve25519_dalek_bits() -> DalekBits {
```
rozbb, line 24: ty I like this name a lot better

rozbb (changes requested):
Nice! Left small comments

build.rs
pinkforest: It's Unknown now :) United Nations :rofl:

build.rs
pinkforest: I appreciate these tyvm :purple_heart:

semenov-vladyslav:
This RP does not fix #508. The logic is pretty much the same is before but with nicer error messages. The issue is that I can't set `curve25519_dalek_bits`. Maybe it's the issue with the tools that I'm using, and not with this crate though.

pinkforest:
Can you please read the original issue and answer the question that has been asked ?

pinkforest:
Referring back to: https://github.com/dalek-cryptography/curve25519-dalek/issues/508#issuecomment-1465050817

PR #513, opened 2023-03-13
fix default dalek

SWvheerden:
ledger compatibility branch

DO NOT MERGE


pinkforest:
Does this have to be a PR here as it doesn't concern dalek ?

That said the custom build tooling needs to be fixed elsewhere to signal the cfg correctly - we've cleared out the error msg.

SWvheerden:
Haha yeah sorry. This should not have been opened up here. Its part of a string of test WIP test PR's.

PR #516, opened 2023-03-17
Fixes cfg with target from env

### LINKED ISSUE #511: Default `curve25519_dalek_bits` on non-standard Rust target / build tooling
**Follow-up**

- #508 and
- #510 

**Use-Case**

- **a)** Non-standard Rust build tooling which doesn't signal `cfg` correctly and

- **b)** Custom non-Rust targets which are not known or

- **c)** Situations where the `RUSTENV` was set incorrectly but relied on

**Current**

We spit out an error https://github.com/dalek-cryptography/curve25519-dalek/issues/508#issuecomment-1465049852 where we ask the downstream to set the desired `cfg(curve25519_dalek_bits)`

This ensures that the downstream is consciously selecting the correct `curve25519_dalek_bits` if the downstream is attempting to use the `cfg` to set it and errors out appropriately when the target platform is non-standard Rust target with unknown `target_pointer_width`

**Alternative**

Set the `cfg(curve25519_dalek_bits)` to `32` silently if either the `cfg` was not signalled or a custom non-standard Rust target.

This would have a downside where it would silently set something else but it would "just work" on other hand but silently.

**Potential Improvements**

1) Maybe we could print out a warning ? Do people read warnings ?

2) Also maybe could add some tests that simulate these error messages.

3) Document these error messages in README

### LINKED ISSUE #515: Incorrect backend selection for non-standard targets
hi folks,

we're compiling for a non-standard (ie. out-of-tree) target which means target detection doesn't work, and setting `'--cfg=curve25519_dalek_bits="32"'` per the docs (via the environment _or_ `.cargo/config`) does not have any effect, resulting in the error:
```
Error: Unknown Rust target platform.
Please set cfg(curve25519_dalek_bits) explicitly either as 32 or 64.
```

digging into the `build.rs` script it appears y'all are using `cfg` flags to switch features, however, when cross compiling these are not propagated to native components (ie. build scripts) at build time, only when they are executed (so in effect these are always ignored for foreign targets?).

via https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags:

> If the --target flag (or [build.target](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget)) is used, then the flags will only be passed to the compiler for the target. Things being built for the host, such as build scripts or proc macros, will not receive the args. Without --target, the flags will be passed to all compiler invocations (including build scripts and proc macros) because dependencies are shared. If you have args that you do not want to pass to build scripts or proc macros and are building for the host, pass --target with the host triple.

based on the intent here it appears this should instead be switching on environmental variables as these are passed when the build script is _executed_ during compilation for the foreign target, so to have this behave as documented the fix would be something like:

``` diff
diff --git a/build.rs b/build.rs
index a4c6b3e..f22df58 100644
--- a/build.rs
+++ b/build.rs
@@ -11,14 +11,17 @@ enum DalekBits {
 }

 fn main() {
-    #[cfg(curve25519_dalek_bits = "32")]
-    let curve25519_dalek_bits = DalekBits::Dalek32;
-
-    #[cfg(curve25519_dalek_bits = "64")]
-    let curve25519_dalek_bits = DalekBits::Dalek64;
-
-    #[cfg(all(not(curve25519_dalek_bits = "64"), not(curve25519_dalek_bits = "32")))]
-    let curve25519_dalek_bits = deterministic::determine_curve25519_dalek_bits();
+    let curve25519_dalek_bits = match std::env::var("CARGO_CFG_CURVE25519_DALEK_BITS").as_deref() {
+        Ok("32") => DalekBits::Dalek32,
+        Ok("64") => DalekBits::Dalek64,
+        _ => deterministic::determine_curve25519_dalek_bits(),
+    };

     match curve25519_dalek_bits {
         DalekBits::Dalek64 => println!("cargo:rustc-cfg=curve25519_dalek_bits=\"64\""),
```

happy to open a PR if y'all would be open to this fix?

pinkforest:
Fixes #515 by @ryankurte fix
Fixes #511 thanks @jcape for pointing out the warning message

We fix the cross build with override which was broken due to rustc not relaying curve cfg's.

Given it is difficult to use we also default to `cfg(curve_25519_dalek_bits="32")` with a warning instead of a hard error.

Slightly vary of using `CARGO_CFG_` given rustc removes them in build.rs for `#[cfg()]` beyond target `cfg` set.

Cc/ @jrose-signal

Test matrix on x86_64-unknown-linux-gnu native host:

$ rustup default 1.60 ; cargo clean
$ rustup default stable ; cargo clean
$ rustup default nightly ; cargo clean

## Dalek64
$ cargo build
$ env RUSTFLAGS='--cfg curve25519_dalek_bits="64"' cargo build
$ env RUSTFLAGS='--cfg curve25519_dalek_bits="64"' cargo build --target wasm32-unknown-unknown
$ env RUSTFLAGS='--cfg curve25519_dalek_bits="64"' cargo build --target aarch64-unknown-linux-gnu
$ env RUSTFLAGS='--cfg curve25519_dalek_bits="64"' cargo build --target i686-unknown-linux-gnu
$ env RUSTFLAGS='--cfg curve25519_dalek_bits="64"' cargo build --no-default-features --target thumbv7em-none-eabi

## Dalek32
$ env RUSTFLAGS='--cfg curve25519_dalek_bits="32"' cargo build
$ env RUSTFLAGS='--cfg curve25519_dalek_bits="32"' cargo build --target wasm32-unknown-unknown
$ cargo build --target wasm32-unknown-unknown
$ env RUSTFLAGS='--cfg curve25519_dalek_bits="32"' cargo build --target aarch64-unknown-linux-gnu
$ cargo build --target i686-unknown-linux-gnu
$ env RUSTFLAGS='--cfg curve25519_dalek_bits="32"' cargo build --target i686-unknown-linux-gnu
$ cargo build --no-default-features --target thumbv7em-none-eabi
$ env RUSTFLAGS='--cfg curve25519_dalek_bits="32"' cargo build --no-default-features --target thumbv7em-none-eabi

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -27,16 +27,14 @@ mod deterministic {
     use super::*;
 
     // Standard Cargo TARGET environment variable of triplet is required
-    static ERR_MSG_NO_TARGET: &str = "Standard Cargo TARGET environment variable is not set.";
+    static ERR_MSG_NO_TARGET: &str = "Standard Cargo TARGET environment variable is not set";
 
     // Custom Non-Rust standard target platforms require explicit settings.
     static ERR_MSG_NO_PLATFORM: &str = "Unknown Rust target platform.";
 
-    // Error handling when the bits setting cannot be determined
-    fn determine_curve25519_dalek_bits_error(cause: &str) -> ! {
-        eprintln!("Error: {cause}");
-        eprintln!("Please set cfg(curve25519_dalek_bits) explicitly either as 32 or 64.");
-        std::process::exit(1)
+    // Warning when the curve25519_dalek_bits cannot be determined
+    fn determine_curve25519_dalek_bits_warning(cause: &str) {
+        println!("cargo:warning=\"Defaulting to curve25519_dalek_bits=32: {cause}\"");
     }
 
     // Determine the curve25519_dalek_bits based on Rust standard TARGET triplet
@@ -47,13 +45,19 @@ mod deterministic {
         // https://doc.rust-lang.org/cargo/reference/environment-variables.html
         let target_triplet = match std::env::var("TARGET") {
             Ok(t) => t,
-            Err(_) => determine_curve25519_dalek_bits_error(ERR_MSG_NO_TARGET),
+            Err(_) => {
+                determine_curve25519_dalek_bits_warning(ERR_MSG_NO_TARGET);
+                return DalekBits::Dalek32;
+            }
         };
 
         // platforms crate is the source of truth used to determine the platform
         let platform = match platforms::Platform::find(&target_triplet) {
             Some(p) => p,
-            None => determine_curve25519_dalek_bits_error(ERR_MSG_NO_PLATFORM),
+            None => {
+                determine_curve25519_dalek_bits_warning(ERR_MSG_NO_PLATFORM);
+                return DalekBits::Dalek32;
+            }
         };
 
         #[allow(clippy::match_single_binding)]
--- a/build.rs
+++ b/build.rs
@@ -2,23 +2,19 @@
 
 #![deny(clippy::unwrap_used, dead_code)]
 
-#[allow(non_camel_case_types)]
+#[allow(non_camel_case_types, dead_code)]
+#[derive(Debug)]
 enum DalekBits {
-    #[cfg_attr(curve25519_dalek_bits = "64", allow(dead_code))]
     Dalek32,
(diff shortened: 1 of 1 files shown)


ryankurte:
🎉 thanks for fixing this @pinkforest

Issue #514, opened 2023-03-14
Scalar arithmetic is inconsistent with point multiplication most of the time


rozbb:
# What's wrong

Scalars in various curve25519-based schemes are considered as integers to be multiplied with a curve point. This is fine. In other protocols, scalars are considered as integers mod ℓ to be multiplied with curve points in the prime ℓ-order subgroup.
This is also fine.

A problem arises when you mix these two interpretations of scalars. It doesn't matter if you use `s` or `s + 3ℓ` to multiply by a point in the prime-order subgroup (since they're the same mod ℓ). But it _does_ matter when the point is not in the prime-order subgroup. Concretely, the following assert fails most of the time:
```rust
// Make a random clamped scalar
let s = {
    let mut bytes = [0u8; 32];
    rng.fill_bytes(&mut bytes);
    Scalar::from_bits_clamped(bytes)
};

// Make a random point
let point = {
    let mut bytes = [0u8; 32];
    rng.fill_bytes(&mut bytes);
    CompressedEdwardsY::from_slice(&bytes)
        .unwrap()
        .decompress()
        .unwrap()
};

assert_eq!(s * point, s.reduce() * point);
```

In other words, scalar mod ℓ arithmetic is incompatible with most of the points on the curve. Here's a demonstration that the distributive property is violated. This assert fails most of the time:
```rust
// Make two random scalars that are reduced mod l
let (s1, s2) = {
    let mut bytes1 = [0u8; 32];
    let mut bytes2 = [0u8; 32];
    rng.fill_bytes(&mut bytes1);
    rng.fill_bytes(&mut bytes2);
    (
        Scalar::from_bytes_mod_order(bytes1),
        Scalar::from_bytes_mod_order(bytes2),
    )
};

// Make a random point
let point = {
    let mut bytes = [0u8; 32];
    rng.fill_bytes(&mut bytes);
    CompressedEdwardsY::from_slice(&bytes)
        .unwrap()
        .decompress()
        .unwrap()
};

assert_eq!((s1 + s2) * point, s1 * point + s2 * point);
```

# What to do

~~You don't need arithmetic for `Scalar` almost ever. I propose removing the `Add` and `Mul` traits for `Scalar`. To keep functionality, there should be a new type, `ScalarFieldElem` which has these traits and is guaranteed to always be reduced mod ℓ.~~

[...]

Issue #507, opened 2023-02-13
Poor Scalar performance


kayabaNerve:
dalek scalars are about 10x as inperformant as k256 Scalars, for both addition and multiplication, according to some recent benchmarking I had performed. While that isn't a fair comparison, one being curve25519 and one being secp256k1, it is a striking difference that shows *something* is likely going on with dalek's impl. The following block of code was pointed out to me as the likely culprit: https://github.com/dalek-cryptography/curve25519-dalek/blob/b2d0f0ecd7808f194e67a2fcd4c47dd5be8a0697/src/scalar.rs#L367-L375 Please note I have not flamegraphed this to confirm.

Would it be possible to expose an addition which does not hedge against the potential use of an API in existence for x25519, greatly accelerating Ed25519 operations? Potentially a distinct ReducedScalar type entirely?

tarcieri:
I'm curious if there are actually good use cases for `Scalar::from_bits` to be public besides `Scalar::from_bits_clamped`.

Also curious what overhead using bytes as the internal "packed" representation of a scalar might be adding (if any) or if it impacts LLVM inlining (`k256` uses an array of limbs, similar to `UnpackedScalar`).

kayabaNerve:
Beyond x25519, which I can't comment on, I think the remaining concern may be how the Group API requires the modulus as a Scalar (not just via char_le_bits), which requires unreduced Scalars as a possibility.

tarcieri:
> I think the remaining concern may be how the Group API requires the modulus as a Scalar 

It doesn't though? For `Scalar` it'd be the `Field` and `PrimeField` traits, and of those there's only `PrimeField::MODULUS`, which is a freeform string.

> not just via `char_le_bits`

(note: there's that on the `PrimeFieldBits` trait as well)

But beyond that there's nothing that requires an unreduced `Scalar` in the API.

kayabaNerve:
... you are right. I've even implemented 0.13 into my own libs. I have no idea why I thought it did that. Sorry.

kayabaNerve:
BASEPOINT_ORDER also currently exists as a way to access an unreduced Scalar.

tarcieri:
Indeed. Wonder if that could be made private at least.

rozbb:
[...]

PR #519, opened 2023-03-20
Make scalars always reduced

rozbb:
This makes unreduced `Scalar` unrepresentable by removing `Scalar::{from_bits, from_bytes_clamped}`. As a consequence, we can remove the additional reduction steps in scalar addition and subtraction. Benchmarks on my M1 Macbook Air show runtime diffs of -52% and -67% on scalar addition and subtraction, respectively.

Since clamping is necessary elsewhere, this change exposes new methods `MontgomeryPoint::{mul_clamped, mul_base_clamped}`, `EdwardsPoint::{mul_clamped, mul_base_clamped}`, and `BasepointTable::mul_base_clamped`, which take byte sequences.

This partially addresses #507 (scalar performance) and #514 (scalar arithmetic well-definedness).

The first snippet of #514 points out that it is possible to construct a clamped scalar `s` and a point `P` such that `s * P` and `s.reduce() * P` are not equal. This change makes this test case impossible to represent, since unreduced scalars are now unrepresentable.

# Upstream changes

https://github.com/dalek-cryptography/ed25519-dalek/pull/293

https://github.com/dalek-cryptography/x25519-dalek/pull/120


src/edwards.rs
```diff
@@ -1289,6 +1327,20 @@ mod test {
         0x2b, 0x42,
     ]);
 
+    /// The largest valid scalar (not mod l). Remember for NAF computations, the top bit has to be
+    // 0. So the largest integer a scalar can hold is 2^255 - 1. Addition and subtraction are
+    // broken on unreduced scalars. The only thing you can do with this is multiplying with a curve
+    // point (and actually also scalar-scalar multiplication, but that's just a quirk of our
+    // implementation).
+    #[cfg(feature = "precomputed-tables")]
+    static LARGEST_UNREDUCED_SCALAR: Scalar = Scalar {
```
tarcieri, line 1336: Why is this defined both here and in scalar.rs?

src/scalar.rs
```diff
@@ -1297,6 +1252,20 @@ mod test {
         ],
     };
 
+    /// The largest valid scalar (not mod l). Remember for NAF computations, the top bit has to be
+    // 0. So the largest integer a scalar can hold is 2^255 - 1. Addition and subtraction are
+    // broken on unreduced scalars. The only thing you can do with this is multiplying with a curve
+    // point (and actually also scalar-scalar multiplication, but that's just a quirk of our
+    // implementation).
+    #[cfg(feature = "precomputed-tables")]
+    static LARGEST_UNREDUCED_SCALAR: Scalar = Scalar {
```
tarcieri, line 1261: I guess this is just using it for tests? But regardless it seems like it could be defined in a single place

src/scalar.rs
```diff
@@ -1336,27 +1305,19 @@ mod test {
...
-            0xe7, 0x6d, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
-            0xff, 0xff, 0xff, 0x0f,
+            0xec, 0xd3, 0xf5, 0x5c, 0x1a, 0x63, 0x12, 0x58, 0xd6, 0x9c, 0xf7, 0xa2, 0xde, 0xf9,
+            0xde, 0x14, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x10,
         ],
     };
 
-    static CANONICAL_LARGEST_ED25519_S_MINUS_ONE: Scalar = Scalar {
+    static LARGEST_ED25519_S: Scalar = Scalar {
```
tarcieri, line 1316: If `clamp_integer` were `const fn`, couldn't this be computed by clamping `LARGEST_UNREDUCED_SCALAR`?

tarcieri:
I like the API and the direction. Preventing the clamped scalars from leaking as returned values seems like an elegant approach.

src/scalar.rs
rozbb: Agreed. None of the tests seem to wanna copy values around, but I can refactor so `edwards.rs` just pulls in `scalar::test::*`

src/edwards.rs
rozbb: Will fix. See below

src/scalar.rs
rozbb: Yup good catch

rozbb:
@jrose-signal would you mind taking a look at this and verifying it wouldn't break your libraries?

Cargo.toml
```diff
@@ -66,6 +66,7 @@ packed_simd = { version = "0.3.4", package = "packed_simd_2", features = ["into_
 default = ["alloc", "precomputed-tables", "zeroize"]
 alloc = ["zeroize?/alloc"]
 precomputed-tables = []
+legacy_compatibility = []
```
rozbb, line 69: I don't love the naming of this. It matches the `ed25519-dalek` one though. Thoughts?

src/montgomery.rs
```diff
@@ -342,6 +370,9 @@ impl<'a, 'b> Mul<&'b Scalar> for &'a MontgomeryPoint {
             W: FieldElement::ONE,
         };
 
+        // NOTE: The below swap-double-add routine skips the first iteration, i.e., it assumes the
+        // MSB of `scalar` is 0. This is allowed, since it follows from Scalar invariant #1.
+
```
rozbb, line 375: I noticed that this was the case because the following test fails.
```rust
let a_bytes = [0xff; 32];
let s1 = Scalar::from_bytes_mod_order(a_bytes);
let s2 = Scalar { bytes: a_bytes };
assert_eq!(
    s1 * constants::X25519_BASEPOINT,
    s2 * constants::X25519_BASEPOINT
);
```
That's kinda surprising. There's no NAF or any fancy arithmetic necessary for the Montgomery ladder. Anyways, neither here nor there, because it works for all scalars < 2^255, which is all we care about.

Cargo.toml
tarcieri: Matching the name in `ed25519-dalek` seems fine.

I was thinking it might also be good to explicitly deprecate the methods that are feature-gated by this too.

rozbb: Wait, do we intend to remove this function? Are we deprecating `ed25519-dalek` legacy compat soon?

tarcieri: My worry is more feature flags are bad at gating APIs we don't want people to use because they're additive, so one crate in the project activating the feature makes the functionality available to all.

Deprecation is a bit clearer about that.

jrose-signal:
> @jrose-signal would you mind taking a look at this and verifying it wouldn't break your libraries?

Yes, this is fine for us (and I had someone more cryptographer than me confirm). I did find one place we were using `Scalar::from_bits` directly: [XEd25519 signatures](https://github.com/signalapp/libsignal/blob/main/rust/protocol/src/curve/curve25519.rs#L52). But those Scalars are always multiplied by the Edwards basepoint, so switching to `from_bytes_mod_order` for an eager reduction is fine (and doesn't seem to cause any significant slowdown).

Cargo.toml
rozbb: Isn't it noisy though? Like if you pull in a library that has to accept legacy signatures, you're going to see a compiler warning every time

rozbb: We can continue the conversation here. I'm gonna merge the rest of it for now so I can update the upstream stuff

jrose-signal: Warnings aren't shown in dependencies by default in Rust+Cargo (for this reason).

The client can use `#[allow(deprecated)]` too (and probably will).

rozbb: Ah, nice! Ok adding now.

src/montgomery.rs
tarcieri: I mean, that's expected for me. One scalar is reduced, the other is not, but `Scalar { bytes: a_bytes }` would not be valid under `Scalar`'s (new) invariant and can't be constructed outside the crate due to field visibility.

sidebar: I'm curious if `Scalar` could be made into a newtype for `ScalarImpl` now.

src/scalar.rs
```diff
@@ -257,51 +250,29 @@ impl Scalar {
...
+    /// that scalars are always reduced. Scalar-scalar arithmetic, i.e., addition, subtraction,
+    /// multiplication, **does not work** on scalars produced from this function. You may only use
+    /// the output of this function for `EdwardsPoint::mul`, `MontgomeryPoint::mul`, and
+    /// `EdwardsPoint::vartime_double_scalar_mul_basepoint`. **Do not use this function** unless
+    /// you absolutely have to.
+    #[cfg(feature = "legacy_compatibility")]
+    #[deprecated(
+        since = "4.0.0",
+        note = "This constructor outputs scalars with undefined scalar-scalar arithmetic. See docs."
+    )]
```
rozbb, line 271: @tarcieri how's this?

tarcieri: Looks great!

tarcieri (approved):

rozbb:
@jrose-signal thanks for the quick turnaround! We'll be propagating the API change to ed- and x- soon

elichai:
We currently use the `Scalar::from_bits` function in order to do full-group (non-clamped) Scalar*Point multiplication via the MontgomeryPoint API,
We need this to be over the whole group + twist and not specifically in the prime subgroup,
Is this something that will still be supported?

(Specifically, this is needed for something similar to Moller04 https://www.bmoeller.de/pdf/pke-pseudo-esorics2004.pdf)

elichai:
> @elichai would a dedicated API for this, e.g. `MontgomeryPoint::mul_unclamped` similar to [`MontgomeryPoint::mul_clamped`](https://docs.rs/curve25519-dalek/4.0.0-rc.3/curve25519_dalek/montgomery/struct.MontgomeryPoint.html#method.mul_clamped) work for your purposes?

Yep, that would definitely work

elichai:
> @elichai would a dedicated API for this, e.g. `MontgomeryPoint::mul_unclamped` similar to [`MontgomeryPoint::mul_clamped`](https://docs.rs/curve25519-dalek/4.0.0-rc.3/curve25519_dalek/montgomery/struct.MontgomeryPoint.html#method.mul_clamped) work for your purposes?

What do you think about this function also overriding invariant #1? i.e. it will allow multiplying with the top bit on, so it can't assume the first bit is zero as in: https://docs.rs/curve25519-dalek/latest/src/curve25519_dalek/montgomery.rs.html#373-374

elichai:
> Perhaps it would need to be fallible and return an e.g. `CtOption`?

Why fallible? if it's bigger than `8 \cdot \ell` ?

tarcieri:
Do you have a different solution in mind for it overriding the invariants?

rozbb:
I think it'd be possible to just extend the iterator to all bits. I'd prefer to not copy the code, so I'll just modify this function and see if we get a performance regression

PR #520, opened 2023-03-21
Support SIMD on Rust stable

koute:
This PR makes the SIMD backend work on stable Rust.

- The nightly-only `packed_simd` dependency was removed.
- A small mostly API-compatible replacement for `packed_simd` was added.
- The AVX2 backend now works on stable Rust.
- The AVX512 backend still requires nightly as those intrinsics are not yet stabilized, so on stable Rust the AVX2 backend will be used even if the target CPU supports AVX512.


tarcieri:
Awesome!

koute:
After this PR gets in I also have another PR in queue (not yet finished, but works already) where I want to switch SIMD to be entirely autodetected at runtime so that it's not necessary to compile with `target_feature` anymore (most likely at little to no performance loss; in fact I sped it up a little from what I can see where e.g. the vartime fixed base multiplication bench dropped from ~4.2ms to ~4.08ms on the AVX2 backend).

tarcieri:
Runtime autodetection sounds great!

jcape:
Can I ask that you use either `cpufeatures` or allow this to be selected manually. There's no CPUID in an SGX enclave.

koute:
> Can I ask that you use either use `cpufeatures` (non-vendored) or allow this to be selected manually. There's no CPUID in an SGX enclave.

Yes, I will still keep the ability to override/disable this; I'd just like autodetection to be the default.

tarcieri:
FWIW, I have a PR open to add AVX-512 support to `cpufeatures`: https://github.com/RustCrypto/utils/pull/862

koute:
(Rebased to resolve the conflict in `Cargo.toml`.)

OtaK:
If I can interject in the conversation, I have a small question.

Knowing `std::simd` (and `packed_simd_2`) supports producing WASM SIMD instructions, I was wondering if it was possible to relax [this check](https://github.com/dalek-cryptography/curve25519-dalek/blob/c982811d117620af420342edf5d68c2e21f16865/src/backend/vector/mod.rs#L15) to allow choosing the SIMD backend when the `platform_family` is `wasm` and `target_features` includes `simd128`.
 
If that's a trivial change, that would allow WASM ed25519 to benefit from WASM basically for free. If it's not, then nevermind.

koute:
> If I can interject in the conversation, I have a small question.
> 
> Knowing `std::simd` (and `packed_simd_2`) supports producing WASM SIMD instructions, I was wondering if it was possible to relax [this check](https://github.com/dalek-cryptography/curve25519-dalek/blob/c982811d117620af420342edf5d68c2e21f16865/src/backend/vector/mod.rs#L15) to allow choosing the SIMD backend when the `platform_family` is `wasm` and `target_features` includes `simd128`.
> 
> If that's a trivial change, that would allow WASM ed25519 to benefit from WASM basically for free. If it's not, then nevermind.

All of the current SIMD backends use explicit architecture-specific compiler intrinsics, so that wouldn't do anything besides breaking compilation when targetting WASM. The current dependency on `packed_simd` is not there because the code takes advantage of the "it's SIMD but it's portable" aspect of that crate, but because it defines a few wrapper types for the SIMD types which are more convenient to use.

You'd basically have to add a WASM-specific backend.

jrose-signal:
It's also important that whatever operations are used remain constant-time, which needs to be verified for every platform.

src/backend/vector/packed_simd.rs
```diff
@@ -0,0 +1,252 @@
...
+    _mm256_extract_epi32
+);
+
+impl_conv!(u64x4 => u32x8, i32x8);
+
+#[allow(dead_code)]
+impl u64x4 {
+    #[inline]
+    pub const fn new_const(x0: u64, x1: u64, x2: u64, x3: u64) -> Self {
+        unsafe { Self(core::mem::transmute([x0, x1, x2, x3])) }
```
rozbb, line 177: Why is this safe?

src/backend/vector/packed_simd.rs
```diff
@@ -0,0 +1,252 @@
...
+        x0: u32,
+        x1: u32,
+        x2: u32,
+        x3: u32,
+        x4: u32,
+        x5: u32,
+        x6: u32,
+        x7: u32,
+    ) -> Self {
+        unsafe { Self(core::mem::transmute([x0, x1, x2, x3, x4, x5, x6, x7])) }
```
rozbb, line 213: Why is this safe?

src/backend/vector/packed_simd.rs
```diff
@@ -0,0 +1,252 @@
...
+            }
+
+            #[inline]
+            pub fn shr<const N: i32>(self) -> Self {
+                unsafe { core::arch::x86_64::$shr_intrinsic(self.0, N).into_bits() }
+            }
+
+            #[inline]
+            pub fn extract<const N: i32>(self) -> $lane_ty {
+                unsafe { core::arch::x86_64::$extract_intrinsic(self.0, N) as $lane_ty }
```
rozbb, line 124: Again, I'm a noob to the tricks of SIMD. Can you make a comment somewhere about why it's okay to cast unsigned to signed SIMD, do adds/subs/shifts, and then cast back to unsigned at the end? It's probably true—I just manually checked `assert_eq!(x.overflowing_add(y).0, (x as i8).overflowing_add(y as i8).0 as u8)` for all `x,y: u8`. But some justification here would be very useful.

src/backend/vector/packed_simd.rs
```diff
@@ -0,0 +1,252 @@
...
+//
+// This file is part of curve25519-dalek.
+// See LICENSE for licensing information.
+
+use core::ops::{Add, AddAssign, BitAnd, BitAndAssign, BitXor, BitXorAssign, Sub};
+
+pub trait IntoBits<T> {
+    fn into_bits(self) -> T;
+}
+
```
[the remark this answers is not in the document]
rozbb: Also, separately, it'd be nice to have some high level comments describing what's happening in this file. We're defining 4 different packed types which are all `__m256i` under the hood, but expose packing-specific functions. That is, one will do overflowing add as if it's 4 `u64`s, and another will do an overflowing add as if it's 8 `i32`s.

src/backend/vector/packed_simd.rs
```diff
@@ -0,0 +1,252 @@
...
+    #[inline]
+    pub fn new(x0: u64, x1: u64, x2: u64, x3: u64) -> Self {
+        unsafe {
+            Self(core::arch::x86_64::_mm256_set_epi64x(
+                x3 as i64, x2 as i64, x1 as i64, x0 as i64,
+            ))
+        }
+    }
+
+    #[inline]
```
rozbb, line 194: A comment here and for `splat` would be nice. "Constructs `Self` as 4 copies of the given `u64`" or something. Ditto for `splat_const`.

rozbb (changes requested):
This is great, and with hardly any changes! I left some commends, mostly in the `packed_simd.rs` file. I'm not super familiar with these intrinsics so some questions might be obvious.

src/backend/vector/packed_simd.rs
```diff
@@ -0,0 +1,252 @@
...
+        $shl_intrinsic:ident,
+        $shr_intrinsic:ident,
+        $extract_intrinsic:ident
+    ) => {
+        #[allow(non_camel_case_types)]
+        #[derive(Copy, Clone, Debug)]
+        #[repr(transparent)]
+        pub struct $ty(core::arch::x86_64::__m256i);
+
+        impl IntoBits<core::arch::x86_64::__m256i> for $ty {
```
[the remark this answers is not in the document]
koute: It isn't; I just kept it because `packed_simd` had it. But you're right that it's unnecessary; I'll remove it!

src/backend/vector/packed_simd.rs
rozbb: Not resolved. This safety question is separate from the i64 to u64 casting question. I think this behavior relies on the fact that `__mm256i` is a newtype for `(u64, u64, u64, u64)`. Is this representation guaranteed to be the same? Further, is it always the case that a newtype has the same memory layout as its underlying type? And is a tuple of 4 the same as a fixed size array of 4?

Some of these questions might be easy, but for the sake of maintainability (and frankly bc I don't actually know the answer) I think they should be written in a comment here.

src/backend/vector/packed_simd.rs
rozbb: Right, makes sense. Can you write that in a comment and maybe give a link to a source if you can find it? I hope it's not too annoying. We just need to make sure that every choice is thoroughly documented. It's high assurance code, after all 😬

src/backend/vector/packed_simd.rs
```diff
@@ -0,0 +1,252 @@
...
+            }
+        }
+
+        impl PartialEq for $ty {
+            #[inline]
+            fn eq(&self, rhs: &$ty) -> bool {
+                unsafe {
+                    let m = core::arch::x86_64::_mm256_cmpeq_epi8(self.0, rhs.0);
+                    core::arch::x86_64::_mm256_movemask_epi8(m) == -1
+                }
```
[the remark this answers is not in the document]
rozbb: Beautiful!

src/backend/vector/packed_simd.rs
rozbb: Oh I apologize @koute I completely missed the comment you added. Resolved

rozbb (approved):

rozbb:
@tarcieri looks good to go, modulo CI and readme updates to reflect that nightly isn't needed for AVX2 anymore.

Also I was looking for a source on why the AVX2 add, sub, mul, shl, and shr functions we use are constant time, but I got nothing.

src/backend/vector/packed_simd.rs
koute: Yeah, sorry! I didn't do this in the first place because I still wasn't exactly sure where and how you want it added. Looks great now!

koute:
> looks good to go, modulo CI and readme updates to reflect that nightly isn't needed for AVX2 anymore.

Do you want me to also update those?

> Also I was looking for a source on why the AVX2 add, sub, mul, shl, and shr functions we use are constant time, but I got nothing.

I can't comment on constant time-ness of the rest of the code existing code as I haven't analyzed it in that regard, but AFAIK all of these should be constant time. Although this does kind of depend on the microarchitectural details of the CPU on which we're running I *think* all of modern Intel and AMD CPUs should have all of these implemented as constant time,

(Also, please note that this PR doesn't really change much here; these were still used before, just indirectly through the `packed_simd` dependency.)

rozbb:
Re docs: that'd be great actually! If you don't have time, one of us will do it, probably this week.

Re constant time: Yup, agreed, and it seems that every serious curve25519 impl uses AVX2. It would just be nice to be able to point to something that justifies our use of it.

koute:
I've updated the README and I've also added a job to the CI to test this. (Hopefully it'll work?)

Initially I only added a job to build the code on Rust stable, but then I realized that these tests are not being run at all, and we should be able to run them, at least for AVX2 (which at this point is *10 years old*, so it's a pretty safe bet to assume that the test runners should support it).

tarcieri:
> Also I was looking for a source on why the AVX2 add, sub, mul, shl, and shr functions we use are constant time, but I got nothing.

They're pretty routinely used for cryptography. I would be fairly surprised to learn they have any data dependent timing variability, which I wouldn't expect from any arithmetic instructions on modern CPUs whether they're SIMD or not.

(They can, however, have timing variability based on thermal throttling, but that's a whole different can of worms)

rozbb:
@tarcieri good to merge? This might also require an MSRV bump right?

tarcieri (approved):
LGTM.

I don't think it needs an MSRV bump? AVX (<512) intrinsics have been stable for quite awhile, and it's passing CI.

rozbb:
Oh ok. I just assumed this wouldn't work with anything prior to the version that stabilized AVX2. Also not sure which version `rustc_version = "0.4.0"` in `Cargo.toml` corresponds to

rozbb:
Oh I was under the impression that stable AVX2 was new. Apparently it was stabilized in 2018. I'll try to consolidate the test to run both builds in MSRV

rozbb:
Thanks again @koute! This is a great addition

koute:
Thanks!

koute:
To anyone who was following this PR, my followup PR with runtime autodetection is now up: https://github.com/dalek-cryptography/curve25519-dalek/pull/523

PR #525, opened 2023-04-28
fix: Use the correct exponent in the description of the `mods` operation

upbqdn:
## Motivation

The exponent defining the upper bound of the result of the `mods` operation in the docs should be `w-1`, not `w`. See the screenshot below for reference.

image

The implementation is not affected by this error.

## Solution

This PR fixes the value of the exponent in the docs.


tarcieri:
Thanks!

PR #528, opened 2023-05-29
Replace `unwrap_u8` with `into`

tarcieri:
Leverages the `From<Choice>` impl for `bool` where applicable instead, which results in clearer logic which more closely matches `bool`.

### DIFF
--- a/src/ristretto.rs
+++ b/src/ristretto.rs
@@ -277,7 +277,7 @@ impl CompressedRistretto {
         let s_encoding_is_canonical = s_bytes_check[..].ct_eq(self.as_bytes());
         let s_is_negative = s.is_negative();
 
-        if (!s_encoding_is_canonical).into() || s_is_negative.into() {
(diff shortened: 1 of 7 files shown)


src/ristretto.rs
```diff
@@ -307,10 +307,7 @@ impl CompressedRistretto {
         // t == ((1+as²) sqrt(4s²/(ad(1+as²)² - (1-as²)²)))/(1-as²)
         let t = &x * &y;
 
-        if ok.unwrap_u8() == 0u8
-            || t.is_negative().unwrap_u8() == 1u8
-            || y.is_zero().unwrap_u8() == 1u8
-        {
+        if (!ok).into() || t.is_negative().into() || y.is_zero().into() {
```
rozbb, line 310: Probably not an issue, but might be better to phrase this as:
```rust
if bool::from(!ok | t.is_negative() | y.is_zero()) {
```

PR #530, opened 2023-06-12
Expand lints

tarcieri:
Adds a lints section to the top of lib.rs with the following:

    #![warn(
        clippy::unwrap_used,
        missing_docs,
        rust_2018_idioms,
        unused_lifetimes,
        unused_qualifications
    )]

`warn` is used instead of `deny` to prevent the lints from firing during local development, however we already configure `-D warnings` in CI so if any lint fails on checked-in code, it will cause a CI failure.

This commit also fixes or explicitly allows any current violations of these lints. The main ones were:

- `clippy::unwrap_used`: replaces usages of `unwrap` with `expect`
- `rust_2018_idioms`: no implicit lifetimes, which were present on usages of `core::fmt::Formatter`
- `unused_lifetimes`: corrects a few instances

### DIFF
--- a/curve25519-dalek/src/backend/serial/curve_models/mod.rs
+++ b/curve25519-dalek/src/backend/serial/curve_models/mod.rs
@@ -527,7 +527,7 @@ impl<'a> Neg for &'a AffineNielsPoint {
 // ------------------------------------------------------------------------
 
 impl Debug for ProjectivePoint {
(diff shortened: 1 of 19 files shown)


robjtede (approved):

rozbb:
These are all good lints imo. Ty

Issue #529, opened 2023-06-01
CPU auto-detection feature / cfg gate


pinkforest:
This should not block landing #523 and can be done as follow-up

Follow-up from 
- https://github.com/dalek-cryptography/curve25519-dalek/pull/523#issuecomment-1572521178

Cc/ @jrose-signal @koute

pinkforest:
Some q's

1) Which use-cases would need to disable the CPU detecion ? Or is this just in case something goes wrong a.k.a "break glass" ?

2) Is there much use for disabling CPU detection or should there be just plain override vs disallowing detection ?

3) Should we just re-use `cfg(curve25519_dalek_backend)` "override" which could be `simd` | `simd_avx512` | `simd_avx2`

4) We already use some hazardous / risky features that should be enabled with care and not sure how many use `--all-features`

Having a plain override with `cfg()` would align to everything else we've done - it would be great to know what use-cases there are to make it easier if it's widely expected to be used ?

tarcieri:
I would think that specifying `curve25519_dalek_backend` would override autodetection

tarcieri:
> If there's ever autodetection for mobile platforms, we might want to turn it off if it's not enough of a speedup to be worth the extra code.

I would think mobile platforms are ARM64, in which case the `neon` target feature is enabled by default (#457)

pinkforest:
Yeah consensus here seems to be that override seems to be the best.

There is this discussion re-opened so let's redirect there to simplify things further:
- https://github.com/dalek-cryptography/curve25519-dalek/issues/414

tarcieri:
Oh sorry, I forgot you opened this issue, but thanks for directing the conversation back to #414

pinkforest:
Yeah I think we got consensus on this issue so your get-together words it well what needs ot happen overall :)

PR #531, opened 2023-06-13
Clean up backend features and vendor curve25519_dalek_derive

### LINKED ISSUE #414: Simplifying backend selection

### LINKED ISSUE #532: Failed backend override - Warning or Error
In implementing the backend overrides over automatic selection:
- #531 

I noticed that if someone tries to override with `--cfg curve25519_dalek_backend="simd"` where the simd override is not possible e.g. in a case of target being wasm32-unknown-unknown - we would get "duplicate" cfg() during the compile time if we try to re-set it as "serial" via build script as a fallback with a warning.

By leaving the failed override as a mere warning - for every simd related gating we would have to include `all(not(curve25519_dalek_backend = "serial"), not(curve25519_dalek_backend = "fiat")` in negative sense ruling out every other possibility in every gate for simd backend.

Since vast majority of the people should not be using this override to begin with and this case should be considered a compile error I've left it to trigger a failure at build time where override fails - which also gets tested via the build test.

By leaving it as a compile error we can have more simplified positive gating - e.g. in simd case it would be mere `cfg(curve25519_dalek_backend = "simd")` instead of a litany of `all(not(every_other))` everywhere - manifest still has to have it that way since it gets evaluated before build.rs.

This also helps ensure somebody doesn't try to supply wrong overrides expecting the override to work in vast minority cases they would be using this.

This also should be forwards compatible given multiple `cfg(param=)` is not really well defined behaviour.

pinkforest:
Closes #414 (again)
Implements:
- #529 consensus to use `cfg(curve25519_dalek_backend = "override")`
- Cleans up the unused backend features which we don't want to advertise for now
- Vendors in unsafe_target_feature in as curve25519-dalek-derive (co-authored by for @koute)
- `cfg(curve25519_dalek_diagnostics="build")` now emits compile_error build diagnostics used for build script testing
- Found a missing zeroize feature gate -[fixed](https://github.com/dalek-cryptography/curve25519-dalek/pull/531/commits/8dd4800af82b38b046ed53abb2b774379f3203ee)
- build script is now tested with tests/build_tests.sh and added to CI
- Clean up & simplify the documentation for backend override

Notes
- github runner had some [weird issue](https://github.com/dalek-cryptography/curve25519-dalek/actions/runs/5256079804/jobs/9496853764?pr=531) with incremental compilation - [fixed](https://github.com/dalek-cryptography/curve25519-dalek/pull/531/commits/ffdefaf7551f7e2004e2e3a22ce23540604c24bf)
- Closes #532 - see: [1ac7b2](https://github.com/dalek-cryptography/curve25519-dalek/pull/531/commits/1ac7b2c96ce302bf98ad2db08548a6db647a32c5) - chose compile_error to simplify gating


vendor/curve25519-dalek-derive/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
```
tarcieri, line 1: Do we really need a `vendor/` directory?

I'd just put this in a toplevel directory called `curve25519-dalek-derive`, and perhaps eventually we can move the rest of the files in the toplevel directory into `curve25519-dalek`

pinkforest: yeah sure, it was already there so I thought to follow that. will move.

src/backend/mod.rs
```diff
@@ -41,26 +41,27 @@ pub mod serial;
...
 
 #[derive(Copy, Clone)]
 enum BackendKind {
     #[cfg(all(
         target_arch = "x86_64",
-        feature = "simd_avx2",
         curve25519_dalek_bits = "64",
-        not(curve25519_dalek_backend = "fiat")
+        not(curve25519_dalek_backend = "fiat"),
+        not(curve25519_dalek_backend = "serial")
```
rozbb, line 56: I think this file would be a lot easier to read if we just made all variants of this enum available. Does it actually matter whether or not `BackendKind::Avx2` exists on non-AVX2 machines? It is properly selected in any case, no?

tests/build_tests.sh
```diff
@@ -0,0 +1,79 @@
...
+      exit 1
+    fi
+}
+
+# Assuming naively 64 bit host
+cargo clean
+OUT=build_1.txt
+env RUSTFLAGS="--cfg curve25519_dalek_diagnostics=\"build\"" cargo build > "$OUT" 2>&1
+match_and_report "curve25519_dalek_backend is 'auto'" "$OUT"
+match_and_report "curve25519_dalek_bits is '64'" "$OUT"
```
rozbb, line 23: oh this is _awesome_

README.md
```diff
@@ -98,17 +95,18 @@ See tracking issue: [curve25519-dalek/issues/521](https://github.com/dalek-crypt
...
 
-| Backend            | Implementation                                                | Target backends             |
-| :---               | :---                                                          | :---                        |
-| `[default]`        | Automatic runtime backend selection (either serial or SIMD)   | `u32` <br/> `u64` <br/> `avx2` <br/> `avx512` |
-| `fiat`             | Formally verified field arithmetic from [fiat-crypto]         | `fiat_u32` <br/> `fiat_u64` |
+| Backend            | Selection | Implementation                                                | Target backends             |
+| :---               | :---      | :---                                                          | :---                        |
+| `serial`           | Automatic | serial software                                               | `u32` <br/> `u64`           |
+| `fiat`             | Manual    | Formally verified field arithmetic from [fiat-crypto]         | `fiat_u32` <br/> `fiat_u64` |
+| `simd`             | Automatic | Intel AVX2 / AVX512 IFMA accelerated backend                  | `avx2` <br/> `avx512`       |
```
rozbb, line 102: I might be a bit dense, but this confuses me as a reader. If I don't set _anything_ myself, what gets selected? IE what's the default? Also, if I set `curve25519_dalek_backend="simd"` on a non-simd system what happens? Or if I set `serial` on a `simd`-enabled system, will it "upgrade" to SIMD anyways because it's automatic?

I just reread the `build.rs` so I now know the answers to these questions, but they were honest ones I had from just the README.

rozbb: Also, `[default]` is still referenced below. And does the below "Word size for serial backends" need to be updated entirely?

rozbb (commented):
Thanks so much for doing this. It's a big PR. Left a few comments about clarity here and there.

rozbb:
Manually verified that `curve25519_dalek_derive/` mirrors the `unsafe_target_feature` repo at [`389ae00`](https://github.com/koute/unsafe_target_feature/tree/389ae00d34cf0ff589cb8d9b38a85ae1b05ebfdc), modulo renaming and metadata changes.

src/backend/mod.rs
pinkforest: We can't use enum's in configuration predicates ? these happen in compile time where enum's are not available

README.md
pinkforest: Ok I've clarified a bit - [Rendered](https://github.com/pinkforest/curve25519-dalek/blob/cleanup-features/README.md)

src/backend/mod.rs
pinkforest: It gives dead code warning - I think in the long run we just may want to have proc macro for capability based selection.
I misread this originally - have fixed now.

pinkforest: Oh it will give warnings about not covered cases for match - so we definitely need to keep it out of enum until we get a procmacro for this.

```
103 |     match get_selected_backend() {
    |           ^^^^^^^^^^^^^^^^^^^^^^ pattern `BackendKind::Avx512` not covered

```

build.rs
```diff
@@ -34,6 +35,33 @@ fn main() {
...
+            if target_arch == "x86_64" && curve25519_dalek_bits == DalekBits::Dalek64 {
+                "simd"
+            // fallback to serial with a warning
+            } else {
+                println!("cargo:warning=Could not override curve25519_dalek_backend to simd - defaulting to serial");
+                "serial"
+            }
+        }
+        // default serial
+        _ => "serial",
```
koute, line 62: Won't this *disable* SIMD by default unless it's enabled by `CARGO_CFG_CURVE25519_DALEK_BACKEND`? Or am I reading this wrong?

README.md
```diff
@@ -90,25 +87,39 @@ latest breaking changes in high level are below:
...
-| :---               | :---                                                          | :---                        |
-| `[default]`        | Automatic runtime backend selection (either serial or SIMD)   | `u32` <br/> `u64` <br/> `avx2` <br/> `avx512` |
-| `fiat`             | Formally verified field arithmetic from [fiat-crypto]         | `fiat_u32` <br/> `fiat_u64` |
+Currently on the intel `x86_64` targets with `64` bit word size, the backend
+selection is automatic between the `simd` and `serial` backends.
 
-To choose a backend other than the `[default]` backend, set the
-environment variable:
+The `simd` backend may be automatically selected if the AVX instruction set is
+detected available on the target CPU.
```
koute, line 106: 1. `AVX` is incorrect; it should be `AVX2`.
2. The wording `may` is unclear. We should precisely specify when each will always be enabled, and when it always be disabled.

src/lib.rs
```diff
@@ -10,14 +10,8 @@
...
-#![cfg_attr(
-    all(target_arch = "x86_64", feature = "simd_avx512", nightly),
-    feature(stdsimd)
-)]
-#![cfg_attr(
-    all(target_arch = "x86_64", feature = "simd_avx512", nightly),
-    feature(avx512_target_feature)
-)]
+#![cfg_attr(all(target_arch = "x86_64", nightly), feature(stdsimd))]
+#![cfg_attr(all(target_arch = "x86_64", nightly), feature(avx512_target_feature))]
```
koute, line 14: Should probably not enable these if a non-simd backend was forced.

build.rs
pinkforest: Oops - yeah my last commit was brainfreeze clearly - fixed I think

README.md
pinkforest: Hmm.. maybe [f41ab86 works](https://github.com/dalek-cryptography/curve25519-dalek/pull/531/commits/f41ab86da454d6dfa3946771011f441e0f2e9a93) - [Rendered](https://github.com/pinkforest/curve25519-dalek/blob/cleanup-features/README.md) ?

src/lib.rs
pinkforest: Yeah tru thanks for catching - How about [edeb5c](https://github.com/dalek-cryptography/curve25519-dalek/pull/531/commits/edeb5c273e0a85010956373c4af1c83974782b5a)

build.rs
pinkforest: @koute I think if if someone tries to override backend to "simd" whilst only "serial" is possible we should do a compile error isntead - this would enable to simplify the gates a lot in backend mod.rs where we don't need a have a litany of all(not()'s

rozbb:
Thansk for the updates! Made a pass on the README which hopefully clarifies how backend selection now works. @koute @tarcieri take a look at just the "Backend" section if you get a chance and lmk if it's good to go.

tarcieri (approved):

koute:
LGTM!

@pinkforest Thank you for pushing this forward!

PR #539, opened 2023-06-30
Monorepo documentation fixes

pinkforest:
We can use this as base on @rozbb to get the docs etc sorted :)


robjtede:
this has a no-op diff from the main branch now, maybe the changes were applied elsewhere

pinkforest:
Yeah I never got into this, been too busy. will address on another PR - just some nits around

PR #541, opened 2023-06-30
Make the to_bytes/from_bytes round trip of ExpandedSecretKey idempotent

poljar:
A mistake snuck in with 9b166b75e0bb0c22bd782665f63638efef72556a. The new `scalar_bytes` field of the `ExpandedSecretKey` is not used in the `ExpandedSecretKey::to_bytes()` method, instead the, now clamped, `scalar` is used.

This in turn means that the following snippet won't produce the same `ExpandedSecretKey`:

```rust
let key = ExpandedSecretKey::from_bytes(bytes);
let another_key = ExpandedSecretKey::from_bytes(&key.to_bytes())
```

Down the line, it means that serializing/deserializing this struct will produce differing keys and thus spurious signature verification errors.

Transferred from: https://github.com/dalek-cryptography/ed25519-dalek/pull/308.

### DIFF
--- a/ed25519-dalek/src/hazmat.rs
+++ b/ed25519-dalek/src/hazmat.rs
@@ -63,7 +63,7 @@ impl ExpandedSecretKey {
     pub fn to_bytes(&self) -> [u8; 64] {
         let mut bytes: [u8; 64] = [0u8; 64];
 
-        bytes[..32].copy_from_slice(self.scalar.as_bytes());
+        bytes[..32].copy_from_slice(&self.scalar_bytes);
         bytes[32..].copy_from_slice(&self.hash_prefix[..]);
         bytes
     }
@@ -221,6 +221,7 @@ mod test {
     use super::*;
 
     use rand::{rngs::OsRng, CryptoRng, RngCore};
+    use sha2::Sha512;
 
     // Pick distinct, non-spec 512-bit hash functions for message and sig-context hashing
     type CtxDigest = blake2::Blake2b512;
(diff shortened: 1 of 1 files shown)


tarcieri:
@poljar what's your use case for `to_bytes`? Could we just get rid of it as I [proposed here](https://github.com/dalek-cryptography/curve25519-dalek/pull/544#issuecomment-1616804527)?

burdges:
It's worth doing https://gitweb.torproject.org/torspec.git/tree/proposals/224-rend-spec-ng.txt#n2135 but this could be expressly supported if whatever way you like.  Ed25519-BIP32 is a mess, so maybe ignore it.

poljar:
> @poljar what's your use case for `to_bytes`? Could we just get rid of it as I [proposed here](https://github.com/dalek-cryptography/curve25519-dalek/pull/544#issuecomment-1616804527)? (Edit: see also #545)

As explained in https://github.com/dalek-cryptography/ed25519-dalek/issues/298#issuecomment-1500616102, [`libolm`] does in fact serialize and persists the unclamped bytes that are now in the `scalar_bytes` field of the `ExpandedSecretKey` type. The persisted `scalar_bytes` are used for a long-term identity key.

We are migrating from [`libolm`] to a Rust-based library, but we need to support existing accounts, which means supporting deserialization and serialization of the `scalar_bytes`. This was previously done using the `Serialize`/`Deserialize` implementation of the `ExpandedSecretKey` type. Currently it's done using the `ExpandedSecretKey::to_bytes()` and `ExpandedSecretKey::from_bytes()` methods.

So the methods are useful to me, of course I can work around `to_bytes()` being removed with a wrapper:

```rust
struct MyExpandedSecretKey {
    source: Box<[u8; 64]>,
    inner: ExpandedSecretKey
}
```

But it does feel a bit strange to not have this method since as far as I can tell, every other type in the ecosystem can be converted to/from a byte slice or array.

[`libolm`]: https://gitlab.matrix.org/matrix-org/olm/

rozbb:
I also proposed in the comments of #544 that we remove `scalar_bytes` entirely. Deserialization should work, but if you load up "legacy" keys then `to_bytes()` might output something different

poljar:
Well, as I said, I can live with this. Just need to implement the above mentioned code snippet in my lib.

Should I close this then or do you guys want to merge this to have correct behavior until the other PRs are ready?

tarcieri:
@rozbb #545 is an alternative to #544 that removes `scalar_bytes` (and with it `to_bytes`)

rozbb:
We decided to get rid of `to_bytes` entirely. Turns out that keeping the unprocessed scalar bytes was unnecessary.

Superseded by #545

PR #542, opened 2023-07-02
Add StreamVerifier

robjtede:
## Context
In web applications, it is desirable to avoid storing the request payload. It is currently impossible to verify a signature based on a request payload that is receieved in a chunk stream without allocating additional memory equal to the size of the payload.

## Solution
A stateful type that holds the public key, candidate signature, and hash state that can be updated incrementally.

## References
- Migrated from https://github.com/dalek-cryptography/ed25519-dalek/pull/196
- As an example of the difference this feature would make: https://github.com/robjtede/actix-web-lab/pull/34/files

### DIFF
--- a/ed25519-dalek/CHANGELOG.md
+++ b/ed25519-dalek/CHANGELOG.md
@@ -34,6 +34,7 @@ Entries are listed in reverse chronological order per undeprecated major series.
 * Add `pkcs` feature to support PKCS #8 (de)serialization of `SigningKey` and `VerifyingKey`
 * Add `fast` feature to include basepoint tables
 * Add tests for validation criteria
+* Add `SigningKey::verify_stream()`, and `VerifyingKey::verify_stream()`
 * Impl `DigestSigner`/`DigestVerifier` for `SigningKey`/`VerifyingKey`, respectively
 * Impl `Hash` for `VerifyingKey`
 * Impl `Clone`, `Drop`, and `ZeroizeOnDrop` for `SigningKey`
--- a/ed25519-dalek/src/signing.rs
+++ b/ed25519-dalek/src/signing.rs
@@ -42,7 +42,7 @@ use crate::{
     errors::{InternalError, SignatureError},
     hazmat::ExpandedSecretKey,
     signature::InternalSignature,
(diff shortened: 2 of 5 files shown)


tarcieri:
> Why not a function that takes impl `Read` instead? 

That requires `std`

> Maybe it's worth adding an MAX_OPTIMAL_BUFFER_SIZE (bikeshedding aside) constant to the Update trait for those reasons? 

I thought we had something like that already but now I can't find it, so maybe we need to add something.

rozbb:
Maybe silly question: isn't this what prehashed signatues are for? Or is this in a setting where you don't have control over what the sender does?

robjtede:
> you don't have control over what the sender does?

yea exactly, my particular use case is Discord Webhooks

There's some relevant discussion in the old PR.

robjtede:
Will rebase after CI fixes here: https://github.com/dalek-cryptography/curve25519-dalek/pull/543

rozbb:
Ok then I like this! One thing to avoid code duplication: do you think you could move the current `VerifyingKey::{verify, verify_strict}` functions to the stream verifier, and redefine the originals to use the stream version? I suspect it won't have any performance cost

robjtede:
Awesome 👍🏻 

> redefine the originals to use the stream version

I've had a go but having trouble doing in a way that actually reduces code. From my limited knowledge of recent refactors, problem seems to be that `VerifyingKey::recompute_R` takes a message and not a `Scalar`. I'm hesitant to do too much refactoring in this PR.

rozbb:
Ah np. I'll take a crack at it

mkj:
I've pushed #556 which builds on `StreamVerifier` and unifies the stream/normal verifiers (as well as adding by-digest methods for signing/verify).

pinkforest:
> `impl Read` needs `std`

fwiw - embedded folk seems to be fan of [a trait](https://docs.rs/embedded-io/latest/embedded_io/trait.Read.html) that works in `no_std`: https://crates.io/crates/embedded-io

tarcieri (approved):

robjtede:
superseded by probably #583 or #556

PR #543, opened 2023-07-02
Address clippy lints

robjtede:


dconnolly (approved):

pinkforest:
@robjtede thanks :sweat_smile: 

Could we also include `--all-features` as well in:
https://github.com/dalek-cryptography/curve25519-dalek/blob/main/.github/workflows/workspace.yml#L68

robjtede:
@pinkforest bump on this since it LG now

robjtede:
bump, this an easy enough merge

tarcieri (approved):
I might suggest disabling `clippy::needless_range_loop` but this is otherwise fine

PR #544, opened 2023-07-02
ed25519-dalek: make `ExpandedSecretKey` fields private

tarcieri:
Motivation can be found in this comment:

https://github.com/dalek-cryptography/ed25519-dalek/pull/293#pullrequestreview-1507115602

Replaces public fields with private fields + accessor methods.

Also adds an `ExpandedSecretKey::from_scalar_and_prefix` constructor which makes it possible to construct `ExpandedSecretKey` using a `Scalar` again, which is useful for protocols that derive the scalar such as Ed25519-BIP32.

Additionally defines a `HashPrefix` type alias to make it semantically clear which `[u8; 32]` is involved in type signatures.

### DIFF
--- a/ed25519-dalek/src/hazmat.rs
+++ b/ed25519-dalek/src/hazmat.rs
@@ -24,6 +24,9 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 use crate::{Signature, VerifyingKey};
 use curve25519_dalek::digest::{generic_array::typenum::U64, Digest};
 
+/// Domain separator used when hashing the message to generate the pseudorandom `r` value.
+pub type HashPrefix = [u8; 32];
+
 /// Contains the secret scalar and domain separator used for generating signatures.
 ///
 /// This is used internally for signing.
@@ -40,14 +43,15 @@ pub struct ExpandedSecretKey {
     // modulo l.
     pub(crate) scalar_bytes: [u8; 32],
     /// The secret scalar used for signing
-    pub scalar: Scalar,
+    pub(crate) scalar: Scalar,
     /// The domain separator used when hashing the message to generate the pseudorandom `r` value
-    pub hash_prefix: [u8; 32],
+    pub(crate) hash_prefix: HashPrefix,
 }
 
 #[cfg(feature = "zeroize")]
 impl Drop for ExpandedSecretKey {
     fn drop(&mut self) {
+        self.scalar_bytes.zeroize();
         self.scalar.zeroize();
         self.hash_prefix.zeroize()
     }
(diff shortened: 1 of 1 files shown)


ed25519-dalek/src/hazmat.rs
```diff
@@ -106,6 +120,18 @@ impl ExpandedSecretKey {
             .into()
         })
     }
+
+    /// Construct an `ExpandedSecretKey` from a scalar and hash prefix.
+    pub fn from_scalar_and_prefix(scalar: Scalar, hash_prefix: HashPrefix) -> Self {
```
tarcieri, line 125: I thought `from_scalar_and_hash_prefix` was a bit unwieldy but it would be more consistent.

This method could probably use a longer comment explaining its use cases (e.g. Ed25519-BIP32).

ed25519-dalek/src/hazmat.rs
```diff
@@ -106,6 +120,18 @@ impl ExpandedSecretKey {
...
+        let scalar_bytes = scalar.to_bytes();
+        Self { scalar_bytes, scalar, hash_prefix }
+    }
+}
+
+impl From<(Scalar, HashPrefix)> for ExpandedSecretKey {
+    fn from(parts: (Scalar, HashPrefix)) -> ExpandedSecretKey {
+        ExpandedSecretKey::from_scalar_and_prefix(parts.0, parts.1)
+    }
 }
```
tarcieri, line 135: This feels a bit debatable

ed25519-dalek/src/hazmat.rs
```diff
@@ -40,14 +43,15 @@ pub struct ExpandedSecretKey {
     // modulo l.
     pub(crate) scalar_bytes: [u8; 32],
```
tarcieri, line 44: Sidebar: perhaps this should be `canonical_scalar_bytes`?

tarcieri:
Alternatively (and alternatively to #541) we could get rid of the `to_bytes` method, which seems to be the reason for retaining it in the first place.

I'm not sure what the motivation is for having it at all. Serializing this format on the wire is an antipattern, IMO.

tarcieri:
I opened #545 as an alternative to this PR to try just getting rid of the private field, which makes all of the fields `pub` now and allows for public construction of `ExpandedSecretKey` again

ed25519-dalek/src/hazmat.rs
rozbb: Actually, now that you bring this up again, I'm not sure if this was even the right choice. The original reason I included `scalar_bytes` was because [section 5.1.5](https://www.rfc-editor.org/rfc/rfc8032#section-5.1.5) didn't say you do a modular reduction of the clamped integer when computing the pubkey.[^1] So I kept this by the letter of the law.

But this detail doesn't matter! Yes it is true that sometimes `s.reduce() * H != s * H`, but this is only for `H` that are not in the prime subgroup. The only place `scalar_bytes` is used (outside of defining `scalar`) is in the pubkey generation step `EdwardsPoint::mul_base_clamped(scalar_bytes)`. Since the generator is in the prime order subgroup, I should be able to just do `scalar * ED25519_BASEPOINT`.

Does this make sense? Basically I think we should remove `scalar_bytes`. Lmk and I'll push to this branch

tarcieri: #545 already does it as ab alternative to this PR

rozbb:
Superceded by #545

PR #545, opened 2023-07-02
ed25519-dalek: remove `ExpandedSecretKey::to_bytes`

tarcieri:
The reason `ExpandedSecretKey` needs a private `scalar_bytes` field is to retain the canonical scalar bytes as output by SHA-512 during key expansion so they can be serialized by the `to_bytes` method.

However, `ExpandedSecretKey`s should not be serialized to the wire.

Removing this method allows the private field to be removed, which allows `ExpandedSecretKey` to be constructed entirely from public fields. This provides an alternative to #544 for use cases like Ed25519-BIP32 where the private scalar is derived rather than clamped from bytes.

One other change is needed: `to_scalar_bytes` was changed to `to_scalar` as the canonical scalar bytes are no longer retained, however this has no impact on its main use case, X25519 Diffie-Hellman exchanges, where the `Scalar` should NOT be written to the wire anyway.

### DIFF
--- a/ed25519-dalek/tests/x25519.rs
+++ b/ed25519-dalek/tests/x25519.rs
@@ -16,17 +16,8 @@ fn ed25519_to_x25519_dh() {
     let ed25519_signing_key_a = SigningKey::from_bytes(&ed25519_secret_key_a);
     let ed25519_signing_key_b = SigningKey::from_bytes(&ed25519_secret_key_b);
 
-    let scalar_a_bytes = ed25519_signing_key_a.to_scalar_bytes();
-    let scalar_b_bytes = ed25519_signing_key_b.to_scalar_bytes();
-
-    assert_eq!(
-        scalar_a_bytes,
-        hex!("357c83864f2833cb427a2ef1c00a013cfdff2768d980c0a3a520f006904de90f")
-    );
-    assert_eq!(
-        scalar_b_bytes,
-        hex!("6ebd9ed75882d52815a97585caf4790a7f6c6b3b7f821c5e259a24b02e502e11")
-    );
+    let scalar_a = ed25519_signing_key_a.to_scalar();
+    let scalar_b = ed25519_signing_key_b.to_scalar();
 
     let x25519_public_key_a = ed25519_signing_key_a.verifying_key().to_montgomery();
     let x25519_public_key_b = ed25519_signing_key_b.verifying_key().to_montgomery();
@@ -44,11 +35,11 @@ fn ed25519_to_x25519_dh() {
         hex!("5166f24a6918368e2af831a4affadd97af0ac326bdf143596c045967cc00230e");
 
     assert_eq!(
-        x25519_public_key_a.mul_clamped(scalar_b_bytes).to_bytes(),
+        (x25519_public_key_a * scalar_b).to_bytes(),
         expected_shared_secret
     );
     assert_eq!(
-        x25519_public_key_b.mul_clamped(scalar_a_bytes).to_bytes(),
+        (x25519_public_key_b * scalar_a).to_bytes(),
         expected_shared_secret
     );
 }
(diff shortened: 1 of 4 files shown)


tarcieri:
Note: this is an alternative to both #541 and #544

ed25519-dalek/tests/x25519.rs
```diff
@@ -16,17 +16,8 @@ fn ed25519_to_x25519_dh() {
...
-    let scalar_b_bytes = ed25519_signing_key_b.to_scalar_bytes();
-
-    assert_eq!(
-        scalar_a_bytes,
-        hex!("357c83864f2833cb427a2ef1c00a013cfdff2768d980c0a3a520f006904de90f")
-    );
-    assert_eq!(
-        scalar_b_bytes,
-        hex!("6ebd9ed75882d52815a97585caf4790a7f6c6b3b7f821c5e259a24b02e502e11")
-    );
```
tarcieri, line 29: Note: the rest of the D-H workflow is tested, however I removed checking for a particular scalar serialization here since the `Scalar` has been pre-clamped at this point.

We could check for the clamped version and make a note of it in a comment.

tarcieri:
Another advantage of this approach is it leaves fewer copies of the secret scalar in memory.

Otherwise every `ExpandedSecretKey` contains two copies of the secret scalar: one unclamped, one clamped.

burdges:
As an aside, Ed25519-BIP32 has some serious issues, so really it should not be encouraged.   Tor has a similar derivation scheme winds up safer.  cc https://github.com/dalek-cryptography/ed25519-dalek/issues/137

tarcieri:
@burdges AFAICT this would also work with Tor's method which also uses a computed scalar

rozbb:
Copying reasoning from my comment in #544:

The original reason I included `scalar_bytes` was because [section 5.1.5](https://www.rfc-editor.org/rfc/rfc8032#section-5.1.5) didn't say you do a modular reduction of the clamped integer when computing the pubkey. So I kept this by the letter of the law.

But this detail doesn't matter! Yes it is true that sometimes `s.reduce() * H != s * H`, but this is only for `H` that are not in the prime-order subgroup. The only place `scalar_bytes` is used (outside of defining `scalar`) is in the pubkey generation step `EdwardsPoint::mul_base_clamped(scalar_bytes)`. Since the generator is in the prime order subgroup, I should be able to just do `scalar * ED25519_BASEPOINT`.

ed25519-dalek/tests/x25519.rs
rozbb: Fixed.

rozbb (approved):
Added my reasoning above. Looks good to me!

ed25519-dalek/tests/x25519.rs
```diff
@@ -16,16 +25,19 @@ fn ed25519_to_x25519_dh() {
...
-        scalar_a_bytes,
-        hex!("357c83864f2833cb427a2ef1c00a013cfdff2768d980c0a3a520f006904de90f")
+        scalar_a.to_bytes(),
+        clamp_and_reduce(&Sha512::digest(ed25519_secret_key_a)[..32]),
     );
     assert_eq!(
-        scalar_b_bytes,
-        hex!("6ebd9ed75882d52815a97585caf4790a7f6c6b3b7f821c5e259a24b02e502e11")
+        scalar_b.to_bytes(),
+        clamp_and_reduce(&Sha512::digest(ed25519_secret_key_b)[..32]),
```
rozbb, line 40: Manually verified that `Sha512::digest` matches the hex here. I removed the hex because it didn't come from any test vector. This seems like a more principled way of testing things.

rozbb:
```console
$ echo "9d61b19deffd5a60ba844af492ec2cc44449c5697b326919703bac031cae7f60" | xxd -r -p | shasum -a 512 | head -c 64
357c83864f2833cb427a2ef1c00a013cfdff2768d980c0a3a520f006904de90f⏎
$ echo "4ccd089b28ff96da9db6c346ec114e0f5b8a319f35aba624da8cf6ed4fb8a6fb" | xxd -r -p | shasum -a 512 | head -c 64
6ebd9ed75882d52815a97585caf4790a7f6c6b3b7f821c5e259a24b02e502e11⏎
```

PR #546, opened 2023-07-05
Add partial precomputation support

AaronFeickert:
Currently, using precomputation for variable-time multiscalar multiplication requires the number of precomputed points and static scalars to be the same; otherwise, the relevant functions will panic.

This limits some use cases of interest. For example, a [Bulletproofs+](https://github.com/tari-project/bulletproofs-plus/pull/19) range proving implementation was made more efficient by precomputing a large set of curve group generators, which allowed for verification of shorter proofs that don't need to use all of them. (Interestingly, the technique applies equally well to the Bulletproofs range proving system, but that's for another day!)

There are probably several ways to support this, but the most straightforward seems to be simply relaxing the panic condition. This PR does precisely that. Providing a smaller number of static scalars will simply use only the corresponding precomputed points when evaluating a multiscalar multiplication. Documentation is updated accordingly.

Comments welcome!

### DIFF
--- a/curve25519-dalek/src/backend/serial/scalar_mul/precomputed_straus.rs
+++ b/curve25519-dalek/src/backend/serial/scalar_mul/precomputed_straus.rs
@@ -75,7 +75,7 @@ impl VartimePrecomputedMultiscalarMul for VartimePrecomputedStraus {
 
         let sp = self.static_lookup_tables.len();
         let dp = dynamic_lookup_tables.len();
-        assert_eq!(sp, static_nafs.len());
+        assert!(sp >= static_nafs.len());
         assert_eq!(dp, dynamic_nafs.len());
 
         // We could save some doublings by looking for the highest
@@ -99,7 +99,7 @@ impl VartimePrecomputedMultiscalarMul for VartimePrecomputedStraus {
             }
 
             #[allow(clippy::needless_range_loop)]
-            for i in 0..sp {
+            for i in 0..static_nafs.len() {
                 let t_ij = static_nafs[i][j];
                 match t_ij.cmp(&0) {
                     Ordering::Greater => {
--- a/curve25519-dalek/src/backend/vector/scalar_mul/precomputed_straus.rs
+++ b/curve25519-dalek/src/backend/vector/scalar_mul/precomputed_straus.rs
@@ -83,7 +83,7 @@ pub mod spec {
 
             let sp = self.static_lookup_tables.len();
             let dp = dynamic_lookup_tables.len();
-            assert_eq!(sp, static_nafs.len());
+            assert!(sp >= static_nafs.len());
             assert_eq!(dp, dynamic_nafs.len());
 
             // We could save some doublings by looking for the highest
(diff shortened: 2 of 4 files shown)


AaronFeickert:
Marked as draft while debugging tests that fail CI.

elichai:
> Adding `Clone` support makes it possible to do things like `Arc` wrapping, where derived cloning (which doesn't actually clone the wrapped type) requires `Clone` support all the way down. While you'd almost certainly never want to actually clone a set of tables, this seemed like a much easier way to support reuse without a major refactor.

Arc doesn't require that `T: Clone` See the [implementation](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-Clone-for-Arc%3CT%3E) and an [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f3862c2539133e60e92b656e8abe853c)

AaronFeickert:
The specific context I was considering involved the use of a generic `Arc` whose trait bound includes `VartimePrecomputedMultiscalarMul`, and which does require underlying `Clone` support or a manual `Clone` implementation.

elichai:
> The specific context I was considering involved the use of a generic `Arc` whose trait bound includes `VartimePrecomputedMultiscalarMul`, and which does require underlying `Clone` support or a manual `Clone` implementation for a containing struct.

Could you give a code example for this pattern?

AaronFeickert:
Apologies that it's not a MWE, but [this pull request](https://github.com/tari-project/bulletproofs-plus/pull/50) helps to demonstrate the issue. A [derived clone](https://github.com/tari-project/bulletproofs-plus/pull/50/files#diff-c3cc77f6a19b0028bd7c97e9ab648a795d79f683193d404c250cedbaf582a35cL40) of a struct containing an `Arc` with a `VartimePrecomputedMultiscalarMul` [trait bound](https://github.com/tari-project/bulletproofs-plus/pull/50/files#diff-172747fc43e38d33de743b2d5f564ad6440570670af7aa912235399448a55c7cL42) apparently must also enforce `Clone` on the underlying precomputation struct.

The alternate solution shown in that PR is to do manual `Arc` cloning, which works as expected. Apparently it's the generic with trait bound that screws up the derived cloning. It seemed more straightforward to me to support `Clone` at the curve library level, but I can understand a desire to avoid cloning due to table sizes.

Anyway, happy to remove `Clone` if it's considered a blocker here.

AaronFeickert:
Is cloning holding this up? Happy to remove it if so.

tarcieri:
@AaronFeickert there's a large backlog of PRs it's going to take awhile to work through

AaronFeickert:
Removed cloning and rebased.

AaronFeickert:
Any updates on this? Would appreciate any feedback from @tarcieri _et al._ as time permits.

AaronFeickert:
This has been rebased to be up to date against recent changes.

Would greatly appreciate brief review from @tarcieri or another maintainer on whether or not this change seems reasonable for merge.

PR #550, opened 2023-07-22
curve: Bump version to 4.0.0

rozbb:
I think we're ready to go


rozbb:
It looks like this won't compile unless I bump all the other `Cargo.toml`s too

tarcieri (approved):

PR #554, opened 2023-07-27
Stabilize x25519-dalek version 2

elichai:
Now that `curve25519-dalek` version 4 has been released and I couldn't find any outstanding PRs/Issues can version 2 of `x25519-dalek` be released?

Also, can https://github.com/dalek-cryptography/x25519-dalek redirect to here? It's pretty confusing


robjtede:
looks like this is being superseded by #560

rozbb:
Can you rebase on latest `main`?

rozbb:
Actually I'll just merge. The clippy lints are fixed in main so it shouldn't break the build. If it does I'll fix it.

rozbb:
Thank you for this! And good catch on the deprecated constructor

PR #555, opened 2023-07-30
curve25519: Add arbitrary integer multiplication with `MontgomeryPoint::mul_bits_be`

rozbb:
There is occasionally [a need](https://github.com/dalek-cryptography/curve25519-dalek/pull/519#issuecomment-1637770888) to multiply a non-prime-order Montgomery point by an integer. There's currently no way to do this, since our only methods are multiplication by `Scalar` (doesn't make sense in the non-prime-order case), and `MontgomeryPoint::mul_base_clamped` clamps the integer before multiplying.

So we define `MontgomeryPoint::mul_bits_be`, gated behind `hazmat`, which takes a big-endian representation of an integer and multiplies the point by that integer.

cc @elichai

### DIFF
--- a/curve25519-dalek/src/montgomery.rs
+++ b/curve25519-dalek/src/montgomery.rs
@@ -117,6 +117,18 @@ impl Zeroize for MontgomeryPoint {
     }
 }
 
+/// Given a bytestring that's little-endian at the byte level, return an iterator over all the
+/// bits, in little-endian order.
+pub(crate) fn bytestring_bits_le(x: &[u8]) -> impl DoubleEndedIterator<Item = bool> + '_ {
+    let bitlen = x.len() * 8;
+    (0..bitlen).map(|i| {
+        // As i runs from 0..256, the bottom 3 bits index the bit, while the upper bits index
+        // the byte. Since self.bytes is little-endian at the byte level, this iterator is
+        // little-endian on the bit level
+        ((x[i >> 3] >> (i & 7)) & 1u8) == 1
+    })
+}
+
 impl MontgomeryPoint {
     /// Fixed-base scalar multiplication (i.e. multiplication by the base point).
     pub fn mul_base(scalar: &Scalar) -> Self {
@@ -126,17 +138,12 @@ impl MontgomeryPoint {
     /// Multiply this point by `clamp_integer(bytes)`. For a description of clamping, see
     /// [`clamp_integer`].
     pub fn mul_clamped(self, bytes: [u8; 32]) -> Self {
-        // We have to construct a Scalar that is not reduced mod l, which breaks scalar invariant
-        // #2. But #2 is not necessary for correctness of variable-base multiplication. All that
-        // needs to hold is invariant #1, i.e., the scalar is less than 2^255. This is guaranteed
-        // by clamping.
-        // Further, we don't do any reduction or arithmetic with this clamped value, so there's no
-        // issues arising from the fact that the curve point is not necessarily in the prime-order
-        // subgroup.
-        let s = Scalar {
-            bytes: clamp_integer(bytes),
-        };
-        s * self
+        // Clamp the integer, convert to bits, and multiply
+        let clamped = clamp_integer(bytes);
+        // Clamping sets the the MSB 0, so we skip it
+        let le_bits = bytestring_bits_le(&clamped).rev().skip(1);
+
+        self._mul_bits_be(le_bits)
     }
 
     /// Multiply the basepoint by `clamp_integer(bytes)`. For a description of clamping, see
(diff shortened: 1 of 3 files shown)


curve25519-dalek/src/montgomery.rs
```diff
@@ -505,22 +521,60 @@ mod test {
...
 
+    // Tests that point._mul_bits_be(bits) is the same as multiplying by the Scalar representation
+    // of the bits.
+    #[test]
+    fn montgomery_mul_bits_be() {
+        let mut csprng = rand_core::OsRng;
+
+        for _ in 0..100 {
+            let p_edwards = rand_point(&mut csprng);
+            let p_montgomery: MontgomeryPoint = p_edwards.to_montgomery();
```
elichai, line 565: I think it's also worth testing with points on the curve's twist and points not in the prime subgroup

elichai (approved):
ACK ed53307c165d6fa5acb105829a5b253717362da1 reviewed the code and it seems correct, haven't tested it yet

curve25519-dalek/src/montgomery.rs
rozbb: Added another test

rozbb:
@tarcieri ready for review. One thing I wanna make sure of personally is that this hasn't strayed too far from Alg 8 of [CS17](https://arxiv.org/pdf/1703.01863.pdf)

curve25519-dalek/src/montgomery.rs
```diff
@@ -394,6 +391,25 @@ impl<'a, 'b> Mul<&'b Scalar> for &'a MontgomeryPoint {
...
+    /// Given `self` \\( = u\_0(P) \\), and a big-endian bit representation of an integer
+    /// \\(n\\), return \\( u\_0(\[n\]P) \\).
+    #[cfg(feature = "hazmat")]
+    fn mul_bits_be(self, bits: impl Iterator<Item = bool>) -> MontgomeryPoint {
+        self.mul_bits_be(bits)
+    }
+}
+
+/// Multiply this `MontgomeryPoint` by a `Scalar`.
+impl<'a, 'b> Mul<&'b Scalar> for &'a MontgomeryPoint {
```
tarcieri, line 404: I think all of these lifetimes can be elided?
```suggestion
impl Mul<&Scalar> for &MontgomeryPoint {
```

curve25519-dalek/src/montgomery.rs
```diff
@@ -394,6 +391,25 @@ impl<'a, 'b> Mul<&'b Scalar> for &'a MontgomeryPoint {
 
         x0.as_affine()
     }
+
+    /// Given `self` \\( = u\_0(P) \\), and a big-endian bit representation of an integer
+    /// \\(n\\), return \\( u\_0(\[n\]P) \\).
+    #[cfg(feature = "hazmat")]
+    fn mul_bits_be(self, bits: impl Iterator<Item = bool>) -> MontgomeryPoint {
+        self.mul_bits_be(bits)
```
tarcieri, line 399: Is this supposed to be `pub`?

It looks like it calls itself and infinitely recurses? Is it supposed to call `self._mul_bits_be`?

If the intent is for this to be `pub`, I worry about people accidentally discovering it. Feature-gating on `hazmat` alone isn't a great deterrent since other dependencies can activate the feature. Putting it in the `hazmat` module would better prevent accidental discovery/misuse IMO (and free up the `mul_bits_be` name so you don't need a `_`-prefixed version)

At the very least, it could use a more stringent warning label.

rozbb: Oh good points. I agree it'd be best to put it in the `hazmat` module. Will fix.

rozbb: Acutally, this is a question: should this method be feature-gated? There's two main misuses I can think of:

1. If someone used this method instead of `mul_clamped`. This could leak up to 3 secret bits. Not great.
2. If someone used this method instead of `Mul<Scalar> for MontgomeryPoint`. This would be fine, actually. In the worst case, this is equivalent to doing the scalar mul. In the best case, this is _more_ correct than the scalar mul, because the point being multiplied might not be in the prime-order subgroup, and so the scalar mul made no sense anyway.

So the question is if case (1) is likely and worth avoiding. I'm not sure. If someone wanted to multiply by some arbitrary bytestring today, they can already do `Scalar::from_bytes_mod_order(bytes) * point`. If the bytestring is uniform, then this is already equivalent to `mul_bits_be` with 50% probability.

My vote is to let this function be public, and let it have the documentation note I just added.

curve25519-dalek/src/montgomery.rs
```diff
@@ -151,6 +151,43 @@ impl MontgomeryPoint {
...
     }
 
+    /// Given `self` \\( = u\_0(P) \\), and a big-endian bit representation of an integer
+    /// \\(n\\), return \\( u\_0(\[n\]P) \\). This is constant time in the length of `bits`.
+    ///
+    /// **NOTE:** You probably do not want to use this function. Almost every protocol built on
+    /// Curve25519 uses _clamped multiplication_, explained
+    /// [here](https://neilmadden.blog/2020/05/28/whats-the-curve25519-clamping-all-about/).
+    /// When in doubt, use [`Self::mul_clamped`].
+    pub fn mul_bits_be(self, bits: impl Iterator<Item = bool>) -> MontgomeryPoint {
```
elichai, line 161: Do we care that this doesn't enforce how many bits are passed? ie you can pass an empty iter and it will be equivalent to multiplying by zero, or you can pass a longer iterator and compute a "longer" multiplication

elichai: Should it use `ExactSizeIterator` and assert len == 256? that would still allow multiplying by something bigger than the curve's order though

curve25519-dalek/src/montgomery.rs
elichai: I have a wild suggestion, how do people think about a `RingElement` type that is over the full curve's order? is this too much for that kind of feature?

curve25519-dalek/src/montgomery.rs
tarcieri: I do like how it's infallible now.

Perhaps it could have a `debug_assert!`? That wouldn't necessarily require `ExactSizeIterator`, it would just need a sort of postcondition check that it iterated over 256 bits.

curve25519-dalek/src/montgomery.rs
rozbb: RingElement is overkill imo. We don't define a way to add or multiply two numbers mod 8p, so it'd be a pretty useless ring element. Integer multiplication is broad enough for this use case imo

curve25519-dalek/src/montgomery.rs
rozbb: Why limit the function at all? Multiplication by arbitrary integers is always well-defined. If you don't want to reduce then you're free not to

PR #557, opened 2023-07-31
ConstantTimeEq and PartialEq for SigningKey

mkj:
Moved over from https://github.com/dalek-cryptography/ed25519-dalek/pull/306


rozbb:
No harm I suppose. And `subtle` is already a dep in curve- so it's not a big change. Thanks!

PR #558, opened 2023-08-07
Mark scalar::clamp_integer as must_use

moiseev-signal:


rozbb (approved):
Makes sense to me. If you're running this function, you probably really don't want to accidentally throw away the result.

pinkforest:
If I would think SemVer pedantically - I would think this is a breaking API behaviour change -

Since it's be damned if you don't and be damned if you do situation since SemVer isn't always practical :)

And since it's still a mandatory compiler warning at the build stage and not an optional clippy dev-only lint.

Some builds might have deny all warnings and if we add a warning those will break - however broken otherwise.

If people agree with me on the pedantry re: breaking - then I would suggest;

For 4.x I would suggest adding `#[cfg_attr(curve25519_dalek_warn_unused), must_use)]` and adopt as default in 5.x

That said - 

Should probably check and add them to all similarly used fn's for the sake of consistency at the same release regardless.

robjtede:
> For 4.x I would suggest adding #[cfg_attr(curve25519_dalek_warn_unused), must_use)] and adopt as default in 5.x

TBH, this would be overkill. std itself adds `must_use` annotations all the time; it's not considered breaking.

pinkforest:
Never seen in a patch release though afaik, at very least it should be a minor bump - maybe this is enough of a heads up :P

pinkforest:
That makes sense yes, I was being pedantic for no reason :sweat_smile:

robjtede (approved):

rozbb:
Sorry, I might have been hasty on this. I didn't realize @tarcieri hadn't approved. Should I roll back?

tarcieri:
I forgot to hit approve, sorry. I agreed in spirit in my comment

PR #562, opened 2023-08-15
Implement `ff` and `group` traits - Rebased

pinkforest:
Did some quick git dark arts :magic_wand: to re-base #473 and resolve the merge conflicts

Couple of warnings needs to be still addressed


curve25519-dalek/Cargo.toml
```diff
@@ -65,6 +66,7 @@ default = ["alloc", "precomputed-tables", "zeroize"]
 alloc = ["zeroize?/alloc"]
 precomputed-tables = []
 legacy_compatibility = []
+group = ["group_crate", "rand_core"]
```
tarcieri, line 69: No need to rename the crate:

```suggestion
group = ["dep:group", "rand_core"]
```

curve25519-dalek/src/scalar.rs
```diff
@@ -1201,6 +1207,126 @@ impl UnpackedScalar {
...
+        Self::from_canonical_bytes(repr)
+    }
+
+    fn from_repr_vartime(repr: Self::Repr) -> Option<Self> {
+        // Check that the high bit is not set
+        if (repr[31] >> 7) != 0u8 {
+            return None;
+        }
+
+        let candidate = Scalar::from_bits(repr);
```
pinkforest, line 1265: ditto

tarcieri: You don't need to gate via `legacy_compatibility` since this still ensures the scalar is fully reduced with the highest bit unset, it just does it in variable time

curve25519-dalek/src/edwards.rs
```diff
@@ -183,25 +190,50 @@ impl CompressedEdwardsY {
     ///
     /// Returns `None` if the input is not the \\(y\\)-coordinate of a
     /// curve point.
-    #[rustfmt::skip] // keep alignment of explanatory comments
     pub fn decompress(&self) -> Option<EdwardsPoint> {
-        let Y = FieldElement::from_bytes(self.as_bytes());
+        let (is_valid_y_coord, X, Y, Z) = decompress::step_1(self);
+        if is_valid_y_coord.unwrap_u8() != 1u8 {
```
tarcieri, line 195:
```suggestion
        if !is_valid_y_coord.into() {
```

curve25519-dalek/src/ristretto.rs
```diff
@@ -246,6 +253,27 @@ impl CompressedRistretto {
     ///
     /// - `None` if `self` was not the canonical encoding of a point.
     pub fn decompress(&self) -> Option<RistrettoPoint> {
+        let (s_encoding_is_canonical, s_is_negative, s) = decompress::step_1(self);
+
+        if s_encoding_is_canonical.unwrap_u8() == 0u8 || s_is_negative.unwrap_u8() == 1u8 {
```
tarcieri, line 258:
```suggestion
        if (!s_encoding_is_canonical | s_is_negative).into() {
```

curve25519-dalek/src/edwards.rs
```diff
@@ -183,25 +190,50 @@ impl CompressedEdwardsY {
...
     /// Returns `None` if the input is not the \\(y\\)-coordinate of a
     /// curve point.
-    #[rustfmt::skip] // keep alignment of explanatory comments
     pub fn decompress(&self) -> Option<EdwardsPoint> {
-        let Y = FieldElement::from_bytes(self.as_bytes());
+        let (is_valid_y_coord, X, Y, Z) = decompress::step_1(self);
+        if (!is_valid_y_coord).into() {
+            return None;
+        }
+        Some(decompress::step_2(self, X, Y, Z))
```
tarcieri, line 198: This might be a bit clearer:

```suggestion
        if is_valid_y_coord.into() {
            Some(decompress::step_2(self, X, Y, Z))
        } else {
            None
        }
```

pinkforest:
Should we shift the trait impls into separate file/s - these files are getting big as noted by @rozbb https://github.com/dalek-cryptography/curve25519-dalek/pull/473#pullrequestreview-1230570618

Also should we refer to the [PrimeField#associatedconstant.S docs](https://docs.rs/ff/latest/ff/trait.PrimeField.html#associatedconstant.S) someway somewhere perhaps

tarcieri:
Just to get it landed I'd propose keeping it close to #473 to simplify review.

Refactoring can happen in follow-up PRs.

curve25519-dalek/src/edwards.rs
```diff
@@ -1236,6 +1270,318 @@ impl Debug for EdwardsPoint {
...
+    }
+}
+
+#[cfg(feature = "group")]
+impl<'a, 'b> Add<&'b SubgroupPoint> for &'a SubgroupPoint {
+    type Output = SubgroupPoint;
+    fn add(self, other: &'b SubgroupPoint) -> SubgroupPoint {
+        SubgroupPoint(self.0 + other.0)
+    }
+}
```
tarcieri, line 1360: I don't think there's any need for explicit lifetimes on any of these impls, and in fact `'a` is completely unused.

They can be elided, both here and on the `Add`, `Sub`, and `Mul` impls below:

```suggestion
impl Add<&SubgroupPoint> for &SubgroupPoint {
    type Output = SubgroupPoint;
    fn add(self, other: &SubgroupPoint) -> SubgroupPoint {
        SubgroupPoint(self.0 + other.0)
    }
}
```

tarcieri (approved):

rozbb:
Thank you @tarcieri! Been extremely busy lately (proposing thesis). Also idk who did it but the 2-stage decompression trick is really slick.

PR #566, opened 2023-08-22
Adapt to new types introduced in fiat-crypto 0.2

divergentdave:
I have an outstanding PR that proposes some breaking changes to fiat-crypto, see mit-plv/fiat-crypto#1623. Briefly, it introduces new structures to replace different type aliases for each field element representation. Encoding this in the type system is intended to prevent misuse of functions that require "tight" preconditions on their inputs' ranges. I'm opening this PR for a future crate upgrade early, with a `[patch.crates-io]` stanza, to test out the breaking change, and for temporary use in fiat-crypto's integration tests. If this looks good, I can back out the patch stanza when `fiat-crypto` is released and mark it ready for review.

### DIFF
--- a/curve25519-dalek/Cargo.toml
+++ b/curve25519-dalek/Cargo.toml
@@ -59,7 +59,7 @@ zeroize = { version = "1", default-features = false, optional = true }
 cpufeatures = "0.2.6"
 
 [target.'cfg(curve25519_dalek_backend = "fiat")'.dependencies]
-fiat-crypto = { version = "0.1.19", default-features = false }
+fiat-crypto = { version = "0.2.1", default-features = false }
 
 [features]
 default = ["alloc", "precomputed-tables", "zeroize"]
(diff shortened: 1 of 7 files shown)


divergentdave:
The new fiat-crypto version is out, so this is ready now.

curve25519-dalek/Cargo.toml
```diff
@@ -59,7 +59,7 @@ zeroize = { version = "1", default-features = false, optional = true }
 cpufeatures = "0.2.6"
 
 [target.'cfg(curve25519_dalek_backend = "fiat")'.dependencies]
-fiat-crypto = "0.1.19"
+fiat-crypto = "0.2.0"
```
pinkforest, line 62:
```suggestion
fiat-crypto = "0.2.1"
```

Re: https://github.com/mit-plv/fiat-crypto/pull/1646 - 

Note to myself: We'll need to test no_std with fiat / other backends and I'm sending proper no_std test to fiat as well

pinkforest:
Thanks :tada:  - I'm benching this atm for any regressions - I noted there was `no_std` issue per above

tarcieri:
@divergentdave can you rebase now that #572 has been merged?

tarcieri (approved):

dconnolly:
🎉

tarcieri:
Merging so we can get this in #574

PR #571, opened 2023-09-01
Allow viewing secret as byte array

dpogorzelski:
This will make it possible to view the secret as byte array which can be handy when printing contents for "debug" purposes via crates such as https://crates.io/crates/pretty-hex


pinkforest:
The current API for `EphemeralSecret` is very intentional to not provide anything that exposes the secret key material.

Users can opt-in to `StaticSecret `via `static_secrets` feature that provides this footgun if they need it and opts in to it but it will not provide forward secrecy as is intent on Ephemeral DH which purpose is to provide forward secrecy.

dpogorzelski:
This makes sense, thanks for the clarification.

PR #572, opened 2023-09-04
curve: Fix `no_std` for `fiat` backend and add test for it

pinkforest:
Fiat backend [required std](https://github.com/dalek-cryptography/curve25519-dalek/actions/runs/6073028741/job/16474141859?pr=572) as we didn't call it with `default-features - false` - I've changed it here.

We only test the serial (default) backend in `no_std` that gets tested changes on any crates

This adds fiat backend specific no_std test when there is/are changes on curve25519-dalek in addition.

Given https://github.com/mit-plv/fiat-crypto/pull/1646 it may be feasible to fiat is tested as well

This should be merged and re-based into the fiat-backend 0.2 bump after this has been tested with 0.1 via this PR


tarcieri:
Suddenly I'm confused why there's both a `no_std.yml` workflow file and `no_std` jobs within `curve25519-dalek.yml`

pinkforest:
Yeah no_std gets applied to whole workspace and curve25519 - or do we want to test both fiat and serial no_std across all crates ? imo redundant - maybe we can rename the no_std.yml to workspace-no_std.yml ?

tarcieri (approved):

PR #579, opened 2023-09-12
Add PrimeFieldBits support to Scalar

kayabaNerve:
The char_le_bits function is a hack I don't legitimately recommend. I just felt it'd be better to make a PR, which works, than to solely file an issue requesting it as a feature.

### DIFF
--- a/curve25519-dalek/Cargo.toml
+++ b/curve25519-dalek/Cargo.toml
@@ -48,7 +48,7 @@ required-features = ["alloc", "rand_core"]
 
 [dependencies]
 cfg-if = "1"
-group = { version = "0.13", default-features = false, optional = true }
+group = { version = "0.13", default-features = false, features = ["bits"], optional = true }
 rand_core = { version = "0.6.4", default-features = false, optional = true }
 digest = { version = "0.10", default-features = false, optional = true }
 subtle = { version = "2.3.0", default-features = false }
--- a/curve25519-dalek/Cargo.toml
+++ b/curve25519-dalek/Cargo.toml
@@ -27,7 +27,7 @@ rustdoc-args = [
     "--html-in-header", "docs/assets/rustdoc-include-katex-header.html",
     "--cfg", "docsrs",
 ]
-features = ["serde", "rand_core", "digest", "legacy_compatibility", "group"]
+features = ["serde", "rand_core", "digest", "legacy_compatibility", "group-bits"]
 
 [dev-dependencies]
 sha2 = { version = "0.10", default-features = false }
@@ -48,7 +48,7 @@ required-features = ["alloc", "rand_core"]
 
 [dependencies]
 cfg-if = "1"
-ff = { version = "0.13", default-features = false, features = ["bits"], optional = true }
+ff = { version = "0.13", default-features = false, optional = true }
 group = { version = "0.13", default-features = false, optional = true }
 rand_core = { version = "0.6.4", default-features = false, optional = true }
 digest = { version = "0.10", default-features = false, optional = true }
@@ -68,6 +68,7 @@ alloc = ["zeroize?/alloc"]
 precomputed-tables = []
 legacy_compatibility = []
 group = ["dep:group", "rand_core"]
+group-bits = ["group", "ff/bits"]
 
 [target.'cfg(all(not(curve25519_dalek_backend = "fiat"), not(curve25519_dalek_backend = "serial"), target_arch = "x86_64"))'.dependencies]
 curve25519-dalek-derive = { version = "0.1", path = "../curve25519-dalek-derive" }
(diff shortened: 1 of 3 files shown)


curve25519-dalek/src/scalar.rs
```diff
@@ -1321,6 +1321,22 @@ impl PrimeField for Scalar {
...
+
+  fn to_le_bits(&self) -> FieldBits<Self::ReprBits> {
+    self.to_repr().into()
+  }
+
+  fn char_le_bits() -> FieldBits<Self::ReprBits> {
+    let mut bytes = (Scalar::ZERO - Scalar::ONE).to_repr();
+    bytes[0] += 1;
+    debug_assert_eq!(DScalar::from_bytes_mod_order(bytes), DScalar::ZERO);
+    bytes.into()
```
tarcieri, line 1336: I think this should work but haven't tested it

```suggestion
    crate::constants::BASEPOINT_ORDER.into()
```

kayabaNerve: I'll test it now, thanks.

kayabaNerve: That's a Scalar, not a byte array. Doesn't that break safety to have an unreduced Scalar in system? Especially now that Scalar addition assumes only unreduced Scalars exist?

kayabaNerve: to_bytes().into() works, will push that.

tarcieri: > Doesn't that break safety to have an unreduced Scalar in system?

Yeah, seems a bit odd to have it in the public API if it's unreduced /cc @rozbb

kayabaNerve:
Sorry for this PR being a mess. I'm unsure why the no-std CIs are failing.

kayabaNerve:
I use ff, with bits, on libraries with no-std targets so that shouldn't be it... (though I may technically solely test them on alloc, not no-std).

I'll make bits a feature.

pinkforest:
It's because we don't do additional cargo hack of the features via std build and we only do it via no_std that detects any errors via features :)

curve25519-dalek/Cargo.toml
```diff
@@ -67,6 +68,7 @@ alloc = ["zeroize?/alloc"]
 precomputed-tables = []
 legacy_compatibility = []
 group = ["dep:group", "rand_core"]
+group-bits = ["group", "ff/bits"]
```
tarcieri, line 71: FWIW the @RustCrypto crates just call this `bits`. I find `group-bits` a bit weird because the `bits` part is only accidentally dependent on `group` because that's how we import `ff`, but I guess it does get across that it activates the `group` feature too.

kayabaNerve: My preference is scoping it since group itself is optional, whereas RustCrypto crates are traditionally built around group. If `bits` is the way forward, please feel free to say so (as I'm inclined to do nothing when a decision on naming is ambiguous as it is right now).

tarcieri: Just a suggestion, I don't care strongly either way

kayabaNerve: @pinkforest for a third opinion? ^_^

pinkforest: `ff-bits` :)

curve25519-dalek/src/scalar.rs
pinkforest: `BASEPOINT_ORDER` will be marked deprecated from 5.0.0 - this will be `BASEPOINT_ORDER_PRIVATE` as pub(crate)
- https://github.com/dalek-cryptography/curve25519-dalek/pull/581

rozbb:
This looks good to me, thanks! Could you add an item in the changelog for this feature?

pinkforest:
Please rebase

curve25519-dalek/src/scalar.rs
```diff
@@ -1321,6 +1323,19 @@ impl PrimeField for Scalar {
...
+#[cfg(feature = "group-bits")]
+impl PrimeFieldBits for Scalar {
+    type ReprBits = [u8; 32];
+
+    fn to_le_bits(&self) -> FieldBits<Self::ReprBits> {
+        self.to_repr().into()
+    }
+
+    fn char_le_bits() -> FieldBits<Self::ReprBits> {
+        constants::BASEPOINT_ORDER.to_bytes().into()
```
pinkforest, line 1335:
```suggestion
        constants::BASEPOINT_ORDER_PRIVATE.to_bytes().into()
```
+rebase pls :)

kayabaNerve:
Thanks for moving this forward before I had the chance, rozbb :) And thanks pinkforest for the actual edit needed.

PR #581, opened 2023-09-12
Deprecate `BASEPOINT_ORDER` from pub API consts

pinkforest:
Deprecates `BASEPOINT_ORDER` from pub API constants

Redirect internally to avoid deprecation errors

### DIFF
--- a/curve25519-dalek/src/constants.rs
+++ b/curve25519-dalek/src/constants.rs
@@ -8,24 +8,7 @@
 // Authors:
 // - isis agora lovecruft <isis@patternsinthevoid.net>
 // - Henry de Valence <hdevalence@hdevalence.ca>
-
 //! Various constants, such as the Ristretto and Ed25519 basepoints.
-//!
-//! Most of the constants are given with
-//! `LONG_DESCRIPTIVE_UPPER_CASE_NAMES`, but they can be brought into
-//! scope using a `let` binding:
-//!
-#![cfg_attr(feature = "precomputed-tables", doc = "```")]
-#![cfg_attr(not(feature = "precomputed-tables"), doc = "```ignore")]
-//! use curve25519_dalek::constants;
-//! use curve25519_dalek::traits::IsIdentity;
-//!
-//! let B = constants::RISTRETTO_BASEPOINT_TABLE;
-//! let l = &constants::BASEPOINT_ORDER;
-//!
-//! let A = l * B;
-//! assert!(A.is_identity());
-//! ```
 
 #![allow(non_snake_case)]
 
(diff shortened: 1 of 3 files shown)


tarcieri (approved):

curve25519-dalek/src/constants.rs
```diff
@@ -94,6 +78,9 @@ pub const BASEPOINT_ORDER: Scalar = Scalar {
     ],
 };
 
+#[allow(deprecated)]
+pub(crate) const BASEPOINT_ORDER_PRIVATE: Scalar = BASEPOINT_ORDER;
```
tarcieri, line 82: Maybe reverse this relationship so `BASEPOINT_ORDER_PRIVATE` defines the constant and `BASEPOINT_ORDER` points to `BASEPOINT_ORDER_PRIVATE`?

pinkforest: ah ofcourse, thanks :purple_heart:

tarcieri (changes requested):
It seems like there are a lot of unrelated changes in this PR now?

pinkforest:
Yes, I forgot to rebase, will do now.

pinkforest:
Fixed - Forgot to rebase

tarcieri (approved):

curve25519-dalek/src/constants.rs
```diff
@@ -86,7 +69,10 @@ pub const RISTRETTO_BASEPOINT_POINT: RistrettoPoint = RistrettoPoint(ED25519_BAS
 /// $$
 /// \ell = 2^\{252\} + 27742317777372353535851937790883648493.
 /// $$
-pub const BASEPOINT_ORDER: Scalar = Scalar {
+#[deprecated(since = "5.0.0", note = "Should not have been in public API")]
```
tarcieri, line 72: Shouldn't this be whatever the next release is, e.g.

```suggestion
#[deprecated(since = "4.1.1", note = "Should not have been in public API")]
```

pinkforest: True, thanks

tarcieri (approved):

marshallpierce:
I have a use for this constant (I need to exponentiate a scalar by `order - 2`) -- would it be objectionable to un-deprecate this? Otherwise I'll need to construct my own copy of a constant, which would be a shame.

trevor-crypto:
@marshallpierce what was the solution? :thinking: Just making your own constant?

marshallpierce:
It turned out I was doing curve operations in a needlessly complex way. I don't have the details in front of me, but tl;dr I found a way to do what I needed to do using other public parts of the API.

PR #583, opened 2023-09-12
Sign/verify by digest update, StreamVerifier refactoring

pinkforest:
This replaces https://github.com/dalek-cryptography/ed25519-dalek/pull/304
Rebases #556

@mkj wrote:

I'd like to be able to sign/verify non-prehash signatures without the whole message in memory. The use case is for running on no_std embedded where the message is serialized directly into the sha512 digest. It's for SSH protocol so I can't use ed25519 prehashed.

The `StreamVerifier` pull request https://github.com/dalek-cryptography/curve25519-dalek/pull/542 provides similar functionality, though streaming is only possible for verify (signing needs two passes). Instead I've added `raw_sign_byupdate()` and `raw_verify_byupdate()` that take a closure to update the message digest.

I've included the `StreamVerifier` commit from https://github.com/dalek-cryptography/curve25519-dalek/pull/542 and moved `recompute_R` into its own struct `RCompute`. That lets all the verifier options use the same code path.

`_byupdate` isn't the best name, but other names I came up with would get confused with prehashed methods. I'm open to other suggestions.

### DIFF
--- a/ed25519-dalek/CHANGELOG.md
+++ b/ed25519-dalek/CHANGELOG.md
@@ -34,6 +34,7 @@ Entries are listed in reverse chronological order per undeprecated major series.
 * Add `pkcs` feature to support PKCS #8 (de)serialization of `SigningKey` and `VerifyingKey`
 * Add `fast` feature to include basepoint tables
 * Add tests for validation criteria
+* Add `SigningKey::verify_stream()`, and `VerifyingKey::verify_stream()`
 * Impl `DigestSigner`/`DigestVerifier` for `SigningKey`/`VerifyingKey`, respectively
 * Impl `Hash` for `VerifyingKey`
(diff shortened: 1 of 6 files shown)


robjtede (approved):

ed25519-dalek/CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ Entries are listed in reverse chronological order per undeprecated major series.
 * Add `pkcs` feature to support PKCS #8 (de)serialization of `SigningKey` and `VerifyingKey`
 * Add `fast` feature to include basepoint tables
 * Add tests for validation criteria
+* Add `SigningKey::verify_stream()`, and `VerifyingKey::verify_stream()`
```
robjtede, line 37: this entry should probably be moved up out of the 2.0.0 section

pinkforest:
I didn't end up having much time but looks like the original branch went forward :tada: 
- #556

PR #585, opened 2023-09-29
Avoid cargo test -p curve25519-dalek-derive failure on ARM

nategraf:
On a MacBook M1, I noticed `cargo test` in the workspace root fails. It fails due to the references to `#[target_feature(enable = "avx2")]` in the `README.md` for `curve25519-dalek-derive`.

This PR avoids the test failure by adding conditional compile statements on the usages of `target_feature`. `cfg` statements are added as hidden lines in the examples and so they will not appear in `cargo doc`. Hidden lines will appear when looking at the README on GitHub.

### DIFF
--- a/curve25519-dalek-derive/README.md
+++ b/curve25519-dalek-derive/README.md
@@ -44,6 +44,7 @@ to build out more elaborate abstractions it starts to become painful to use.
     }
 
     struct AVX;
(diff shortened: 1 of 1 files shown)


nategraf:
Let me know if this seems like the right solution, or if you'd like me to rework it.

tarcieri:
Seems fine to me

tarcieri (approved):

PR #590, opened 2023-10-29
Fix licensing on -derive repo

rozbb:
The `curve25591-dalek-derive` repo is just a vendored copy of `[unsafe_target_feature](https://github.com/koute/unsafe_target_feature)` (see https://github.com/dalek-cryptography/curve25519-dalek/pull/531). As such I have copied over those license files.

### DIFF
--- a/curve25519-dalek-derive/LICENSE-APACHE
+++ b/curve25519-dalek-derive/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+   "License" shall mean the terms and conditions for use, reproduction,
+   and distribution as defined by Sections 1 through 9 of this document.
+
+   "Licensor" shall mean the copyright owner or entity authorized by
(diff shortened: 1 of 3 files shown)


rozbb:
@koute it appears that there is no copyright line on either of these files (here or in your repo). Would you like to add a copyright line crediting yourself? Let me know what to write.

koute:
> @koute it appears that there is no copyright line on either of these files (here or in your repo). Would you like to add a copyright line crediting yourself? Let me know what to write.

I usually don't add explicit copyright lines since legally it makes no difference and practically it's mostly useless (unless you're going to add *every* contributor, in which case it can act as a list of everyone who has copyright over a given work, although that's kinda redundant if you have the git history) and [can sometimes be a bad idea](https://daniel.haxx.se/blog/2021/02/19/i-will-slaughter-you/), but if you do want to add an explicit copyright line feel free to use the name of my employer: `Parity Technologies (UK) Ltd.`

rozbb:
Understood, thank you for the quick response. I have no strong desire to include a copyright line, so we can just leave it how it is.

tarcieri (approved):

rozbb:
Quick note: the build warnings I was getting were because`field.rs` had unused `pub use` statements, but the module itself is only `pub(crate)`. They're vestigial, it seems. Now the module is all `pub(crate)`

PR #592, opened 2023-10-31
ed25519-dalek: hide secret in `SigningKey`'s `Debug` impl

### LINKED ISSUE #591: `Debug` implementation of `SigningKey` will print secret key material
Since the `Debug` implementation of `SigningKey` is derived, the `secret_key` is being printed and because `SecretKey` is just a type alias for `[u8; 32]`, it uses the std debug implementation for `[T; N]`:

https://github.com/dalek-cryptography/curve25519-dalek/blob/598695c4007d7ee3f48760668c47cf89b9aefb67/ed25519-dalek/src/signing.rs#L61-L64

I think in order to protect users from leaking their secret key, this should be omitted in the Debug output.

tarcieri:
Uses `finish_non_exhaustive` in lieu of printing the `secret_key` component of a `SigningKey`, only showing the corresponding `verifying_key` field which can be used to identify the public key.

Closes #591

### DIFF
--- a/ed25519-dalek/src/signing.rs
+++ b/ed25519-dalek/src/signing.rs
@@ -9,6 +9,8 @@
 
 //! ed25519 signing keys.
 
+use core::fmt::Debug;
+
 #[cfg(feature = "pkcs8")]
 use ed25519::pkcs8;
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Oh joy, we're getting hit by an unrelated nightly regression that misidentifies unused imports.

Let me open a separate PR for that.

robjtede (approved):

tarcieri:
Rebased. Hopefully that fixes the build.

rozbb (approved):

PR #593, opened 2023-10-31
CI: fix minimal-versions resolution

tarcieri:
To avoid nightly regressions breaking the build, the CI configuration has been updated to *only* use nightly for resolving Cargo.lock by using `cargo update -Z minimal-versions`.

Previously, it was running `cargo check` which would attempt to compile all of the dependencies and the code, which is why the diagnostic bug was triggered. By avoiding any kind of code compilation using nightly we can avoid such regressions in the future.

Additionally, the clippy job has been changed to run on the latest stable release (1.73.0) rather than nightly, which will prevent future clippy lints from breaking the build. Instead, they can be addressed when clippy is updated.

### DIFF
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -85,7 +85,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: dtolnay/rust-toolchain@nightly
+    - uses: dtolnay/rust-toolchain@1.73.0
       with:
         components: clippy
     - run: cargo clippy --target x86_64-unknown-linux-gnu --all-features
--- a/.github/workflows/curve25519-dalek.yml
+++ b/.github/workflows/curve25519-dalek.yml
@@ -134,10 +134,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    # First run `cargo +nightly -Z minimal-verisons check` in order to get a
-    # Cargo.lock with the oldest possible deps
+    # Re-resolve Cargo.lock with minimal versions
     - uses: dtolnay/rust-toolchain@nightly
-    - run: cargo -Z minimal-versions check --no-default-features --features serde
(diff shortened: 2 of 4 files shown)


rozbb:
Unused import stuff is fixed in https://github.com/dalek-cryptography/curve25519-dalek/pull/590

tarcieri:
It is? How?

Regardless, this approach should be more futureproof in regard to any other nightly regressions.

.github/workflows/workspace.yml
```diff
@@ -85,7 +87,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: dtolnay/rust-toolchain@nightly
+    - uses: dtolnay/rust-toolchain@1.73.0
```
tarcieri, line 90: We should definitely pin clippy rather than running nightly clippy

rozbb:
Some of the contents of `field.rs` was `pub use`, but the module itself is only `pub(crate)`. I suppose the new warning could be a bug, but it still points out correctly that there was a lot of superfluous pub use.

I agree about the CI being more robust. My only concern is that this adds another chore. Is there a buildabot type automation for this?

tarcieri:
@rozbb if you merge #590 I can unpin nightly

rozbb:
Merged

tarcieri:
Updated with the `nightly` pin removed, although notably if we don't pin nightly it's a source of nondeterminism in the build.

I still think the other changes simplify/stabilize the build if we don't do that, but personally I find it frustrating whenever the build breaks due to things unrelated to my changes.

rozbb:
I agree. If we had a more streamlined way of addressing it separately from our PRs then I'd go for it. Only reason I think we should have nightly in the CI is bc it forces us to make changes for what's coming down the pike into stable.

If you still think it's more reasonable to pin, though, then I'm happy to do it

tarcieri:
Let's cross that bridge when we get there (again), I guess.

At least these changes will limit the scope of the failures to the nightly test.

PR #596, opened 2023-11-01
add hash_to_curve method for Edwards curves

mikelodder7:
Adds hash_to_curve for Edwards points according to [RFC9390](https://datatracker.ietf.org/doc/rfc9380/)


burdges:
RFC9390 requires an XoF, and gives sha2 xmd as a named cipher suite, but RFC9390 permits choosing another XoF like shake128 instead.  I guess maybe the traits expose this option?

mikelodder7:
Yes they do. If you look ExpandMsg supports ExpandMsgXmd or ExpandMsgXof

burdges:
Alright, so already possible to choose shake128 with this PR?  nice  :)

mikelodder7:
Yes any XOF, Shake128, Shake256, etc.

dignifiedquire:
Fyi it is RFC 9380 not 9390 :)

Issue #565, opened 2023-08-20
ed25519-dalek: SigningKey.to_scalar() docs are incorrect


DavidBuchanan314:
https://github.com/dalek-cryptography/curve25519-dalek/blob/c66973c823b45ef99cd9ed3c6cc58cca9805d6d9/ed25519-dalek/src/signing.rs#L477-L494

The docs say that it's a) a byte representation b) unreduced - but iiuc, the Scalar type is neither. I assume these docs used to be correct, but became stale during refactoring.

Further, I don't see an obvious way to use the resulting value to instantiate a `x25519_dalek::StaticSecret` - from my PoV it would be more useful if `to_scalar` really did return the unreduced raw bytes, but the name would be a bit misleading. (maybe making that use-case difficult is intentional, if so, fair enough I guess)

rozbb:
Indeed, this test fails
```rust
#[test]
fn privkey_scalar_yields_x25519_pubkey() {
    let signing_privkey = SigningKey::generate(&mut rand::thread_rng());
    let signing_pubkey = signing_privkey.verifying_key().to_montgomery();
    let signing_scalar = signing_privkey.to_scalar();

    let x_privkey = x25519_dalek::StaticSecret::from(signing_scalar.to_bytes());
    let x_pubkey = x25519_dalek::PublicKey::from(&x_privkey);

    assert_eq!(signing_pubkey.to_bytes(), x_pubkey.to_bytes());
}
```

rozbb:
The reason [these](https://github.com/dalek-cryptography/curve25519-dalek/blob/ac51ef6ecf90ae21f0880d22fe875b3a7eaee884/ed25519-dalek/tests/x25519.rs) tests pass is because they're computing the key shares directly as `signing_key.verifying_key().to_montgomery()`, and using `signing_key.to_scalar()` as the scalars.

I think an appropriate solution to this is to rename the method in question to `SigningKey::to_scalar_bytes()` and return the real unreduced scalar bytes. This partially reverts the change in  https://github.com/dalek-cryptography/curve25519-dalek/pull/545 .

tarcieri:
@rozbb we've been over that before. @dignifiedquire hit something very similar to it, which is why we added the `Scalar` based APIs to avoid transcoding Scalar -> bytes -> Scalar.

https://github.com/RustCrypto/nacl-compat/pull/137/files

[...]

PR #599, opened 2023-11-14
ed: Change `SigningKey::to_scalar` back to `SigningKey::to_scalar_bytes`

rozbb:
As mentioned in the conversation in https://github.com/dalek-cryptography/curve25519-dalek/issues/565, it is currently not clear how you would use a clamped and reduced scalar from `SigningKey::to_scalar()` to build an x25519 keypair. In fact, the test code in that thread shows that `x25519_dalek::PublicKey::from(StaticScalar::from(sk.to_scalar().to_bytes()))` does not match the ed25519 public key. This is because the second public key computation does clamping again, so it's clamp -> reduce -> clamp, which is not a no-op.

This PR replaces `SigningKey::to_scalar()` with `SigningKey::to_scalar_bytes()`, which returns the unreduced scalar bytes determined by the ed25519 secret key. This is actually how things used to be before https://github.com/dalek-cryptography/curve25519-dalek/pull/545, and what the docs (currently erroneously) reflect.

I've also added a regression test to the ed25519 tests.

### DIFF
--- a/ed25519-dalek/CHANGELOG.md
+++ b/ed25519-dalek/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Entries are listed in reverse chronological order per undeprecated major series.
 
+# Unreleased
+
+* Add `SigningKey::to_scalar_bytes` for getting the unclamped scalar from signing key
+
 # 2.x series
 
 ##  2.0.0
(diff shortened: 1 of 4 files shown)


tarcieri:
This is a breaking change that would break running code, namely @dignifiedquire is using these conversions to instantiate a `crypto_box` using a `Scalar` originally computed as an Ed25519 key, via these APIs:

https://github.com/dalek-cryptography/ed25519-dalek/pull/296

That's all working just fine and the existing API is useful.

If you really want to go through with this, you either need to deprecate `to_scalar` instead of just outright deleting it, or bump ed25519-dalek's major version to `3`.

tarcieri:
To be clear: I'm fine with adding `to_scalar_bytes`, but I think it should be a purely additive change.

rozbb:
Ok agreed. I think simply adding it would be fine. And updating the docs on `to_scalar`

tarcieri (approved):

PR #602, opened 2023-11-16
{Signing,Verifying}KeyVisitor: visit_borrowed_bytes -> visit_bytes

awesterb:
This is a slight improvement to the deserialisation code for `SigningKey` and `VerifyingKey`.

The default implementations of `visit_borrowed_bytes` and `visit_byte_buf` [forward to](https://docs.rs/serde/1.0.192/serde/de/trait.Visitor.html#method.visit_borrowed_bytes) `visit_bytes`, but only `visit_borrowed_bytes` was implemented for the `{Signing,Verifying}KeyVisitor`.  This means that `{Signing,Verifying}Key`s can only be deserialised from borrowed byte arrays, but not from [owned or transient byte arrays](https://serde.rs/data-model.html).  Example:

```rust
let bytes = base16ct::lower::decode_vec(
    "66b1419fae979516fb3807dda1b05026b2570a7ab2190254e524af4f0934ddd2",
)
.unwrap();
    
let d = serde::de::value::BorrowedBytesDeserializer::<serde::de::value::Error>::new(&bytes);
ed25519_dalek::VerifyingKey::deserialize(d).unwrap(); // deserialising from a borrowed byte array works
    
let d = serde::de::value::BytesDeserializer::<serde::de::value::Error>::new(&bytes);
ed25519_dalek::VerifyingKey::deserialize(d).unwrap_err(); // but not from a transient byte array
```

Implementing `visit_bytes` instead of `visit_borrowed_bytes` enables deserialisation from all three byte array types.

### DIFF
--- a/ed25519-dalek/src/signing.rs
+++ b/ed25519-dalek/src/signing.rs
@@ -746,10 +746,7 @@ impl<'d> Deserialize<'d> for SigningKey {
                 write!(formatter, concat!("An ed25519 signing (private) key"))
             }
 
-            fn visit_borrowed_bytes<E: serde::de::Error>(
-                self,
-                bytes: &'de [u8],
-            ) -> Result<Self::Value, E> {
+            fn visit_bytes<E: serde::de::Error>(self, bytes: &[u8]) -> Result<Self::Value, E> {
                 SigningKey::try_from(bytes).map_err(E::custom)
             }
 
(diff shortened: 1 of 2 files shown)


rozbb:
Thanks! This makes sense. The only reason we'd implement `visit_borrowed_bytes` separately is if we had a zero-copy deserialization routine for these keys. But all of them copy, so there's no special case here.

robjtede:
Worth mentioning that this same issue was filed over on the old repo (https://github.com/dalek-cryptography/curve25519-dalek/pull/602) and this fix hasn't been released yet.

PR #611, opened 2024-01-01
Update dependencies

qdrs:
Zeroize, Subtle and Serde updates.


robjtede:
Changes to a library crate's manifest like this are unnecessary. Minimum versions for the crate's needs are specified and raising them (within a semver compatible range) should only be done if the crate needs a newly introduced feature.

tarcieri:
This also includes a number of irrelevant formatting changes, and things like changing serde's version requirement from `1.0` to `^1.0` even though the `^` is implicit

Issue #533, opened 2023-06-18
Elligator2 inverse mapping


SnowyCoder:
Hello,
I've been developing a Proof of Concept protocol indistinguishable from random data, as described in [Noise - Handshake Indistinguishability](https://noiseprotocol.org/noise.html#handshake-indistinguishability).
To make this a hash-to-point, inverse elligator2 mapping is required.

I've seen that a pull-request implementing it is already open #357 but there is no discussion regarding if it should be merged or closed. Is there any plan regarding its merging, or do you plan to never merge it to avoid code complexity?

Workarounds:
- Using #389 to provide an extension crate, (`curve25519-dalek-elligator`?) that provides the same features but using an external crate (to avoid code complexity issues)

Additional Pointers:
- If you plan to merge #357 I would suggest adding a tweak/mask field as in the [Monocypher](https://monocypher.org/manual/elligator) implementation and add tests with known implementations (from Monocypher or [elligator.org](https://elligator.org/vectors/)
- There is another independent Rust implementation (with test vectors and tweak field) that I found, but it is completely separate from `curve25519-dalek`, and thus it hinders compatibility with other crates: https://github.com/codahale/qdsa

If you need a hand I'm open to helping, but I'm just an interested Computer Science student with little background in the math behind cryptography.

jrose-signal:
For further comparison, [Signal has an Elligator variant used for Ristretto points](https://github.com/signalapp/curve25519-dalek/blob/lizard2/src/lizard/lizard_ristretto.rs) rather than the whole curve. We've thought about trying to upstream it in the past but haven't really gotten around to it because it's a clunky API—either it exposes the limitation that you can't encode a full 256 bits in a RistrettoPoint (or a MontgomeryPoint, for that matter), or it has to be wrapped in something like the "lizard" method in that file, where you use a good chunk of the point-space as a hash.

jrose-signal:
I'll bump this issue from Signal's perspective: would the crate maintainers be interested in a pair of raw Elligator mapping APIs

- one from `[u8; 32]` to RistrettoPoint
- one from RistrettoPoint to eight candidates of `[u8; 32]`

? Even given that the latter can only produce 253 bits of value.

rozbb:
[...]

PR #612, opened 2024-01-05
Elligator2 Forward and Reverse Mappings

jmwample:
Hello,

This is an implementation of the Elligator2 forward and reverse mappings --- points to representative values,  as well as representative values to points. The specific goal is to make Elligator mapping for x25519 handshakes possible in pure rust.

This implementation is gated behind a feature flag `"elligator2"`  in the `curve25519_dalek` crate for point transforms. I have tested this against the test vectors from the `Kleshni/Elligator-2` C implementation as well as (a fork of) the `agl/ed25519/extra25519` golang implementation. These can be seen in the test cases for this PR.

I also added a feature flag `"elligator2"` to the `x25519_dalek` crate that exposes a `PublicRepresentative` type that mirrors the `PublicKey` type in order to exchange only Elligator Representatives when performing a diffie-hellman handshake. 

This implementation of the Elligator2 transforms (to the best of my ability):
- runs in constant time
- is equivalent to `agl/ed25519/extra25519`
- is equivalent to `Kleshni/Elligator-2`
- supports existing backends (I have tested with u64 and u32, I am not sure how to test others.)

I have seen a couple issues and PRs  dealing with elligator2 mappings and this library, so I apologize if this is muddying the water . This is a feature that I really needed so I have implemented the functionality based on several existing implementations in other languages. 

I am also not an expert in crypto implementations and would greatly appreciate any feedback. I hope this is helpful and potentially solid enough to include in the library. :+1: 

Related to:
- #357 
- #533

### DIFF
--- a/curve25519-dalek/src/elligator2.rs
+++ b/curve25519-dalek/src/elligator2.rs
@@ -1,0 +1,583 @@
+// -*- mode: rust; -*-
+
+//! Functions mapping curve points to representative values
+//! (indistinguishable from random), and back again.
+use crate::constants::{MONTGOMERY_A, MONTGOMERY_A_NEG};
+use crate::field::FieldElement;
+use crate::montgomery::MontgomeryPoint;
+use crate::EdwardsPoint;
+
+use subtle::{
+    Choice, ConditionallyNegatable, ConditionallySelectable, ConstantTimeEq, ConstantTimeGreater,
+    CtOption,
+};
+
+/// (p - 1) / 2 = 2^254 - 10
+pub(crate) const DIVIDE_MINUS_P_1_2_BYTES: [u8; 32] = [
+    0xf6, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+    0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x3f,
+];
+
+/// Gets the public representative for a key pair using the private key
+pub fn representative_from_privkey(privkey: &[u8; 32]) -> Option<[u8; 32]> {
+    let pubkey = EdwardsPoint::mul_base_clamped(privkey.clone()).to_montgomery();
+    let v_in_sqrt = v_in_sqrt(privkey);
+    point_to_representative(&pubkey, v_in_sqrt.into()).into()
+}
+
+/// This function is used to map a curve point (i.e. an x25519 public key)
+/// to a point that is effectively indistinguishable from random noise.
+///
+/// This operation may fail because not all curve points are capable of being
+/// hidden. On failure, this function will return None.
+///
+/// This implementation is adapted from both:
+/// - [kleshni C implementation](https://github.com/Kleshni/Elligator-2)
+/// - [agl/ed25519 golang forks](https://gitlab.com/yawning/edwards25519-extra)
+///
+///
+/// Note: the output field elements of this function are uniformly
+/// distributed among the nonnegative field elements, but only if the
+/// input points are also uniformly distributed among all points of
+/// the curve. In particular, if the inputs are only selected from
+/// members of the prime order group, then the outputs are
+/// distinguishable from random.
+///
+/// # Inputs
+///
+/// * `point`: the \\(u\\)-coordinate of a point on the curve. Not all
+/// points map to field elements.
+///
+/// * `v_in_sqrt`: true if the \\(v\\)-coordinate of the point is negative.
+///
+/// # Returns
+///
+/// Either `None`, if the point couldn't be mapped, or `Some(fe)` such
+/// that `fe` is nonnegative and `map_to_point(&fe) == point`.
+///
+/// [elligator paper](https://elligator.cr.yp.to/elligator-20130828.pdf)
+/// [elligator site](https://elligator.org/)
+///
+pub fn point_to_representative(
+    point: &MontgomeryPoint,
+    v_in_sqrt: bool,
+) -> CtOption<[u8; 32]> {
+    let divide_minus_p_1_2 = FieldElement::from_bytes(&DIVIDE_MINUS_P_1_2_BYTES);
+
+    // a := point
+    let a = &FieldElement::from_bytes(&point.0);
+    let mut a_neg = a.clone();
+    a_neg.negate();
+
+    let is_encodable = is_encodable(a);
+
+    // Calculate r1 = sqrt(-a/(2*(a+A)))
+    let (_r1_sqrt, r1) = FieldElement::sqrt_ratio_i(
+        &a_neg,
+        &(&(a + &MONTGOMERY_A) * &(&FieldElement::ONE + &FieldElement::ONE)),
+    );
+
+    // Calculate  r0 = sqrt(-(a+A)/(2a))
+    let (_r0_sqrt, r0) = FieldElement::sqrt_ratio_i(
+        &(&a_neg - &MONTGOMERY_A),
+        &(a * &(&FieldElement::ONE + &FieldElement::ONE)),
+    );
+
+    // if v_in_sqrt root=r0 otherwise r1
+    let mut b = FieldElement::conditional_select(&r1, &r0, Choice::from(v_in_sqrt as u8));
+
+    // If root > (p - 1) / 2, root := -root
+    let negate = divide_minus_p_1_2.ct_gt(&b);
+    FieldElement::conditional_negate(&mut b, negate);
+
+    CtOption::new(b.as_bytes(), is_encodable)
+}
+
+/// Determines whether a point is encodable as a representative. Approximately
+/// 50% of points are not encodable.
+#[inline]
+fn is_encodable(u: &FieldElement) -> Choice {
+    let b0 = u + &MONTGOMERY_A;
+    let b1 = &(&(&b0.square().square() * &b0.square()) * &b0) * u; // b1 = u * (u + A)^7
+    let c = b1.pow_p58();
+
+    let b2 = &(&b0.square().square().square() * &b0.square().square()) * &b0.square(); // (u + A)^14
+    let mut chi = &(&c.square().square() * &u.square()) * &b2; // chi = -c^4 * u^2 * (u + A)^14
+    chi.negate();
+
+    let chi_bytes = chi.as_bytes();
+
+    // chi[1] is either 0 or 0xff
+    chi_bytes[1].ct_eq(&0_u8)
+}
+
+///  `high_y` - Montgomery points can have have two different values for a
+///  single X coordinate (based on sign). The Kleshni implementation determines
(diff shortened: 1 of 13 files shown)


rozbb:
Thank you! It will take a bit for me to take a look at this (trying to make some paper deadlines).

Re the IETF standard, hasn't that officially been standardized now? RFC 9380. Would be good to have test vectors from there.

jmwample:
I will get those added. I haven't spent much time looking at that RFC yet, so I can try to work in the interface there as well (and get the rest of the tests passing).

jmwample:
I have added the [RFC9380](https://datatracker.ietf.org/doc/rfc9380/) test vectors for the elligator2 implementation, and an interface that should work if someone wanted to use this to implement the `h2c` interface for `curve25519` and/or `edwards25519`. Not sure if the full `h2c` implementation belongs in this crate or a different crate with a more general / uniform interface.

The CI checks should also be passing now.

randombit:
Any chance of this landing soon?

jmwample:
I am working on adding some final tests to ensure that the bits of the ellgator2 representatives appear as uniform random.

TLDR: The `sqrt_ratio_i` function seems to be canonical so this library shouldn't suffer from the described computational distinguisher. However, deriving an `elligator2` representative only gives 254 bits of random to begin with, so the high order bits need handled in some way to prevent a trivial distinguisher.

The specific issue that this is testing for can be described as:

```txt
An instantiation of Elligator is parameterized by what might be called
a “canonical” square root function, one with the property that
`√a2 = √(−a)2` for all field elements `a`. That is, we designate just
over half the field elements as “non-negative,” and the image of the
square root function consists of exactly those elements. A convenient
definition of “non-negative” for Curve25519, suggested by its authors,
is the lower half of the field, the elements `{0, 1, …, (q − 1)+/+2}`.
When there are two options for a square root, take the smaller of the two.
```

Any Elligator implementation that does not do this canonicalization of the final square root, and instead it maps a given input systematically to either its negative or non-negative root is vulnerable to the following computational distinguisher.

```txt
[An adversary could] observe a representative, interpret it as a field
element, square it, then take the square root using the same
non-canonical square root algorithm. With representatives produced by
an affected version of [the elligator2 implementation], the output of
the square-then-root operation would always match the input. With
random strings, the output would match only half the time.
```

The solution from the `agl/ed25519` is to randomize the high two bits when getting the representative, and clear the those same high order bits when performing a `map_to_curve` operation. One challenge is determining how this aligns with the test vectors from other implementations (`kleshni` and `signal` specifically, the `rfc9380` tests seem to be handled properly). 

For a more in-depth explanation see:
- https://github.com/agl/ed25519/issues/27
- https://www.bamsoftware.com/papers/fep-flaws/

This should not impact the general interface of the PR, and I am hoping to have the changes finished within the week.

jmwample:
I have added another refactor to the elligator2 implementation motivated by feedback based on issues encountered with encoded key distinguishability in obfs4. The changes required to fix the distinguisher resulted in two versions of the elligator2 algorithm which are not interchangeable.

More information on the issue can be found [here](https://www.reddit.com/r/crypto/comments/fd9t3m/elligator_with_x25519_is_broken_what_workaround/), the solution added to the `Randomized` variant is described in [Method 1: add a random low order point](https://elligator.org/key-exchange). The `RFC9380` variant is the default and follows the RFC including test vectors.

A test variant exists for legacy implementations that do not use least-square-root value of the representative (i.e. kleshni & signal), but it is not exposed by default.

---

For now I have published my fork as its own crate (see [curve25519-elligator2](https://docs.rs/curve25519-elligator2/latest/curve25519_elligator2/)), but my intention is to hopefully get this merged here and eventually yank the forked crate.

rozbb:
Hi, thank you for this! I played around with this and had some notes:

1. It seems like I can't get the Elligator2 tests to fail, even when they definitely should. For example, in `src/field.rs`, I replaced the return value of `FieldElement::ct_gt` with the constant `Choice::from(0u8)`. Running `cargo test --features "alloc,elligator2"` did not produce any errors. Is a KAT missing?

2. The reason I was playing with the above is because I was wondering about the correctness of the `gt` function in the fiat backend in the diff. My understanding is that this does a libgmp-style bigint subtraction without the subtraction. But the difference between a bigint and a [`fiat_25519_tight_field_element`](https://docs.rs/fiat-crypto/latest/fiat_crypto/curve25519_64/struct.fiat_25519_tight_field_element.html) is that the latter has nontrivial equivalences. I'm worried `gt` might consider an unreduced `fiat_25519_tight_field_element` of value, say 2²⁵⁵ - 1 (equivalent to 18) as greater than 30, for example. I was trying to test this but ran into point (1) above.

I'll be reviewing the rest of the PR, but I think these should be addressed before merging. Thanks again!

jmwample:
> 1. It seems like I can't get the Elligator2 tests to fail, even when they definitely should. Is a KAT missing?

> 2. I was wondering about the correctness of the gt function in the fiat backend in the diff. [fiat_25519_tight_field_element] has nontrivial equivalences. I'm worried `gt` might consider an unreduced fiat_25519_tight_field_element of value, say 2²⁵⁵ - 1 (equivalent to 18) as greater than 30, for example.

You are definitely correct about the implementation of the gt function. I was relying on the idea that other implementations were using a reliable compare, but it turns out they fall victim to the exact issue that you describe. Really what they are trying to check is whether the value is negative, but they do so with by subtracting $(p-1)/2$ and
checking for overflow. However a direct negative check doesn't work for the `Legacy` implementations because the signaland kleshni implementations [use the subtract and check for overflow](https://github.com/Kleshni/Elligator-2/blob/c767c43c8a93601bec65983af28e5e3ef96bee44/main.c#L193) which has a different result from a proper greater than comparison (or `is_negative()`) on the canonical representation of the `FieldElement`. I am trying to decide what the best way to handle the separate cases is. I have a proper `ct_gt()` implementation, but I am not sure it is even useful as RFC9380 and Randomized can use `is_negative()` and `Legacy` can't use a proper greater-than.

> I replaced the return value of `FieldElement::ct_gt` with the constant `Choice::from(0u8)`. Running `cargo test --features "alloc,elligator2"` did not produce any errors.

For the RFC9380 (and Randomized) the effect is less obvious as it only takes effect when $\sqrt{\frac{-(a+A)}{2a}} \ne \sqrt{\frac{-a}{2*(a+A)}}$ [(code ref)](https://github.com/jmwample/curve25519-dalek/blob/15ad0b3f66100d7ca63bb84dbbf4335ddbba8014/curve25519-dalek/src/elligator2.rs#L583) which for some reason is not the case for the test vectors from RFC8380 or my randomly selected test vectors. However a simple test where they are different is the identity value (which I have added in my work branch looking into this).

---

As an aside, is the implementation of `is_negative()` flawed in a similar way? In the elligator2 implementations
the check for negative values is to determine if the FieldElement is less than $(p-1)/2 = 2^{254}-10$ while the
`FieldElement::is_negative()` function only checks if the low bit is set meaning that values between $2^{254}-9$ and $2^{254}-1$ would be considered positive?

rozbb:
I'm gonna spend some time this week going thru this. I really appreciate this PR @jmwample. It's been tough co-maintaining this while having a full time job. I don't know a lot about Elligator2, and especially the distinguishing attacks mentioned here, so I will have to read a lot here, and probably ask a bunch of questions. Thank u in advance!

carloskiki:
Thanks @rozbb!

jmwample:
Great to hear! I rebased onto the latest changes in `main`. I am happy to help with any clarifications and work in any interface changes.

curve25519-dalek/src/elligator2.rs
```diff
@@ -0,0 +1,786 @@
...
+/// element, square it, then take the square root using the same
+/// non-canonical square root algorithm. With representatives produced by
+/// an affected version of [the elligator2 implementation], the output of
+/// the square-then-root operation would always match the input. With
+/// random strings, the output would match only half the time.
+/// ```
+///
+/// For a more in-depth explanation see:
+/// <https://github.com/agl/ed25519/issues/27>
+/// <https://www.bamsoftware.com/papers/fep-flaws/>
```
rozbb, line 404: I don't understand the distinction. The two things you linked above do not seem to point out any flaw in the RFC, rather implementation errors in some libraries. Is there something inherent in the RFC that requires modification? The impl above already appears to do randomization.

It appears that the actual issue is that the above RFC 9380 mapping always clears the cofactor, which is a leaky thing to do. If that's the issue, why not change the API to do the real mappings point -> field rather than scalar -> point -> field?

curve25519-dalek/src/elligator2.rs
```diff
@@ -0,0 +1,786 @@
...
+    /// Takes a public representative and returns an Edwards curve field element
+    /// if one exists.
+    fn from_representative(representative: &[u8; 32]) -> CtOption<EdwardsPoint>;
+
+    /// Takes a private key value and gets a byte representation of the public representative.
+    ///
+    /// The `tweak` parameter is used to adjust the computed representative making
+    /// it computationally indistinguishable from uniform random. If this property
+    /// is not required then the provided tweak value does not matter.
+    fn to_representative(point: &[u8; 32], tweak: u8) -> CtOption<[u8; 32]>;
```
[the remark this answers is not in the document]
rozbb: Also, why is this the API and not the actual point -> repr mapping?

curve25519-dalek/src/elligator2.rs
```diff
@@ -0,0 +1,786 @@
...
+
+    // complete Y
+    let y2 = FieldElement::conditional_select(&gx2, &gx1, eps_is_sq);
+    let (_, mut y) = FieldElement::sqrt_ratio_i(&y2, &one);
+    y.conditional_negate(eps_is_sq ^ y.is_negative());
+
+    (&x * &d_1, d_1, y, one)
+}
+
+pub(crate) fn map_fe_to_montgomery(r: &FieldElement) -> (FieldElement, FieldElement) {
```
rozbb, line 720: Comment needed

curve25519-dalek/src/elligator2.rs
```diff
@@ -0,0 +1,786 @@
...
+    let (xmn, xmd, y, _) = map_to_curve_parts(r);
+    (&xmn * &(xmd.invert()), y)
+}
+
+pub(crate) fn map_fe_to_edwards(r: &FieldElement) -> (FieldElement, FieldElement) {
```
rozbb, line 725: Should mention this is `map_to_curve_elligator2_edwards25519` from RFC 9380 section G.2.2

rozbb (commented):
Ok this was a lot but I left a first round of comments. The high level comment is that I'm not sure where any of this math comes from. It seems like a mix of the RFC, the paper, Signal's impl, and agl's impl. If that's actually the case, then this will need a lot of references and explanations.

One thing I realized is the inverse maps are not standardized. So the RFC cannot be followed, and in fact it looks like it is not followed (eg using a range check versus `x mod 2` for the is-nonneg function). These choices should probably be documented somewhere public. We cannot interoperate with other Elligator implementations if we don't have a reference.

Sorry for the wall of text. This was a bunch of reading and I learned a lot

JobDoesburg:
Is there any chance on this being merged in the foreseeable future? That would be great!

Issue #618, opened 2024-02-06
Build fails on nightly-2024-02-05


leighmcculloch:
Builds using curve25519-dalek started to fail on `nightly-2024-02-05`. These builds passed on `nightly-2024-02-03`.

Here's an extract of the build error, from a GitHub Actions CI run where the full build is captured [here](https://github.com/stellar/rs-soroban-sdk/actions/runs/7793122785/job/21252345250#step:5:335).

```
error[E0658]: use of unstable library feature 'stdarch_x86_avx512'
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/curve25519-dalek-4.1.1/src/backend/vector/ifma/field.rs:26:5
   |
26 |     _mm256_madd52lo_epu64(z.into(), x.into(), y.into()).into()
   |     ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #111137 <https://github.com/rust-lang/rust/issues/111137> for more information
   = help: add `#![feature(stdarch_x86_avx512)]` to the crate attributes to enable
   = note: this compiler was built on 2024-02-05; consider upgrading it if it is out of date

error[E0658]: use of unstable library feature 'stdarch_x86_avx512'
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/curve25519-dalek-4.1.1/src/backend/vector/ifma/field.rs:25:9
   |
25 |     use core::arch::x86_64::_mm256_madd52lo_epu64;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #111137 <https://github.com/rust-lang/rust/issues/111137> for more information
   = help: add `#![feature(stdarch_x86_avx512)]` to the crate attributes to enable
   = note: this compiler was built on 2024-02-05; consider upgrading it if it is out of date

error[E0658]: use of unstable library feature 'stdarch_x86_avx512'
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/curve25519-dalek-4.1.1/src/backend/vector/ifma/field.rs:34:5
   |
34 |     _mm256_madd52hi_epu64(z.into(), x.into(), y.into()).into()
   |     ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #111137 <https://github.com/rust-lang/rust/issues/111137> for more information
[...]

PR #457, opened 2022-12-08
NEON backend for aarch64

rubdos:
This is joint work with @Tarinn. In fact, @Tarinn did most of the work :-)

I'm working on benchmarks across several devices now (including an unpublished very hacky armv7 version), will update here when they become available. We're seeing speedups of 20-30% in relevant benchmarks. This is the same code as in https://github.com/zkcrypto/curve25519-dalek-ng/pull/19.

I'll be cleaning up the commits in the next few minutes (trailing white space fixes, mostly), and I'll run a comparison benchmark for 343be3a, because that was only introduced for future ARMv7 support, and has not been tested under load.

PR #619, opened 2024-02-06
Fix nightly build

jimmygchen:
Addresses #618.

This PR fixes the failing build on the latest Rust nightly compiler `rustc 1.78.0-nightly (f067fd608 2024-02-05)`.

The `stdsimd` feature has been split into separate features in the latest Rust nightly compiler and its usage needs to be updated to use `#![feature(stdarch_x86_avx512)]` instead.

See https://github.com/rust-lang/stdarch/pull/1486 and https://github.com/rust-lang/rust/issues/111137.

### DIFF
--- a/curve25519-dalek/src/lib.rs
+++ b/curve25519-dalek/src/lib.rs
@@ -10,7 +10,10 @@
 // - Henry de Valence <hdevalence@hdevalence.ca>
 
 #![no_std]
-#![cfg_attr(all(curve25519_dalek_backend = "simd", nightly), feature(stdsimd))]
+#![cfg_attr(
+    all(curve25519_dalek_backend = "simd", nightly),
+    feature(stdarch_x86_avx512)
+)]
 #![cfg_attr(
     all(curve25519_dalek_backend = "simd", nightly),
     feature(avx512_target_feature)
--- a/curve25519-dalek/src/lib.rs
+++ b/curve25519-dalek/src/lib.rs
@@ -11,7 +11,7 @@
 
 #![no_std]
 #![cfg_attr(
-    all(curve25519_dalek_backend = "simd", nightly),
+    all(curve25519_dalek_backend = "simd", nightly, any(target_arch = "x86", target_arch = "x86_64")),
     feature(stdarch_x86_avx512)
 )]
 #![cfg_attr(
--- a/curve25519-dalek/src/lib.rs
+++ b/curve25519-dalek/src/lib.rs
@@ -11,7 +11,11 @@
 
 #![no_std]
 #![cfg_attr(
-    all(curve25519_dalek_backend = "simd", nightly, any(target_arch = "x86", target_arch = "x86_64")),
+    all(
+        curve25519_dalek_backend = "simd",
+        nightly,
+        any(target_arch = "x86", target_arch = "x86_64")
+    ),
     feature(stdarch_x86_avx512)
 )]
 #![cfg_attr(


rozbb:
Thank you, this looks great! For anyone reading: we have two SIMD implementations for x86_64—AVX2 and AVX-512. The former has already been stabilized, so it's not an issue. The latter is unstable, and was previously gated by a nightly feature `stdsimd`. As @jimmygchen said, this feature has been split into more fine grained features. So, we get errors in #618 like
```
error[E0658]: use of unstable library feature 'stdarch_x86_avx512'
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/curve25519-dalek-4.1.1/src/backend/vector/ifma/field.rs:25:9
   |
25 |     use core::arch::x86_64::_mm256_madd52lo_epu64;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #111137 <https://github.com/rust-lang/rust/issues/111137> for more information
   = help: add `#![feature(stdarch_x86_avx512)]` to the crate attributes to enable
   = note: this compiler was built on 2024-02-05; consider upgrading it if it is out of date
```
because [`_mm256_madd52lo_epu64`](https://doc.rust-lang.org/nightly/core/arch/x86_64/fn._mm256_madd52lo_epu64.html) is behind that new AVX-512 gate.

rozbb:
@tarcieri happy to merge. Shall I also cut a new version (minor or patch?)?

curve25519-dalek/src/lib.rs
```diff
@@ -10,7 +10,10 @@
 // - Henry de Valence <hdevalence@hdevalence.ca>
 
 #![no_std]
-#![cfg_attr(all(curve25519_dalek_backend = "simd", nightly), feature(stdsimd))]
+#![cfg_attr(
+    all(curve25519_dalek_backend = "simd", nightly),
+    feature(stdarch_x86_avx512)
+)]
```
tarcieri, line 16: Perhaps this should consider `target_arch` when setting `stdarch_x86_avx512` in order to be a little more futureproof? Something like:

```suggestion
#![cfg_attr(
    all(curve25519_dalek_backend = "simd", nightly, any(target_arch = "x86", target_arch = "x86_64")),
    feature(stdarch_x86_avx512)
)]
```

rozbb: I believe this is covered in the build file. Lmk if I'm wrong

https://github.com/dalek-cryptography/curve25519-dalek/blob/ba737a379071191158bacfa6d138f6249b12fc09/curve25519-dalek/build.rs#L50-L52

tarcieri: The issue would occur if we added other non-x86 SIMD backends, e.g. #457

rozbb: I see, bc in that case `is_capable_simd` ought to be updated to be `true` when on aarch64. Ok yeah let's add that constraint.

leighmcculloch:
Thanks for the fast fix.

Which version is this targeting?

rozbb:
This will be a patch bump (ie +0.0.1) for curve-, ed-, and x- crates. Sound good @tarcieri ?

tarcieri:
Yep, sounds good

sergiimk:
This change unfortunately breaks the build for older nightlies.

Our company is probably not the only one who uses nightly `rustc` not just for testing, but for access to unstable features, so we don't upgrade `rustc` that often and still running:
```
rustc 1.76.0-nightly (90e321d82 2023-12-02)
```

Don't know if this particular change could've been done in a backwards-compatible way or not, but just wanted to mention that the latest patch release likely both fixed AND broke a lot of builds.

tarcieri:
@sergiimk this is a breaking change in the nightly compiler. It requires a corresponding breaking change in this library to support.

The nightly compiler is definitionally the unstable version of the compiler. Breaking changes can come at any time, the onus is on you to keep what you're building on it stable via meticulous dependency management.

rozbb:
Is there an accepted way to handle the following scenario? You have a crate that relies on nightly. You pull in curve25519-dalek as a repo, but you don't really care what it does. Can you force dalek to behave as it would in stable, while still using nightly for your own crate?

tarcieri:
I think it's unreasonable ask for support for anything but the latest version of the nightly compiler

rozbb:
Right, what I was thinking is something along the lines of a negative feature like `no-nightly` which basically overrides our toolchain autodetect and skips all nightly features.

IIRC we had a `nightly` feature we killed. So not sure if the inverse is any better. Just an idea I was wondering if there's precedence for

tarcieri:
@rozbb disabling the `nightly` cfg attribute would accomplish that. It gets set automatically in `build.rs` here: https://github.com/dalek-cryptography/curve25519-dalek/blob/40cf5aff99ef455acf62ff2d83a340ebaddf77e4/curve25519-dalek/build.rs#L29

rozbb:
Ah nice, that answers it then

PR #620, opened 2024-02-06
Bump digest and sha2 to pre-release versions

baloo:
Upcoming changes in `digest` and `sha2` crates

Depends:
 - https://github.com/RustCrypto/signatures/pull/807


tarcieri:
Would definitely be good to get #619 out first

This should probably also bump the versions of `curve25519-dalek`/`ed25519-dalek` to e.g. 4.2.0-pre/2.2.0-pre respectively per the exemptions to our SemVer policy, to denote any release needs a bumped minor version number

baloo:
I also need to figure out the exact MSRV of both.

Issue #623, opened 2024-02-09
How to check a VerifyingKey point is within the prime order subgroup


randombit:
Given a `VerifyingKey` that I've created from bytes sent to me by another party, how can I check that the point is within the prime order subgroup?

IIUC, I can check if it is contained entirely within the torsion subgroup using `is_weak()`, but to check that it is within the subgroup mod \ell the best I can find is

`pk.to_montgomery().to_edwards(0).unwrap().is_torsion_free()`

which seems quite contorted, not to mention pointlessly expensive.

Am I missing something that would make this easier/cheaper?

tarcieri:
I guess one of the immediate concerns here is there is no `From<VerifyingKey> for EdwardsPoint` impl, or a `to_edwards` method, to get the `EdwardsPoint` out of a `VerifyingKey`, which seems like an oversight.

I believe this would be checked if we implemented NIST's "D.1.3.2. Full Public Key Validation": see https://github.com/dalek-cryptography/curve25519-dalek/issues/380#issuecomment-1487681764. Namely step 3:

> 3. Verify that nQ = (0,1). Output REJECT if verification fails.

randombit:
Opened a patch proposal in #624

randombit:
Thank you for the fast review!

PR #624, opened 2024-02-12
Add VerifyingKey::to_edwards

randombit:
See #623

### DIFF
--- a/ed25519-dalek/src/verifying.rs
+++ b/ed25519-dalek/src/verifying.rs
@@ -505,6 +505,11 @@ impl VerifyingKey {
     pub fn to_montgomery(&self) -> MontgomeryPoint {
         self.point.to_montgomery()
     }
+
+    /// Return this verifying key in Edwards form.
+    pub fn to_edwards(&self) -> EdwardsPoint {
+        self.point
+    }
 }
 
 impl Verifier<ed25519::Signature> for VerifyingKey {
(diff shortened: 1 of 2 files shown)


ed25519-dalek/src/verifying.rs
```diff
@@ -505,6 +505,11 @@ impl VerifyingKey {
     pub fn to_montgomery(&self) -> MontgomeryPoint {
         self.point.to_montgomery()
     }
+
+    /// Return this verifying key in Edwards form.
+    pub fn to_edwards(&self) -> EdwardsPoint {
+        self.point
+    }
```
tarcieri, line 512: Perhaps add a `From` impl too?

randombit: There is already a `impl From<EdwardsPoint> for VerifyingKey` - is your suggestion here to add the cooresponding `impl From<VerifyingKey> for EdwardsPoint`?

tarcieri:
Hmm, an unrelated failure due to new nightly lints for a debugging trait which needs to get annotated with `#[allow(dead_code)]`

ed25519-dalek/src/verifying.rs
randombit: Done

tarcieri:
@randombit #625 is merged. If you rebase it should fix the test failure

rozbb:
You can merge `main` now to fix the CI errors

randombit:
Rebased with #625

rozbb:
Thank you!

PR #625, opened 2024-02-12
curve: mark `ValidityCheck` trait as `allow(dead_code)`

tarcieri:
Recent nightlies have started emitting a dead code lint which is breaking the build: https://github.com/dalek-cryptography/curve25519-dalek/actions/runs/7873600894/job/21481401663?pr=624#step:4:152


rozbb (commented):
Thanks for the quick change. One thought: would it make more sense for this to be `#[cfg(test)]`? It's not accessible outside the crate anyway so it won't bother external users

rozbb:
Ah ok then

Issue #452, opened 2022-12-06
Please release (3.2.2?) with relaxed zeroize dependency


soleinik-figment:
Please release with [this](https://github.com/dalek-cryptography/curve25519-dalek/commit/90292dccf84cdc614c1a5640f2d6cb13bfff9761) commit at the top of main (relax **zeroize** dependency)

### Reason 
would resolve dependency issues across **solana-program**, **cosmrs** and **subxt** (so far, maybe more)

tarcieri:
See #405 for the release tracking issue

juchiast:
Can we just release 3.2.2 instead of waiting for 4.0.0?

tarcieri:
The release of 4.0.0 is imminent

juchiast:
@tarcieri  Do you have a specific time?

tarcieri:
Should hopefully be this week

tarcieri:
It's out: https://crates.io/crates/curve25519-dalek/4.0.0

ilya-bobyr:
We can not upgrade to 4.0.0 immediately as `curve25519-dalek` is an indirect dependency via `ed25519-dalek` version `1.0.1`.
And we can not immediately upgrade from `ed25519-dalek` `1.0.1` to `2.0.0-rc` as it is not a backward compatible change.  We use some of the types provided by `ed25519-dalek` in our public API.

It would be really great if the `zerioze` constrain removal would not be coupled with the whole library upgrade.

I have a private fork of `curve25519-dalek` `3.2.1` that we have to use: https://github.com/dalek-cryptography/curve25519-dalek/compare/3.2.1...ilya-bobyr:curve25519-dalek:3.2.1-unpin-zeroize [1].

Would it make sense to send a PR that cherry picks one of these commits on top of `3.2.1` with the `releases/3.2` target?

* "Relax version constraints for zeroize dependency." https://github.com/dalek-cryptography/curve25519-dalek/commit/841b3a65dec5598e665e62181f64ce1ea1effcba (x25519-2.0.0 / curve25519-4.0.0)
* "Relaxed zeroize dependency" https://github.com/dalek-cryptography/curve25519-dalek/commit/90292dccf84cdc614c1a5640f2d6cb13bfff9761 (x25519-2.0.0 / 4.0.0-pre.3)
[...]

PR #628, opened 2024-02-14
Backport: Relax zeroize version in 3.2

JakkuSakura:
This is to correspond to 
https://github.com/dalek-cryptography/curve25519-dalek/issues/452#issuecomment-1749807071

However, it increased MSRV and may break not locked versions that are not using newer rustc


tarcieri:
3.x and older versions are unmaintained. Please upgrade to the newest version.

PR #640, opened 2024-03-08
ed25519: Add wycheproofs

pinkforest:
- Generate wycheproof driven tests
- Machine generated wycheproof test cases for ed25519-dalek
- If any wycheproof is added this can be seen via diff easily

Problem by dynamically adding static tests is these can be missed silently if the dyn generation fails - google doesn't add / remove them much in any case so :woman_shrugging:

### DIFF
--- a/wycheproofs/generate-wycheproofs/Cargo.toml
+++ b/wycheproofs/generate-wycheproofs/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "generate-wycheproofs"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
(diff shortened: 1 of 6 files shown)


wycheproofs/generate-wycheproofs/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "generate-wycheproofs"
+version = "0.1.0"
+edition = "2021"
```
koba-e964, line 4: Maybe adding `publish = false` may help because the crate looks primarily for testing and not meant to be used by other people.

PR #647, opened 2024-03-27
Updates license field to valid SPDX format

ibeckermayer:
The popular [`dependency-review-action`](https://github.com/actions/dependency-review-action) requires the license field be in valid SPDX format. (We are hitting false positives on this dependency via `dependency-review-action` over at [Teleport](https://github.com/gravitational/teleport)).


ibeckermayer:
Giving this one a bump.

tarcieri (approved):

PR #652, opened 2024-05-08
Mitigate check-cfg until MSRV 1.77

pinkforest:
1.80 and nightly (2024-05-05) has introduced cfg-check spitting out some warnings.

The  problem is if we declare the cfg's as intended - this  will then complain about MSRV despite where we would only print these flags when rustc is >= 1.77 a.k.a this feature hard-checks MSRV and then asks to bump MSRV 1.77 despite build.rs never using the feature as gated in build.rs for below <= 1.77 as checked.

In order to mitigate this we can just temporarily allow the status-quo and when MSRV jumps to 1.77 we can declare them when MSRV bumps to 1.77.

Also need to allow two occassions unnecessary qualification (group::ff) where when only group feature is used when ff is available directrly where as it's always available via group::ff - leading to lint complaining.

### DIFF
--- a/curve25519-dalek-derive/tests/tests.rs
+++ b/curve25519-dalek-derive/tests/tests.rs
@@ -24,10 +24,6 @@ where
 }
 
 #[unsafe_target_feature("sse2")]
-#[cfg(feature = "dummy")]
-fn function_with_cfg() {}
-
-#[unsafe_target_feature("sse2")]
 #[rustfmt::skip]
 fn function_with_rustfmt_skip() {}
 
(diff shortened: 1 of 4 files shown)


pinkforest:
Yes it will result to MSRV check and it has a hard complaint to require bumping to 1.77 regardless if this is gated only to >= 1.77

tarcieri:
How does that actually manifest? Do older rusts complain about it?

pinkforest:
No the new versions complain that is the problem - where we actually would declare them and where it takes effect.

tarcieri:
Aah, I see now:

```
error: the `cargo::` syntax for build script output instructions was added in Rust 1.77.0, but the minimum supported Rust version of `curve25519-dalek v4.1.2 (/Users/tony/src/curve25519-dalek/curve25519-dalek)` is 1.60.0.
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
```

tarcieri (approved):

Issue #380, opened 2022-01-10
EdwardsPoint::decompress accepts y higher than field modulo


paulmillr:
EdwardsPoint::decompress accepts y higher than field modulo

Which is invalid as per RFC8032 5.1.3.

```
   Decoding a point, given as a 32-octet string, is a little more
   complicated.

   1.  First, interpret the string as an integer in little-endian
       representation.  Bit 255 of this number is the least significant
       bit of the x-coordinate and denote this value x_0.  The
       y-coordinate is recovered simply by clearing this bit.  If the
       resulting value is >= p, decoding fails.
```

AnomalRoil:
RFC8032 is about EdDSA, while this repo is really "just" about Curve25519, so I'd argue that this is not an issue for curve25519, but rather for the ed25519 repo.

That being said, it seems that currently https://github.com/dalek-cryptography/ed25519-dalek is directly using the `decompress` function without doing that check [because](https://github.com/dalek-cryptography/ed25519-dalek/blob/925eb9ea56192053c9eb93b9d30d1b9419eee128/src/public.rs#L96-L98):

``` 
    /// The caller is responsible for ensuring that the bytes passed into this
    /// method actually represent a `curve25519_dalek::curve::CompressedEdwardsY`
    /// and that said compressed point is actually a point on the curve.
 ```

Also the struct `EdwardsPoint` is guaranteed to hold a valid point on the curve:
https://github.com/dalek-cryptography/curve25519-dalek/blob/967d8b6c0e67100401ad66125b7399ccf509ae22/src/edwards.rs#L199-L201

E.g. what if I consider the CompressedEdwardsY value `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f`, once decompressed it will give me a point that is equivalent to the point `0x1200000000000000000000000000000000000000000000000000000000000000`:

```
PublicKey(CompressedEdwardsY: [255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 127]), 
EdwardsPoint{
	X: FieldElement51([1765291371327528, 828421974544742, 17095614122396, 1955301457663811, 970631510557499]),
	Y: FieldElement51([2251799813685247, 2251799813685247, 2251799813685247, 2251799813685247, 2251799813685247]),
[...]

PR #653, opened 2024-05-10
ed: Removed warning in docs for `VerifyingKey::from_bytes`

rozbb:
There was a big warning in `VerifyingKey::from_bytes` that said the user is responsible for ensuring that the provided bytes represent a verifying key. I think this was confusing, since the function returns a `Result` that clearly reflects this possible error mode. I've fixed up the wording in the docs.

### DIFF
--- a/ed25519-dalek/src/verifying.rs
+++ b/ed25519-dalek/src/verifying.rs
@@ -124,12 +124,6 @@ impl VerifyingKey {
 
     /// Construct a `VerifyingKey` from a slice of bytes.
     ///
-    /// # Warning
-    ///
-    /// The caller is responsible for ensuring that the bytes passed into this
-    /// method actually represent a `curve25519_dalek::curve::CompressedEdwardsY`
-    /// and that said compressed point is actually a point on the curve.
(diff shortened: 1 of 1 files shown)


robjtede (approved):

pinkforest:
Wouldn't this lose the information what actually the byte string input represents and constructor here is not fallible:
```
 let compressed = CompressedEdwardsY(*bytes);
```
Only the decompress is fallible but isn't it too late and makes figuring out what the correct bytes input hard/er ?

tarcieri:
I think the old warning was kind of bad (point decompression ensures a valid curve point), but it might be good to replace that note with one that it's using the ZIP-215 rules, since the point validation rules being applied have been a source of confusion in the past (see #380, #623)

Genio40921:
> There was a big warning in `VerifyingKey::from_bytes` that said the user is responsible for ensuring that the provided bytes represent a verifying key. I think this was confusing, since the function returns a `Result` that clearly reflects this possible error mode. I've fixed up the wording in the docs.

PR #656, opened 2024-06-05
Enable unexpected cfgs lint

Wuelle:
The `unexpected_cfgs` lint was previously disabled in 9252fa5c0d09054fed. Since the lint can now be configured using `Cargo.toml`[^2] no `build.rs` shenanigans are necessary anymore.

This PR also fixes some immediate bugs found, two tests used comma-seperated config flags `cfg(target_feature = "avx512ifma,avx512vl")`, which is invalid (and always false). Instead, it should be `cfg(all(target_feature = "avx512ifma", target_feature = "avx512vl"))`.
From what I've seen, this pattern also exists in the proc-macros used to enable target features for certain functions. But I don't understand the codebase well enough to say this for certain, and I didn't look too much into it either.

I believe usage `cfg(nightly)` is also a bug - the documentation[^1] seems to suggest that features are automatically enabled when compiled 
> If compiled on a non-nightly compiler, curve25519-dalek will not include AVX512 code, and therefore will never select it at runtime.

This is incorrect, since `cfg(nightly)` does not exist, and AFAIK there is no way to detect whether code is being compiled by a "nightly" compiler (and the definition of "nightly" changes all the time anyways). But perhaps this was intended to be used manually? (via `--cfg nightly`)

[^2]:https://blog.rust-lang.org/2024/05/06/check-cfg.html#expecting-custom-cfgs
[^1]:https://docs.rs/curve25519-dalek/latest/curve25519_dalek/#simd-backend

### DIFF
--- a/curve25519-dalek/Cargo.toml
+++ b/curve25519-dalek/Cargo.toml
@@ -71,3 +71,13 @@ group-bits = ["group", "ff/bits"]
 
 [target.'cfg(all(not(curve25519_dalek_backend = "fiat"), not(curve25519_dalek_backend = "serial"), target_arch = "x86_64"))'.dependencies]
 curve25519-dalek-derive = { version = "0.1", path = "../curve25519-dalek-derive" }
+
+[lints.rust.unexpected_cfgs]
+level = "warn"
+check-cfg = [
+    'cfg(allow_unused_unsafe)',
+    'cfg(curve25519_dalek_backend, values("fiat", "serial", "simd"))',
+    'cfg(curve25519_dalek_diagnostics, values("build"))',
+    'cfg(curve25519_dalek_bits, values("32", "64"))',
+    'cfg(nightly)',
+]
--- a/curve25519-dalek/src/backend/vector/ifma/edwards.rs
+++ b/curve25519-dalek/src/backend/vector/ifma/edwards.rs
@@ -249,7 +249,7 @@ impl<'a> From<&'a edwards::EdwardsPoint> for NafLookupTable8<CachedPoint> {
     }
 }
 
-#[cfg(target_feature = "avx512ifma,avx512vl")]
+#[cfg(all(target_feature = "avx512ifma", target_feature = "avx512vl"))]
 #[cfg(test)]
 mod test {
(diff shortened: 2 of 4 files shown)


tarcieri:
> I believe usage cfg(nightly) is also a bug - the documentation[2](https://github.com/dalek-cryptography/curve25519-dalek/pull/656#user-content-fn-1-142c9cabbb91de7e9b4fc041a060257e) seems to suggest that features are automatically enabled when compiled

They're automatically enabled when compiled because the build script detects the `nightly` compiler and sets `cfg(nightly)` accordingly:

https://github.com/dalek-cryptography/curve25519-dalek/blob/main/curve25519-dalek/build.rs#L38-L44

You'll still need to preserve `cfg(nightly)` for this to work.

Wuelle:
> > I believe usage cfg(nightly) is also a bug - the documentation[2](https://github.com/dalek-cryptography/curve25519-dalek/pull/656#user-content-fn-1-142c9cabbb91de7e9b4fc041a060257e) seems to suggest that features are automatically enabled when compiled
> 
> They're automatically enabled when compiled because the build script detects the `nightly` compiler and sets `cfg(nightly)` accordingly:
> 
> https://github.com/dalek-cryptography/curve25519-dalek/blob/31ccb67/curve25519-dalek/build.rs#L38-L44
> 
> You'll still need to preserve `cfg(nightly)` for this to work.



> > I believe usage cfg(nightly) is also a bug - the documentation[2](https://github.com/dalek-cryptography/curve25519-dalek/pull/656#user-content-fn-1-142c9cabbb91de7e9b4fc041a060257e) seems to suggest that features are automatically enabled when compiled
> 
> They're automatically enabled when compiled because the build script detects the `nightly` compiler and sets `cfg(nightly)` accordingly:
> 
> https://github.com/dalek-cryptography/curve25519-dalek/blob/31ccb67/curve25519-dalek/build.rs#L38-L44
> 
> You'll still need to preserve `cfg(nightly)` for this to work.

Ah thanks, i didn't see the build script. Fixed

tarcieri (approved):

PR #661, opened 2024-06-18
SECURITY: fix timing variability in backend/serial/u32/scalar.rs

tarcieri:
Similar security fix to #659, but for the 32-bit backend. See that PR for more information about the problem.


rozbb:
Relevant compiler outputs (thanks to @tarcieri):

[Without fix](https://godbolt.org/z/zvaWxzvqv)

[With fix](https://godbolt.org/z/jc9j7eb8E)

tarcieri:
@rozbb the godbolt links might be worth including in the commit message when you squash-and-merge

PR #665, opened 2024-06-25
curve: use `subtle::Choice` for constant-time fixes

tarcieri:
Alternative to #659/#661 and #662 which leverages `subtle::Choice` and `subtle::ConditionallySelectable` as the optimization barriers.

Really the previous masking was there to conditionally add the scalar field modulus on underflow, so instead of that, we can conditionally select zero or the modulus using a `Choice` constructed from the underflow bit.

TODO: verify codegen and check benchmarks

### DIFF
--- a/curve25519-dalek/src/backend/serial/u32/scalar.rs
+++ b/curve25519-dalek/src/backend/serial/u32/scalar.rs
@@ -12,7 +12,7 @@
 
 use core::fmt::Debug;
 use core::ops::{Index, IndexMut};
-use subtle::BlackBox;
+use subtle::{Choice, ConditionallySelectable};
 
 #[cfg(feature = "zeroize")]
 use zeroize::Zeroize;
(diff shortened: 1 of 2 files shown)


curve25519-dalek/src/backend/serial/u32/scalar.rs
```diff
@@ -187,23 +187,22 @@ impl Scalar29 {
     /// Compute `a - b` (mod l).
     pub fn sub(a: &Scalar29, b: &Scalar29) -> Scalar29 {
         let mut difference = Scalar29::ZERO;
-        let mask = BlackBox::new((1u32 << 29) - 1);
+        let mask = (1u32 << 29) - 1;
```
tarcieri, line 190: `BlackBox` isn't doing anything here since the mask is a constant value

isislovecruft (approved):

rozbb (approved):

PR #667, opened 2024-07-01
docs: update `VartimeRistrettoPrecomputation` documentation

AaronFeickert:
It is the case that performing variable-time Ristretto multiscalar multiplication evaluation using [`VartimeRistrettoPrecomputation`](https://docs.rs/curve25519-dalek/latest/curve25519_dalek/ristretto/struct.VartimeRistrettoPrecomputation.html) may be less efficient than the corresponding [`VartimeMultiscalarMul`](https://docs.rs/curve25519-dalek/latest/curve25519_dalek/traits/trait.VartimeMultiscalarMul.html) implementation, depending on the number of points used.

This PR updates the documentation to make this clear, since the decision of whether or not to use precomputation may depend highly on the use case.

### DIFF
--- a/curve25519-dalek/src/ristretto.rs
+++ b/curve25519-dalek/src/ristretto.rs
@@ -1007,7 +1007,7 @@ impl VartimeMultiscalarMul for RistrettoPoint {
 }
 
 /// Precomputation for variable-time multiscalar multiplication with `RistrettoPoint`s.
-/// 
+///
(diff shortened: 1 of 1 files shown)


rozbb:
Nice! I can't quite tell what the formatting error is here

AaronFeickert:
Looks like the formatter doesn't like it when an empty comment line includes a trailing space. Fixed.

rozbb:
Thank you!

PR #668, opened 2024-07-01
curve: partial precomputation

AaronFeickert:
It may be useful to support multiscalar multiplication evaluation using partial precomputation, where a larger set of precomputation tables is generated than is needed for an evaluation.

This PR adds such functionality with minimal impact. Specifically, it does so by relaxing the panic conditions on the relevant `VartimePrecomputedMultiscalarMul` functions (and their corresponding implementations). It also adds tests and updates the documentation carefully.

Note that this is an improved version of #546, which was closed due to repository reorganization.


rozbb:
Very nice! New docs and docs look great. Anything else needed @tarcieri ?

tarcieri (approved):

PR #669, opened 2024-07-05
curve: use constant-time compressed equality testing

AaronFeickert:
In line with the [safety goals](https://github.com/dalek-cryptography/curve25519-dalek/tree/main/curve25519-dalek#safety), this PR ensures that `CompressedRistretto` equality testing is always done in constant time.

Previous work in #229 implemented `ConstantTimeEq` for `CompressedRistretto`, but this is not used for `Eq` equality testing. It's already the case that [`RistrettoPoint`](https://github.com/dalek-cryptography/curve25519-dalek/blob/5b7082bbc8e0b2106ab0d956064f61fa0f393cdc/curve25519-dalek/src/ristretto.rs#L822-L826) and [`Scalar`](https://github.com/dalek-cryptography/curve25519-dalek/blob/5b7082bbc8e0b2106ab0d956064f61fa0f393cdc/curve25519-dalek/src/scalar.rs#L294-L298) perform all equality testing in constant time; this PR unifies this behavior for compressed points as well.


AaronFeickert:
Note that dalek-cryptography/subtle#131 would also supply a marker trait that could be useful here to signal this "all equality is constant time" behavior.

curve25519-dalek/src/ristretto.rs
```diff
@@ -215,9 +215,16 @@ use crate::traits::{MultiscalarMul, VartimeMultiscalarMul, VartimePrecomputedMul
 ///
 /// The Ristretto encoding is canonical, so two points are equal if and
 /// only if their encodings are equal.
-#[derive(Copy, Clone, Eq, PartialEq, Hash)]
+#[derive(Copy, Clone, Hash)]
```
tarcieri, line 218: While it's unlikely to cause real-world breakages, I believe this is technically a breaking change.

Here's an example of what is possible with a derived `PartialEq` which won't be possible with this change: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=33e85cff0772c968767a0417a8a7a541

AaronFeickert: Interesting! I was not aware of this behavior. Thoughts on the tradeoff between the functionality and the implications of the breaking change?

tarcieri: I would strongly agree with making this change if it weren't for the potential breakages

AaronFeickert: Ought this PR stay open until the next breaking release? I remain of the opinion that the benefit (consistency with safety goals) outweighs the cost.

rozbb:
Yeah that's unfortunate. I really doubt anyone has ever used a `CompressedRistrettoPoint` in a match statement, especially since this is a [relatively new](https://github.com/rust-lang/rust/issues/120362) feature. But semver is semver. Tbh I'm not opposed to bumping the major version number soon.

AaronFeickert:
Given that this seems to be a desired change, I'll keep it open for the next breaking release. Feel free to close if this isn't the case!

PR #670, opened 2024-07-05
curve: use derive macro for zeroizing

AaronFeickert:
Several types implement `Zeroize` manually. It seems safer to do this via the supplied [derive macro](https://docs.rs/zeroize/latest/zeroize/#derives), since any field changes will be caught automatically.

This PR moves to this functionality.

### DIFF
--- a/curve25519-dalek/Cargo.toml
+++ b/curve25519-dalek/Cargo.toml
@@ -53,7 +53,7 @@ rand_core = { version = "0.6.4", default-features = false, optional = true }
 digest = { version = "0.10", default-features = false, optional = true }
 subtle = { version = "2.6.0", default-features = false }
 serde = { version = "1.0", default-features = false, optional = true, features = ["derive"] }
-zeroize = { version = "1", default-features = false, optional = true }
+zeroize = { version = "1", default-features = false, optional = true, features = ["zeroize_derive"] }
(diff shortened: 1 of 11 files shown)


tarcieri:
Note: the proc macro stack is not currently a hard dependency of `curve25519-dalek` (this is where having `Cargo.lock` checked in is useful for reviewing the effects of dependency changes, particularly in conjunction with GitHub bots that can help monitor / analyze them)

Enabling this feature, while it would save 50 lines and automatically catch field additions / removals, would also pull in the following additional dependencies which are not currently required:

- `proc-macro2`
- `unicode-ident`
- `quote`
- `syn`
- `zeroize_derive`

Now granted, these are commonly used dependencies, but it is a pretty big increase in total dependency surface.

Regarding the problem of field additions/changes: I don't think that's a particularly big problem in practice with this set of types, which are more or less all in their "final form".

AaronFeickert:
That's a fair point. Happy to close if you don't think the added baggage is worth it in this case.

tarcieri:
I think the extra dependencies probably aren't worth the ~50loc it saves in this case, although that could change in the future if there are more usages of `zeroize`.

AaronFeickert:
Noted! Closing.

PR #671, opened 2024-07-08
docs: minor documentation fixes

AaronFeickert:
This PR makes a few small documentation fixes:
- It updates the `ristretto` module documentation to correctly note that `RistrettoPoint` equality testing is always done [in constant time](https://github.com/dalek-cryptography/curve25519-dalek/blob/5b7082bbc8e0b2106ab0d956064f61fa0f393cdc/curve25519-dalek/src/ristretto.rs#L822-L826).
- It updates the `edwards` module documentation to correctly note that `EdwardsPoint` equality testing is always done [in constant time](https://github.com/dalek-cryptography/curve25519-dalek/blob/5b7082bbc8e0b2106ab0d956064f61fa0f393cdc/curve25519-dalek/src/edwards.rs#L495-L499).
- It updates the `EdwardsPoint` documentation to fix what appears to be a copy-paste error for a type name.

### DIFF
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -19,8 +19,8 @@
 //! ## Equality Testing
 //!
 //! The `EdwardsPoint` struct implements the [`subtle::ConstantTimeEq`]
-//! trait for constant-time equality checking, and the Rust `Eq` trait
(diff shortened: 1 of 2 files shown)


rozbb:
Indeed you're right! Thank you for the fixes

PR #675, opened 2024-07-28
Fix incorrect usage in README example

thomasbreydo:
I was using the docs to get started, and noticed that they seem to be out of date. `EphemeralSecret::random()` doesn't exist.

* I've changed the README example to use`random_from_rng()`. 
* I also deleted the "ignore" tag on the example code blocks, since now they should successfully compile.

Please let me know if I'm missing something here, and thank you!

### DIFF
--- a/x25519-dalek/README.md
+++ b/x25519-dalek/README.md
@@ -31,18 +31,20 @@ the rest of the afternoon nomming some yummy pie!
 First, Alice uses `EphemeralSecret::random()` and then
 `PublicKey::from()` to produce her secret and public keys:
 
-```ignore
+```rust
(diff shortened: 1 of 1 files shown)


thomasbreydo:
Oh, I now see that `random()` is indeed a method! I just had to enable the `getrandom` feature.

rozbb:
Thank you for the PR! Yup, we use the `doc_auto_cfg` feature to make sure the feature flag notice shows up in the docs: https://docs.rs/x25519-dalek/latest/x25519_dalek/struct.EphemeralSecret.html#method.random

Gonna close this, since the current docs are a valid use of the API

PR #676, opened 2024-07-28
[WIP] Bump `digest`, `ed25519`, `signature`, and `sha2`

tarcieri:
Bumps the aforementioned dependencies to their latest (pre)releases

### DIFF
--- a/curve25519-dalek/Cargo.toml
+++ b/curve25519-dalek/Cargo.toml
@@ -30,7 +30,7 @@ rustdoc-args = [
 features = ["serde", "rand_core", "digest", "legacy_compatibility", "group-bits"]
 
 [dev-dependencies]
-sha2 = { version = "0.10", default-features = false }
+sha2 = { version = "=0.11.0-pre.4", default-features = false }
 bincode = "1"
 criterion = { version = "0.5", features = ["html_reports"] }
 hex = "0.4.2"
@@ -50,7 +50,7 @@ cfg-if = "1"
 ff = { version = "0.13", default-features = false, optional = true }
 group = { version = "0.13", default-features = false, optional = true }
 rand_core = { version = "0.6.4", default-features = false, optional = true }
-digest = { version = "0.10", default-features = false, optional = true }
+digest = { version = "=0.11.0-pre.9", default-features = false, optional = true }
 subtle = { version = "2.6.0", default-features = false }
 serde = { version = "1.0", default-features = false, optional = true, features = ["derive"] }
 zeroize = { version = "1", default-features = false, optional = true }
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -104,7 +104,7 @@ use core::ops::{Mul, MulAssign};
 use cfg_if::cfg_if;
 
 #[cfg(feature = "digest")]
-use digest::{generic_array::typenum::U64, Digest};
+use digest::{array::typenum::U64, Digest};
 
 #[cfg(feature = "group")]
 use {
--- a/curve25519-dalek/src/ristretto.rs
+++ b/curve25519-dalek/src/ristretto.rs
@@ -173,7 +173,7 @@ use core::ops::{Mul, MulAssign};
 use rand_core::CryptoRngCore;
 
 #[cfg(feature = "digest")]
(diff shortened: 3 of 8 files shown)


wiktor-k:
Hi,

Are there any blockers for un-drafting this PR and merging it? FWIW I'm using -pre versions of these crates and they work really well :)

Thanks for your time! :wave:

tarcieri:
We'd need to create separate branches for stable vs development so not as to block stable work

cvengler:
What is the current state of this? I am currently trying to integrate `ed25519-dalek` in an already existing larger project, that already moved to the pre-release versions in its development branch, thereby causing conflicts. ☹️

tarcieri:
We are discussing how to proceed. We might need to do a major version bump due to how aggressively people are reacting to MSRV bumps

robjtede:
> We might need to do a major version bump

For 1.60 -> 1.72? Surely 1.72 (Aug 2023) is old enough to not warrant that.

> how aggressively people are reacting

Yikes, really?

tarcieri:
The new MSRV will be 1.85 when the `digest` feature is active, due to 2024 edition bumps in the @RustCrypto crates.

You can see how people reacted here: https://github.com/RustCrypto/formats/issues/1684

robjtede:
It's slightly surprising that /formats hopped on 2024 so quickly but I assume it had a good reason and was basically waiting for it.

Still, lockfiles exist to solve this already; doesn't require pinning in manifests. And post-1.84 you can opt in to resolver=3 to get MSRV resolving automatically.

Sucks that people are reacting like that instead of learning about the tools provided to help them.

tarcieri:
I pointed that out, but it seems some people were complaining about uncached `cargo install` failing on old Rust versions (where they could still use `--locked`).

It seems it doesn't matter if there are solutions, people have poorly configured builds and will show up to bitch you out with "Your shitty software which I don't care about broke my build"

robjtede:
I expect (and indeed have experienced) the same vocal minority objecting to (cascading) major releases. The solutions are much simpler for the MSRV case.

tarcieri:
Yeah, it's definitely much simpler not to bump the version, though I'm not sure you can necessarily class them as a "vocal minority" (especially when you're not the one dealing with them)

curve25519-dalek/Cargo.toml
```diff
@@ -68,7 +68,7 @@ precomputed-tables = []
 legacy_compatibility = []
 group = ["dep:group", "rand_core"]
 group-bits = ["group", "ff/bits"]
-digest = ["dep:digest", "digest/core-api"]
+digest = ["dep:digest"]
```
tarcieri, line 71: We could arguably get rid of this entirely

rozbb: As in we always use digest no matter what? I'd be fine with that. Idk why it was gated to begin with

ed25519-dalek/src/errors.rs
```diff
@@ -87,7 +87,7 @@ impl Display for InternalError {
     }
 }
 
-#[cfg(feature = "std")]
+#[cfg(feature = "alloc")]
```
rozbb: Why does this need alloc? Isn't `core::error::Error` always existent? I think we only need `alloc` for the `From` impl below

baloo: because the inner error gets boxed
[curve25519-dalek/benches/dalek_benchmarks.rs](https://github.com/RustCrypto/traits/blob/3b1f80e651bb6f7c1b5dd43a7b8ef531a7a00355/signature/src/error.rs#L28)

curve25519-dalek/Cargo.toml
tarcieri: The dependency can remain `optional = true`. This line doesn't really do anything, but some people prefer to always use explicit namespaced features like this.

ed25519-dalek/src/errors.rs
tarcieri: Indeed. Let's do it.

```suggestion
```

curve25519-dalek/Cargo.toml
tarcieri: I'll leave it for now

PR #677, opened 2024-07-29
Implement ConditionallySelectable for MontgomeryPoint

elichai:


curve25519-dalek/src/montgomery.rs
```diff
@@ -84,6 +84,27 @@ impl ConstantTimeEq for MontgomeryPoint {
...
 }
 
+#[allow(clippy::needless_range_loop)]
+impl ConditionallySelectable for MontgomeryPoint {
+    fn conditional_select(a: &Self, b: &Self, choice: Choice) -> Self {
+        let mut bytes = [0u8; 32];
+        for i in 0..32 {
+            bytes[i] = u8::conditional_select(&a.0[i], &b.0[i], choice);
+        }
+        Self(bytes)
```
tarcieri, line 94: I'd imagine this would work instead:

```suggestion
        Self(<[u8; 32]>::conditional_select(&a.0, &b.0, choice)
```

curve25519-dalek/src/montgomery.rs
```diff
@@ -84,6 +84,27 @@ impl ConstantTimeEq for MontgomeryPoint {
...
+    fn conditional_assign(&mut self, other: &Self, choice: Choice) {
+        for i in 0..32 {
+            self.0[i].conditional_assign(&other.0[i], choice);
+        }
+    }
+    fn conditional_swap(a: &mut Self, b: &mut Self, choice: Choice) {
+        for i in 0..32 {
+            u8::conditional_swap(&mut a.0[i], &mut b.0[i], choice);
+        }
+    }
```
tarcieri, line 105: Are these really necessary versus using the defaults?

curve25519-dalek/src/montgomery.rs
elichai: https://docs.rs/subtle/latest/subtle/trait.ConditionallySelectable.html Is not implemented for arrays, I can open a PR to subtle though :)

tarcieri: The impl is gated under the `const-generics` feature of `subtle`. Unfortunately it seems the docs.rs build doesn't enable it.

elichai: I'll at least open a PR for that lol :)

tarcieri: And really now that `subtle` is MSRV 1.60, it should probably just be enabled by default.

elichai: Fixed :)

tarcieri (approved):

curve25519-dalek/src/montgomery.rs
```diff
@@ -84,6 +84,13 @@ impl ConstantTimeEq for MontgomeryPoint {
     }
 }
 
+#[allow(clippy::needless_range_loop)]
```
tarcieri, line 87: You can probably remove this I would imagine:
```suggestion
```

rozbb:
Thanks!

PR #679, opened 2024-07-30
Update ed25519 and signature crates to latest pre-releases

dvdplm:
The RustCrypto crates are close to a plethora of releases and while trying to get my project compiling I got here.

This PR does not update the `sha2` crate, because doing so breaks a fair amount of stuff. I wanted to test the waters here first before doing work that may or may not be welcome. :)


tarcieri:
Note: I already have #676 opened

dvdplm:
Ohh, my bad then! Tyvm. :)

PR #683, opened 2024-08-02
docs: remove unnecessary `CofactorGroup` documentation

AaronFeickert:
The `CofactorGroup` documentation for `EdwardsPoint` merely notes that Ristretto has a trivial cofactor. This is true, but seems out of place here, and seems like it could confuse the reader into thinking this holds for `EdwardsPoint` as well (which it most certainly [does not](https://github.com/dalek-cryptography/curve25519-dalek/blob/0964f800ab2114a862543ca000291a6e3531c203/curve25519-dalek/src/edwards.rs#L1186-L1189)).


rozbb:
Thank you!

PR #684, opened 2024-08-02
curve!: use constant-time compressed Edwards equality testing

AaronFeickert:
This ensures that `CompressedEdwardsY` equality testing is always done in constant time.

BREAKING CHANGE: This can [break](https://github.com/dalek-cryptography/curve25519-dalek/pull/669#discussion_r1667195282) certain uses of `match`.


AaronFeickert:
See #669 for discussion and similar work on the Ristretto side.

rozbb:
Thank you for these!

PR #685, opened 2024-08-03
curve: add precomputation length

AaronFeickert:
Adds functionality to `VartimePrecomputedMultiscalarMul` to return the number of points in the precomputation.

### DIFF
--- a/curve25519-dalek/src/backend/mod.rs
+++ b/curve25519-dalek/src/backend/mod.rs
@@ -128,6 +128,30 @@ impl VartimePrecomputedStraus {
         }
     }
 
+    pub fn len(&self) -> usize {
+        use crate::traits::VartimePrecomputedMultiscalarMul;
+
+        match self {
+            #[cfg(curve25519_dalek_backend = "simd")]
+            VartimePrecomputedStraus::Avx2(inner) => inner.len(),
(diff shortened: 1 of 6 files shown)


AaronFeickert:
This has been updated to account for the changes in #695, and should be good to go.

rozbb:
LGTM! Can I ask, what was the motivation to add these methods?

AaronFeickert:
@rozbb: I was working on a project that used [partial precomputation](https://github.com/dalek-cryptography/curve25519-dalek/pull/668), and for which it would be useful to know the size of an existing precomputation table in order to determine if it was valid for a given use case.

rozbb:
Ah, ok, sounds good!

curve25519-dalek/src/backend/mod.rs
```diff
@@ -128,6 +128,30 @@ impl VartimePrecomputedStraus {
         }
     }
 
+    pub fn len(&self) -> usize {
```
rozbb, line 131: Only note: could we have docs for these methods?

AaronFeickert: Done!

rozbb (approved):

PR #690, opened 2024-08-14
Update README.md Malleability definition

bluk:
- Remove extra ' in definition of malleability

The README states:

> A signature scheme is considered to produce _malleable signatures_ if a passive attacker with knowledge of a public key _A_, message _m_, and valid signature _σ'_ can produce a distinct _σ'_ such that _σ'_ is a valid signature of _m_ with respect to _A_.

Haven't had my morning coffee but I think the first `σ` in `valid signature _σ'_ can produce a distinct _σ'_ ` should be just `σ` (without the `'`).

### DIFF
--- a/ed25519-dalek/README.md
+++ b/ed25519-dalek/README.md
@@ -144,7 +144,7 @@ In this section, we mention some specific details about our validation criteria,
 
 ## Malleability and the `legacy_compatibility` Feature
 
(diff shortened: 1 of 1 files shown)


tarcieri (approved):

AaronFeickert (approved):
It's not the most modern or general definition, but it works.

rozbb:
Thank you!

Issue #693, opened 2024-08-27
AVX512 code will need to change to handle the whole AVX10 business


workingjubilee:
Hello.

You heard about the whole AVX10 business, I presume? With AVX10/256 and AVX10/512?

We will likely need to adjust what the "avx512" features even mean, unfortunately, as a result of  this. Fortunately, they're unstable, so we can do that! So it might need some modifications in dalek, but it's not like there will be any unforeseen complications from changing nightly features in code that didn't enable it intentionally... right?

https://github.com/dalek-cryptography/curve25519-dalek/blob/0964f800ab2114a862543ca000291a6e3531c203/curve25519-dalek/build.rs#L38-L44

...right?

RalfJung:
In other words: building this crate may fail on nightly when the work on redoing these features happens. The compiler team considers automatic use of nightly features in crates that otherwise work on stable to be a breach of the nightly compiler contract, so this is not considered a breaking change -- but if this happens, it will mean that everyone using nightly to build this crate (or any crate that depends on this crate) will get compilation failures.

For context:
- https://github.com/rust-lang/rust/issues/120804
- https://github.com/rust-lang/rust/pull/124339
- https://github.com/SergioBenitez/version_check/pull/23

The preferred way to avoid this issue is to not enable nightly features automatically, but instead make them cargo features that need to be opt-in. That makes it possible to build the crate on nightly in exactly the same way that it would build on stable, which is extremely valuable for doing compiler development and testing.

Another option is to use something like [autocfg](https://docs.rs/autocfg/1.3.0/autocfg/struct.AutoCfg.html#method.probe_raw) to test whether the concrete kind of code you need works on the current compiler. If that code stops working, the crate can then automatically fall back to the stable code path. However, this reduces the usefulness of nightly as a testing vehicle: the point of nightly is that we can know that things will break long before they break on stable. To achieve this, however, it is important that testing a crate on nightly actually tests the thing that would happen when the nightly became stable. By making the same crate build different code on nightly vs stable, this is not the case any more.

tarcieri:
I can't speak to auto-enabling nightly features, but in general I think `nightly` users need to expect that upstream rustc changes can break nightly builds at any time, and all we can do is release new versions of these crates when such upstream breaking changes happen, and nightly users need to upgrade to the latest versions which such breaking changes occur.

RalfJung:
[...]

Issue #635, opened 2024-02-26
AVX512-IFMA & AVX10 status


pinkforest:
Re: https://rustcrypto.zulipchat.com/#narrow/stream/260040-elliptic-curves/topic/AVX-512.20IFMA.20.2F.20AVX10.20future

This is probably still 1) worth supporting and this may be still 2) set as default without heavy opt-in ? 

The downside by keeping it as default nightly-on is dealing with the nightly stuff changes if any

The upside is that the current AMD consumer flagships seems to be have good numbers and performance-per-watt with it that would indicate the enterprise stuff may work well as well.

I don't have 7950X to test withbut phoronix has some general numbers: https://www.phoronix.com/review/amd-zen4-avx512

intel however seems to have pretty much yanked it by now and AVX10 seems to promise a lot and legacy backwards compatibility. chipsandcheese thinks AVX10 is a [silly idea](https://chipsandcheese.com/2023/10/11/avx10-128-is-a-silly-idea-and-should-be-completely-removed-from-the-specification/) and the register has [interesting article](https://www.theregister.com/2023/08/15/avx10_intel_interviews/) + tom's hardware [hot take](https://www.tomshardware.com/news/intels-new-avx10-brings-avx-512-capabilities-to-e-cores).

This also poses a question how AVX-512 code is tested reasonably and does it need to be changed in near future - it doesn't currently really change at all but anyways :woman_shrugging: 

Thoughts ?

nazar-pc:
On AMD 7970X non-AVX512 version is ~40% slower for Sr25519 signature verification using https://github.com/w3f/schnorrkel, a massive gap in performance

tarcieri:
Closing in favor of #758

PR #695, opened 2024-08-29
Make AVX512IFMA opt-in backend

### LINKED ISSUE #694: Experimental supports cfg
Re:
- https://github.com/dalek-cryptography/curve25519-dalek/issues/635
- https://github.com/dalek-cryptography/curve25519-dalek/issues/693

p.s. how do we implement the opt-in given we have declared backend details as implementation detail.

feature will not do as it poses problems of a) relaying across dependency chain b) not having top-level binary in control

Same as #414 we could use cfg for this - I  propose:

`cfg(curve25519_dalek_unstable = "avx512")`

Which will in turn support for unstable AVX512 through nightly opt-in unstables basis.

pinkforest:
Closes #694 #635 #693

Provides `cfg(curve25519_dalek_backend = "unstable_avx512")`

**Errors**

- If nightly is not in use -> Must use nightly in order to use it (similar to compiler main when attempting to use unstables)
- If not x86_64 simd capable target -> As with simd error out similarly

**Note**: I avoided whole lot of any(curve25519_dalek_backend = "simd", curve25519_dalek_backend = "unstable_avx512") gating because compilre allows multi-value per cfg-key where we emit "simd" in addition to unstable_avx512 when potential use validated giving AVX2 fallback through simd.

If it's desirable to instead have explicit gating then let me know - but it pollutes a lot of gating.

### DIFF
--- a/curve25519-dalek/build.rs
+++ b/curve25519-dalek/build.rs
@@ -84,7 +84,7 @@ fn main() {
             }
         }
         Ok("unstable_avx512") if !nightly => {
-            panic!("Coult not override curve25519_dalek_backend to unstable_avx512 as this is nigthly only.");
+            panic!("Could not override curve25519_dalek_backend to unstable_avx512, as this is nigthly only");
         }
         // default between serial / simd (if potentially capable)
         _ => match is_capable_simd(&target_arch, curve25519_dalek_bits) {
--- a/curve25519-dalek/build.rs
+++ b/curve25519-dalek/build.rs
@@ -35,13 +35,16 @@ fn main() {
 
     println!("cargo:rustc-cfg=curve25519_dalek_bits=\"{curve25519_dalek_bits}\"");
 
-    if rustc_version::version_meta()
+    let nightly = if rustc_version::version_meta()
         .expect("failed to detect rustc version")
         .channel
         == rustc_version::Channel::Nightly
     {
         println!("cargo:rustc-cfg=nightly");
-    }
+        true
+    } else {
+        false
+    };
 
     let rustc_version = rustc_version::version().expect("failed to detect rustc version");
     if rustc_version.major == 1 && rustc_version.minor <= 64 {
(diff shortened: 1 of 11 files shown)


curve25519-dalek/build.rs
```diff
@@ -51,25 +54,44 @@ fn main() {
...
             }
-            // default between serial / simd (if potentially capable)
-            _ => match is_capable_simd(&target_arch, curve25519_dalek_bits) {
-                true => "simd",
-                false => "serial",
-            },
-        };
+        }
+        Ok("unstable_avx512") if !nightly => {
+            panic!("Coult not override curve25519_dalek_backend to unstable_avx512 as this is nigthly only.");
```
rozbb, line 87: Spelling fix: "Could not override curve25519_dalek_backend to unstable_avx512, as this is nightly only."

curve25519-dalek/build.rs
```diff
@@ -51,25 +54,44 @@ fn main() {
...
+                false => panic!("Could not override curve25519_dalek_backend to simd"),
+            }
+        }
+        Ok("unstable_avx512") if nightly => {
+            // simd can only be enabled on x86_64 & 64bit target_pointer_width
+            match is_capable_simd(&target_arch, curve25519_dalek_bits) {
+                true => {
+                    // In addition enable Avx2 fallback through simd stable backend
+                    // NOTE: Compiler permits duplicate / multi value on the same key
+                    println!("cargo:rustc-cfg=curve25519_dalek_backend=\"simd\"");
```
rozbb, line 77: So this prints `curve25519_dalek_backend="unstable_avx512"` and`curve25519_dalek_backend="simd"`? Why both?

curve25519-dalek/src/backend/mod.rs
```diff
@@ -46,14 +46,14 @@ pub mod vector;
 enum BackendKind {
     #[cfg(curve25519_dalek_backend = "simd")]
     Avx2,
-    #[cfg(all(curve25519_dalek_backend = "simd", nightly))]
+    #[cfg(all(curve25519_dalek_backend = "unstable_avx512", nightly))]
```
rozbb, line 49: If both `backend="simd"` and `backend="unstable_avx512"` are set, then are both variants included?

curve25519-dalek/build.rs
```diff
@@ -51,25 +54,44 @@ fn main() {
...
+            match is_capable_simd(&target_arch, curve25519_dalek_bits) {
+                true => "simd",
+                // If override is not possible this must result to compile error
+                // See: issues/532
+                false => panic!("Could not override curve25519_dalek_backend to simd"),
+            }
+        }
+        Ok("unstable_avx512") if nightly => {
+            // simd can only be enabled on x86_64 & 64bit target_pointer_width
+            match is_capable_simd(&target_arch, curve25519_dalek_bits) {
```
rozbb, line 73: Comment should read "AVX512 can only be enabled on...". Also the function name is confusing now. Maybe it should be renamed `is_x86_64()` or something?

curve25519-dalek/src/backend/mod.rs
```diff
@@ -46,14 +46,14 @@ pub mod vector;
...
     Avx512,
     Serial,
 }
 
 #[inline]
 fn get_selected_backend() -> BackendKind {
-    #[cfg(all(curve25519_dalek_backend = "simd", nightly))]
+    #[cfg(all(curve25519_dalek_backend = "unstable_avx512", nightly))]
```
rozbb, line 56: Why is `, nightly` necessary? Isn't it now implicit with the flag? Same for the rest of the file. Actually, it seems like this occurs a lot in the codebase, e.g., in the scalar_mul files below.

rozbb (changes requested):
Thank you! Left some small nits here and there

curve25519-dalek/build.rs
pinkforest: Because we need to have AVX2 fallback at runtime otherwise you get SIGILL - by doing both we get the fallback. I did ask about explicitly gating this but it would pollute a lot of gating and difficult to maintain.

curve25519-dalek/src/backend/mod.rs
pinkforest: yes. unstable_avx512 includes simd as fallback given it would get selected as default in any case with simd capable platform.

curve25519-dalek/build.rs
pinkforest: no. simd is arch non-agnostic and we've juts have x86_64 support but in future may have more so it's apt here.

curve25519-dalek/src/backend/mod.rs
pinkforest: because nightly is required for unstable features. it's for ensuring this does not compile in stable if someone uses the cfg which would lead to SIGILL

curve25519-dalek/build.rs
rozbb: Understood, makes sense to me

rozbb:
Ok this seems good to me, thoughts @tarcieri ?

tarcieri (approved):

rozbb:
Thanks again!

nazar-pc:
Can this be reverted considering AVX512 was [stabilized](https://github.com/rust-lang/rust/pull/138940) in Rust and [AVX10/256 is no longer a thing](https://github.com/dalek-cryptography/curve25519-dalek/issues/693#issuecomment-2737185469). The impact of this change is [very substantial](https://github.com/dalek-cryptography/curve25519-dalek/issues/635#issuecomment-2726284861) on modern AVX512-capable CPUs.

tarcieri:
@nazar-pc a simple revert won't help, since that just goes back to `nightly` autodetection, which is undesirable.

 The code instead needs to actually be updated to support `stable` Rust, once 1.89 is released, I believe.

nazar-pc:
Since `unstable_avx512` wasn't included in any of the published releases yet, reverting it seems like the first step towards stable support anyway since it would only cause more churn. Not sure how much of the original justification is still relevant with AVX512 being already stable in the nightly channel.

Requiring `unstable_avx512` configuration option is a lot less friendly, so I'm worried about the next release dropping performance on modern CPUs to almost half of what it was previously (for those who use nightly of course).

Maybe extending feature check to `nightly` or Rust 1.89.0+ instead could already provide a future-compatible solution, I don't think waiting for stable to be out is a blocker for this.

tarcieri:
Autodetecting nightly is antipattern that has caused massive ecosystem breakages (see e.g. ahash). We had several people from rust-core lecture us that we should not be doing it

Issue #626, opened 2024-02-12
Support NIST validation criteria for Edwards points


tarcieri:
The current implementation uses [ZIP-215](https://zips.z.cash/zip-0215) rules. We've received requests for stricter validation (#380, #623) which correspond to the NIST validation criteria, namely ensuring that the Edwards y-coordinate doesn't overflow the field modulus, and that the resulting point belongs to the prime order subgroup, which map to the NIST partial and full public key validation rules respectively.

NIST defines public key validation criteria in [SP 800-186 Appendix D.1.3: Twisted Edwards Curves](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186.pdf):

> ## D.1.3. Twisted Edwards Curves
> ### D.1.3.1. Partial Public Key Validation
> **Inputs:**
> 1. Edwards curve E<sub>a,d</sub> defined over the prime field GF(p)
> 2. Point Q
>
> **Output:** ACCEPT or REJECT Q as an affine point on E<sub>a,d</sub>.
>
> **Process:**
> 1. Verify that both x and y are integers in the interval [0, p−1]. Output REJECT if verification fails.
> 2. Let Q = (x, y). Verify that (x, y) is a point on E<sub>a,d</sub> by checking that (x, y) satisfies the defining equation ax<sup>2</sup> + y<sup>2</sup> = 1 + dx<sup>2</sup>y<sup>2</sup>, where computations are carried out in GF(p). Output REJECT if verification fails.
> 3. Otherwise, output ACCEPT.
>
> ## D.1.3.2. Full Public Key Validation
> **Inputs:**
> 1. Edwards curve E<sub>a,d</sub> defined over the prime field GF(p)
> 2. Point Q
>
> **Output:** ACCEPT or REJECT Q as a point on E<sub>a,d</sub> of order n.
>
> **Process:**
> 1. Perform partial public key validation on Q using the procedure of Appendix D.1.3.1. Output REJECT if this procedure outputs REJECT.
> 2. If Q = (0,1), output REJECT.
> 3. Verify that nQ = (0,1). Output REJECT if verification fails.
> 4. Otherwise, output ACCEPT.

Some possibilities for APIs:

- Inherent methods: e.g. `EdwardsPoint::decompress_nist_partial` / `EdwardsPoint::decompress_nist_full`
- Single method which accepts an enum specifying the validation criteria: `EdwardsPoint::decompress_with(PointValidation::NistPartial)`

[...]

PR #704, opened 2024-09-19
ed: update `VerifyingKey::from_bytes` with ZIP-215 info

tarcieri:
Removes the previous warning that points are unvalidated: they're validated using the ZIP-215 rules, which allows unreduced y-coordinates. Points are ensured valid by performing decompression, which finds a solution to the curve equation, or returns an error.

Adds references to ZIP-215 and dalek-cryptography/curve25519-dalek#626 which is an issue about potentially adding support for the RFC8032/NIST validation criteria in the future.


TheButlah (approved):
Makes sense. Thanks!

NikolaWissenklaus:
Great!

PR #710, opened 2024-09-30
{curve,ed}25519-dalek: clippy fixes

baloo:
Clippy 1.81 brings new lints, this fixes those warnings


curve25519-dalek/src/backend/vector/packed_simd.rs
```diff
@@ -240,7 +240,9 @@ impl u64x4 {
     pub const fn new_const(x0: u64, x1: u64, x2: u64, x3: u64) -> Self {
         // SAFETY: Transmuting between an array and a SIMD type is safe
         // https://rust-lang.github.io/unsafe-code-guidelines/layout/packed-simd-vectors.html
-        unsafe { Self(core::mem::transmute([x0, x1, x2, x3])) }
+        unsafe {
+            Self(core::mem::transmute::<[u64; 4], core::arch::x86_64::__m256i>([x0, x1, x2, x3]))
```
tarcieri, line 244: Will merge this as-is to prevent conflicts on your other branch, but `core::arch::x86_64::__m256i` is probably worth importing so you can just call this `__m256i` (here and below)

tarcieri (approved):

PR #716, opened 2024-10-22
feat : Trait Reduce implemented for U256 and U512 on struct Scalar

### LINKED ISSUE #688: Implement `Reduce` trait
# Inconsistent scalar reduction across different curve implementations

## Context
I'm developing a threshold signing library that needs to be generic over different elliptic curves, specifically targeting `k256` and `curve25519-dalek`. The library currently uses the `Group` trait for most operations, but I've encountered an issue with scalar reduction from U256.

## Problem
When implementing generic operations like Discrete Log proofs, I need to reduce a U256 value to a curve scalar. However, there's inconsistency in how different curve implementations handle this:

- `k256::Scalar` implements the [`Reduce`](https://docs.rs/k256/latest/k256/struct.Scalar.html#impl-Reduce%3CUint%3Ccrypto_bigint::::uint::U256::%7Bconstant%230%7D%3E%3E-for-Scalar) trait.
- `curve25519_dalek::Scalar` doesn't implement `Reduce`. Instead, it provides:
  - A static method `from_bytes_mod_order`
  - An implementation of the `FromUniformBytes` trait

There's no common trait that both curves implement for modular reduction, which complicates writing generic code.

## Questions
1. Are there plans to implement the `Reduce` trait for `curve25519-dalek::Scalar`?
2. If not, is there a recommended workaround or best practice for handling this discrepancy in generic code?

## Additional Information
- Dependency versions:
  - k256: `0.13.2`
  - curve25519-dalek: `4.1.3` using `group` feature. 

Any guidance or suggestions would be greatly appreciated. Thank you!

mahmudsudo:
This implementation will help standardize scalar reduction across different elliptic curve libraries, addressing the inconsistency 
closes #688

### DIFF
--- a/curve25519-dalek/src/traits.rs
+++ b/curve25519-dalek/src/traits.rs
@@ -421,3 +421,9 @@ pub(crate) trait ValidityCheck {
     /// Checks whether the point is on the curve. Not CT.
     fn is_valid(&self) -> bool;
 }
+
+// Reduce trait 
+
+pub trait Reduce<T>{
+    fn reduce(value:T) -> Self
+}
--- a/curve25519-dalek/src/scalar.rs
+++ b/curve25519-dalek/src/scalar.rs
@@ -1362,7 +1362,17 @@ fn read_le_u64_into(src: &[u8], dst: &mut [u64]) {
         );
     }
 }
+// Implementing the Reduce trait for curve25519_dalek::Scalar
+impl Reduce<U256> for Scalar {
+    fn reduce(value: U256) -> Self {
+        Self::from_bytes_mod_order(value.to_bytes())
+    }
+}
 
+impl Reduce<U512> for Scalar {
+    fn reduce(value: U512) -> Self {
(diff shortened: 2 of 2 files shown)


mahmudsudo:
@tarcieri

curve25519-dalek/src/traits.rs
```diff
@@ -421,3 +421,9 @@ pub(crate) trait ValidityCheck {
     /// Checks whether the point is on the curve. Not CT.
     fn is_valid(&self) -> bool;
 }
+
+// Reduce trait 
+
+pub trait Reduce<T>{
```
rozbb, line 427: Why is this re-defined here? Isn't the point of the referenced issue that we should be implementing the `Reduce` trait from `elliptic_curve`?

tarcieri: Also note we'll have another breaking release of `elliptic-curve` out soon, probably early 2025, so it would probably make sense to target these changes along #676

PR #721, opened 2024-12-15
x25519: add RFC7748 Diffie-Hellman Curve25519 tests

shampoofactory:
Hi. Whilst x25519 `RFC7748 5.2. Test vector` tests are present, `RFC7748 6.1. Curve25519 Test vector` tests appear to be missing. I've added them in.

Thank you.

Reference: https://datatracker.ietf.org/doc/html/rfc7748#section-6.1


AaronFeickert (approved):
Checked that the test vectors match those in the RFC.

rozbb:
Tested locally, thank you!

PR #712, opened 2024-10-03
Implements `DynSignatureAlgorithmIdentifier` trait for ed25519

juliusl:
- Implements `DynSignatureAlgorithmIdentifier` trait for SigningKey/VerifyingKey from ed25519-dalek

Addresses #707

Issue #707, opened 2024-09-25
Implement RFC8410: Algorithm Identifiers for Ed25519, Ed448, X25519, and X448 for Use in the Internet X.509 Public Key Infrastructure


stevefan1999-personal:
https://datatracker.ietf.org/doc/html/rfc8410

We can workaround this by using a wrapper that implements https://docs.rs/spki/latest/spki/trait.DynSignatureAlgorithmIdentifier.html...

juliusl:
Realized that the x509 should've probably been a separate issue so I removed it from the PR

tarcieri:
>  I worked around this in my PR but I'm not sure if it's better to fix ed25519 first or to just fix the issue in this crate.

@juliusl if you can add support in the `ed25519` crate that would be great. I will try to open a tracking issue for that.

PR #728, opened 2025-01-22
ed25519-dalek: bring signature identifiers

baloo:
This allows to use ed25519 to create certificates for example.

Related to #707 & https://github.com/RustCrypto/signatures/pull/889

Partially reverts #712

### DIFF
--- a/.github/workflows/ed25519-dalek.yml
+++ b/.github/workflows/ed25519-dalek.yml
@@ -20,7 +20,7 @@ env:
 jobs:
 
   msrv:
-    name: Current MSRV is 1.72.0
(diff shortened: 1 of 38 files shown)


rozbb:
Also approved. Just wanna rerun the tests somehow

baloo:
Rebased to get the tests running, realized it was partially fixed by #712 although I don't believe we needed the `DynSignatureAlgorithmIdentifier` but `SignatureAlgorithmIdentifier` was enough.

baloo:
for the rest of the tests, I think the fix is in #762

PR #729, opened 2025-01-28
Bump to rand[_core] 0.9

pinkforest:
@tarcieri Any idea re: this complaint with 1.84 on dev-deps `rand_core, .. features = ["getrandom"]`

Using `getrandom` (which is implicit dep)
```
he package `curve25519-dalek` depends on `rand_core`, with features: `getrandom` but `rand_core` does not have these features.
 It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
```

Using `dep:getrandom` as it suggests
```toml
feature `dep:getrandom` in dependency `rand_core` is not allowed to use explicit `dep:` syntax
  If you want to enable an optional dependency, specify the name of the optional dependency without the `dep:` prefix, or specify a feature from the dependency's `[features]` table that enables the optional dependency.
```

### DIFF
--- a/curve25519-dalek/Cargo.toml
+++ b/curve25519-dalek/Cargo.toml
@@ -34,8 +34,8 @@ sha2 = { version = "0.10", default-features = false }
 bincode = "1"
 criterion = { version = "0.5", features = ["html_reports"] }
 hex = "0.4.2"
-rand = "0.8"
-rand_core = { version = "0.6", default-features = false, features = ["getrandom"] }
+rand = "0.9"
+rand_core = { version = "0.9", default-features = false, features = ["getrandom"] }
 
 [build-dependencies]
 rustc_version = "0.4.0"
@@ -49,7 +49,7 @@ required-features = ["alloc", "rand_core"]
 cfg-if = "1"
 ff = { version = "0.13", default-features = false, optional = true }
 group = { version = "0.13", default-features = false, optional = true }
-rand_core = { version = "0.6.4", default-features = false, optional = true }
+rand_core = { version = "0.9", default-features = false, optional = true }
 digest = { version = "0.10", default-features = false, optional = true }
 subtle = { version = "2.6.0", default-features = false, features = ["const-generics"]}
 serde = { version = "1.0", default-features = false, optional = true, features = ["derive"] }
--- a/x25519-dalek/Cargo.toml
+++ b/x25519-dalek/Cargo.toml
@@ -39,14 +39,14 @@ features = ["getrandom", "reusable_secrets", "serde", "static_secrets"]
 
 [dependencies]
 curve25519-dalek = { version = "4", path = "../curve25519-dalek", default-features = false }
-rand_core = { version = "0.6", default-features = false }
+rand_core = { version = "0.9", default-features = false }
 serde = { version = "1", default-features = false, optional = true, features = ["derive"] }
 zeroize = { version = "1", default-features = false, optional = true, features = ["zeroize_derive"] }
 
 [dev-dependencies]
 bincode = "1"
 criterion = "0.5"
-rand_core = { version = "0.6", default-features = false, features = ["getrandom"] }
+rand_core = { version = "0.9", default-features = false, features = ["getrandom"] }
 
 [[bench]]
 name = "x25519"
--- a/ed25519-dalek/Cargo.toml
+++ b/ed25519-dalek/Cargo.toml
@@ -34,7 +34,7 @@ subtle = { version = "2.3.0", default-features = false }
 
 # optional features
 merlin = { version = "3", default-features = false, optional = true }
-rand_core = { version = "0.6.4", default-features = false, optional = true }
+rand_core = { version = "0.9", default-features = false, optional = true }
 serde = { version = "1.0", default-features = false, optional = true }
 zeroize = { version = "1.5", default-features = false, optional = true }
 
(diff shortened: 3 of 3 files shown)


tarcieri:
@pinkforest it needs to be `os_rng` now

NimonSour:
Since `rand_core` changed the trait for `OsRng` from `CryptoRngCore`( which is `RngCore + CryptoRng` ) to `TryRngCore` +`TryCryptoRng`it makes sense to keep all the methods of types across `dalek` crates that are dependent on `CryptoRngCore` ( like `generate` and `random` ) as `R : RngCore + CryptoRng ` plus adding the additional methods ( `try_generate` and `try_random` ) respectively where `R : TryRngCore + CryptoRngCore`.

Example for ed25519/src/signing.rs
```rust 
    pub fn generate<R: RngCore + CryptoRng + ?Sized>(csprng: &mut R) -> SigningKey {
        let mut secret = SecretKey::default();
        csprng.fill_bytes(&mut secret);
        Self::from_bytes(&secret)
    }
    
    pub fn try_generate<R: TryRngCore + TryCryptoRng + ?Sized>(csprng: &mut R) -> Result<SigningKey,<R as TryRngCore>::Error> {
        let mut secret = SecretKey::default();
        csprng.try_fill_bytes(&mut secret)?;
        Ok(Self::from_bytes(&secret))
    }

```
It will offer some backward compatibility while aligning with `rand_core`'s `RngCore`and `TryRngCore` traits, where the user has a choice on RNG implementation.

tarcieri:
> it makes sense to keep all the methods of types across dalek crates that are dependent on CryptoRngCore ( like generate and random ) as R : RngCore + CryptoRng 

`CryptoRng` now has a supertrait bound on `RngCore`, so instead `CryptoRngCore` should just be replaced with just `CryptoRng`. Adding `RngCore + ...` is redundant due to the supertrait bound. See:

https://docs.rs/rand_core/0.9.0/rand_core/trait.CryptoRng.html

Also you can call `OsRng.unwrap_err()` to get a `CryptoRng`-friendly version of `OsRng`.

pinkforest:
I assume we can live with ff/group will get duplicate 0.8 rand until ff/group upgrades - unblocking us to adopt 0.9 ?

baloo:
Could you retarget `rustcrypto-new-releases`?

nresare:
Since it seems we have ended up in a bit of a holding pattern where everyone waits on everyone else to make the move to rand 0.9.x I went ahead and scoped out what would be needed for getting at least all the tests to compile. The changes needed can be viewed in [this commit](https://github.com/nresare/curve25519-dalek/commit/22282150a326cd81f466dcfaef4f398515c85b1e) but it seemed impolite to open a second PR pretty much the same thing. 

@pinkforest please let me know how/if you would like to incorporate those changes, I'm happy to see this addressed in whatever way. Some observations:

* To the best of my understanding, it is not possible to have both rand 0.9 and <0.9 in the dependency tree at the same time. Which means that we need to ensure that no dependency brings the old versions.
* This translates into a requirement to have new releases of `ff` and `group`. There are already a prerelease branch of `ff` 0.14 and a PR that updates `group` 
* A somewhat more complex case is the dev-dependency `merlin` which in turn depends back on `curve25519`. I'll open a PR to at least get the ball rolling on getting that sorted, but I would imagine that this requires a coordingated release of new packages. Fun.
* The bulk of the changes needed are related to `OsRng` no longer implementing `CryptoRng`, instead opting to implement TryCryptorRng (presumably because things like reading from /dev/urandom means filesystem IO theoretically can fail). I have opted to use the ThreadRng available from `rand::rng()` but I guess we could change that to a OsRng wrapper that panics on failures

tarcieri:
> The bulk of the changes needed are related to OsRng no longer implementing CryptoRng, instead opting to implement TryCryptorRng (presumably because things like reading from /dev/urandom means filesystem IO theoretically can fail). I have opted to use the ThreadRng available from rand::rng() but I guess we could change that to a OsRng wrapper that panics on failures

You can get an `OsRng` that panics on failures with:

```rust
OsRng.unwrap_err()
```

...which returns an `UnwrapErr<OsRng>` that impls `CryptoRng`.

`OsRng` is a more conservative choice for things like key generation. Userspace RNGs have more sharp edges, e.g. fork safety, VM suspend/resume

curve25519-dalek/Cargo.toml
```diff
@@ -34,8 +34,8 @@ sha2 = { version = "0.10", default-features = false }
 bincode = "1"
 criterion = { version = "0.5", features = ["html_reports"] }
 hex = "0.4.2"
-rand = "0.8"
-rand_core = { version = "0.6", default-features = false, features = ["getrandom"] }
+rand = "0.9"
+rand_core = { version = "0.9", default-features = false, features = ["getrandom"] }
```
tcerqueira, line 38:
```suggestion
rand_core = { version = "0.9", default-features = false, features = ["os_rng"] }
```

x25519-dalek/Cargo.toml
```diff
@@ -39,14 +39,14 @@ features = ["getrandom", "reusable_secrets", "serde", "static_secrets"]
...
-rand_core = { version = "0.6", default-features = false }
+rand_core = { version = "0.9", default-features = false }
 serde = { version = "1", default-features = false, optional = true, features = ["derive"] }
 zeroize = { version = "1", default-features = false, optional = true, features = ["zeroize_derive"] }
 
 [dev-dependencies]
 bincode = "1"
 criterion = "0.5"
-rand_core = { version = "0.6", default-features = false, features = ["getrandom"] }
+rand_core = { version = "0.9", default-features = false, features = ["getrandom"] }
```
tcerqueira, line 49:
```suggestion
rand_core = { version = "0.9", default-features = false, features = ["os_rng"] }
```

tarcieri:
Closing in favor of #777

Issue #235, opened 2019-02-23
what is the best way to get the x coordinate of a ristretto point ? 


omershlo:

burdges:
It's not considered well defined because a non-cannonical choice gets made by the ristretto implementation.  It'd risks only partially compliant implementations if one specifies a particular choice and exposes the mapping to the curve, which then increases testing complexity for users, etc.  You'll likely want to hash the `CompresedRistretto` instead.

omershlo:
@burdges thanks for the answer. But I must say I didn't understand it.
There should not suppose to be any problem to use ristretto to implement digital signatures and in some of them (ecdsa for example) we need x coordinate. 
x coordinate is well defined for ed25519 curve and any other elliptic curve. What's wrong with https://doc.dalek.rs/src/curve25519_dalek/ristretto.rs.html#249 ?

tarcieri:
@omershlo I'm curious why you would want to use ECDSA in combination with Ristretto. Now that the Schnorr patent has expired, Schnorr signatures (e.g. EdDSA) are generally preferred in new designs. ECDSA's design (and vicariously DSA's design) are they way they are in order to avoid the Schnorr patent).

Is there a specific property of ECDSA you want, or are you just looking for a way to create signatures with Ristretto?

I'm unfamiliar with anyone using ECDSA with anything besides (short) Weierstrass curves, aside from the "FrankenECDSA" ECDSA-in-Edwards concept briefly explored by the CFRG (i.e. use Edwards x-coordinate mod l as r), but with the Schnorr patent expired it makes little sense to pursue such an approach.

omershlo:
I certainly don't want to use ECDSA in combination with ristretto :) 

my claim is simple - right now it is not possible to extract/ calculate the x coordinate of a point. And my question is if you can allow some way to do it? (or if there is a way that I didn't find).

I accept @burdges claim that hash of compressed ristretto will work for many cases and that there should not be an API for x coordinate. But this is the first elliptic curve library that I looked into that you can't even calculate to x-coor yourself if you really really really want to.

hdevalence:
> right now it is not possible to extract/ calculate the x coordinate of a point

Yup! Ristretto doesn't provide an elliptic curve, it provides an abstract group, so all you can ask for is the canonical encoding. This provides safety benefits, but allows a lot of implementation flexibility -- implementations aren't even required to use Curve25519 internally! 

[...]

PR #734, opened 2025-03-03
Add methods to expose inner elements

ohaddahan:
Currently `RistrettoPoint` hides the `EdwardsPoint`.
And `EdwardsPoint` hides `X,Y,Z,T` , but sometimes (like in my case), we need access to them.
Hence adding access methods.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ build*.txt
 *.bak
 
 *.s
+.idea
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -375,6 +375,13 @@ pub struct EdwardsPoint {
     pub(crate) T: FieldElement,
 }
 
+impl EdwardsPoint {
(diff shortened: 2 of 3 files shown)


tarcieri:
Access to the raw extended twisted Edwards coordinates, particularly in the context of Ristretto, is a huge footgun. See past discussion at [#235](https://github.com/dalek-cryptography/curve25519-dalek/issues/235).

You have not spelled out your use case. Please start by opening an issue to discuss why you need raw coordinate access.

ohaddahan:
> Access to the raw extended twisted Edwards coordinates, particularly in the context of Ristretto, is a huge footgun. See past discussion at [#235](https://github.com/dalek-cryptography/curve25519-dalek/issues/235).
> 
> You have not spelled out your use case. Please start by opening an issue to discuss why you need raw coordinate access.

Want to use https://github.com/Electron-Labs/ed25519-circom/blob/main/circuits/scalarmul.circom , so need to get access to the underlying data.

PR #735, opened 2025-03-08
Sign/verify by digest update + StreamVerifier

robjtede:
see description and discussion in https://github.com/dalek-cryptography/curve25519-dalek/pull/556

this PR reuses all commits from that one, and adds in the feature guard we talked about

~nightly CI failure is unrelated~

### DIFF
--- a/ed25519-dalek/src/hazmat.rs
+++ b/ed25519-dalek/src/hazmat.rs
@@ -311,4 +311,111 @@ mod test {
             .unwrap();
         raw_verify_prehashed::<CtxDigest, MsgDigest>(&vk, h, Some(ctx_str), &sig).unwrap();
     }
+
+    #[test]
+    fn sign_byupdate() {
+        // Generate the keypair
+        let mut rng = OsRng;
+        let esk = ExpandedSecretKey::random(&mut rng);
+        let vk = VerifyingKey::from(&esk);
+
+        let msg = b"realistic";
+        // signatures are deterministic so we can compare with a good one
+        let good_sig = raw_sign::<CtxDigest>(&esk, msg, &vk);
+
+        let sig = raw_sign_byupdate::<CtxDigest, _>(
+            &esk,
+            |h| {
+                h.update(msg);
+                Ok(())
+            },
+            &vk,
+        );
+        assert!(sig.unwrap() == good_sig, "sign byupdate matches");
+
+        let sig = raw_sign_byupdate::<CtxDigest, _>(
+            &esk,
+            |h| {
+                h.update(msg);
+                Err(SignatureError::new())
+            },
+            &vk,
+        );
+        assert!(sig.is_err(), "sign byupdate failure propagates");
+
+        let sig = raw_sign_byupdate::<CtxDigest, _>(
+            &esk,
+            |h| {
+                h.update(&msg[..1]);
+                h.update(&msg[1..]);
+                Ok(())
+            },
+            &vk,
+        );
+        assert!(sig.unwrap() == good_sig, "sign byupdate two part");
+    }
+
+    #[test]
+    fn verify_byupdate() {
+        // Generate the keypair
+        let mut rng = OsRng;
+        let esk = ExpandedSecretKey::random(&mut rng);
+        let vk = VerifyingKey::from(&esk);
+
+        let msg = b"Torrens title";
+        let sig = raw_sign::<CtxDigest>(&esk, msg, &vk);
+        let wrong_sig = raw_sign::<CtxDigest>(&esk, b"nope", &vk);
+
+        let r = raw_verify_byupdate::<CtxDigest, _>(
+            &vk,
+            |h| {
+                h.update(msg);
+                Ok(())
+            },
+            &sig,
+        );
+        assert!(r.is_ok(), "verify byupdate success");
+
+        let r = raw_verify_byupdate::<CtxDigest, _>(
+            &vk,
+            |h| {
+                h.update(msg);
+                Ok(())
+            },
+            &wrong_sig,
+        );
+        assert!(r.is_err(), "verify byupdate wrong fails");
+
+        let r = raw_verify_byupdate::<CtxDigest, _>(
+            &vk,
+            |h| {
+                h.update(&msg[..5]);
+                h.update(&msg[5..]);
+                Ok(())
+            },
+            &sig,
+        );
+        assert!(r.is_ok(), "verify byupdate two-part");
+
+        let r = raw_verify_byupdate::<CtxDigest, _>(
+            &vk,
+            |h| {
+                h.update(msg);
+                h.update(b"X");
+                Ok(())
+            },
+            &sig,
+        );
+        assert!(r.is_err(), "verify byupdate extra fails");
+
+        let r = raw_verify_byupdate::<CtxDigest, _>(
+            &vk,
+            |h| {
+                h.update(msg);
+                Err(SignatureError::new())
+            },
+            &sig,
+        );
+        assert!(r.is_err(), "verify byupdate error propagates");
+    }
 }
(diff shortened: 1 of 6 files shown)


tarcieri (approved):
I'll give a tentative utACK with questions about if this has impact on benchmarks.

The main notable thing about this PR IMO is it refactors the `recompute_R` function into a stateful/IUF-ish `RCompute` struct. That seems fine to me.

I'll also argue the core rationale for this PR is similar to the "external mu" approach of ML-DSA in that rather than making a mutant IUF variant of an algorithm (i.e. Ed25519ph) you instead expose an API which allows IUF-like operations on standard Ed25519 signatures. Though "external mu" is perhaps a lot more elegant than this, but I digress.

ed25519-dalek/src/verifying.rs
```diff
@@ -511,6 +504,78 @@ impl VerifyingKey {
...
+    h: CtxDigest,
+}
+
+#[allow(non_snake_case)]
+impl<CtxDigest> RCompute<CtxDigest>
+where
+    CtxDigest: Digest<OutputSize = U64>,
+{
+    // If `prehash_ctx.is_some()`, this does the prehashed variant of the computation using its contents.
+    pub fn compute(
```
rozbb: Nit: I think these should be `pub(crate)` just for clarity's sake, since we're not exposing `RCompute`

ed25519-dalek/tests/ed25519.rs
```diff
@@ -334,6 +334,45 @@ mod integrations {
...
+            verifier.finalize_and_verify().is_err(),
+            "Verification of a signature on a different message passed!"
+        );
+
+        let mut verifier = keypair.verify_stream(&good_sig).unwrap();
+        verifier.update(&bad);
+        assert!(
+            verifier.finalize_and_verify().is_err(),
+            "Verification of a signature on a different message passed!"
+        );
```
rozbb, line 378: Could we also have a test of the case where you call `verifier.update()` multiple times before finalizing?

ed25519-dalek/src/verifying.rs
```diff
@@ -253,7 +208,30 @@ impl VerifyingKey {
...
+
+    #[allow(non_snake_case)]
+    pub(crate) fn raw_verify_byupdate<CtxDigest, F>(
+        &self,
+        msg_update: F,
+        signature: &ed25519::Signature,
+    ) -> Result<(), SignatureError>
+    where
+        CtxDigest: Digest<OutputSize = U64>,
+        F: Fn(&mut CtxDigest) -> Result<(), SignatureError>,
```
rozbb: This can actually be an `FnOnce`, since it's only called once. Ditto for the `hazmat` public function

ed25519-dalek/src/hazmat.rs
```diff
@@ -202,6 +230,26 @@ where
...
 
+/// Performs an ordinary Ed25519 verification check, with the message passed incrementally.
+///
+/// Instead of passing the message directly ([`raw_verify()`]), the caller
+/// provides a `msg_update` closure that will be called to feed the
+/// hash of the message being verified.
+///
+/// `CtxDigest` is the digest used to calculate the pseudorandomness needed for signing.
+/// According to the Ed25519 spec, `CtxDigest = Sha512`.
+pub fn raw_verify_byupdate<CtxDigest, F>(
```
rozbb: Remind me, why do we want this function? I can't think of a scenario where this supports streaming. `msg_update` would have to be a synchronous IO function, which I don't imagine people wanna use. And if the answer, like [discussed previously](https://github.com/dalek-cryptography/curve25519-dalek/pull/556#issuecomment-1707656494), is "the message might be non-contiguous in memory", then I think this function could be better written as taking `msgs: &[&[u8]]`, and feeding each slice into the hash.

So I think we can remove this or else change its type signature to something simpler. Thoughts?

ed25519-dalek/src/hazmat.rs
```diff
@@ -169,6 +169,34 @@ where
...
+/// calculate the pseudorandomness needed for signing. According to the Ed25519 spec, `CtxDigest =
+/// Sha512`.
+///
+///
+/// # ⚠️  Cryptographically Unsafe
+///
+/// Do NOT use this function unless you absolutely must. Using the wrong values in
+/// `ExpandedSecretKey` can leak your signing key. See
+/// [here](https://github.com/MystenLabs/ed25519-unsafe-libs) for more details on this attack.
+pub fn raw_sign_byupdate<CtxDigest, F>(
```
rozbb: Are you sure it makes sense to have this function? The (prefixed) hash of `M` needs to be computed twice for signing. This restricts the kinds of things the `msg_update` closure can do. Eg it cannot consume an incoming network stream, since it'll have nothing to hash the second time it's called.

So I propose: don't make new signing procedures, and leave the current signing procedures alone. Either that, or, if we really want, we can make it take `msgs: &[&[u8]]` to handle non-continguous messages (as suggested below). What do you think?

rozbb (changes requested):
Thank you! Left some notes on ways we can simplify this. I've always got an eye towards long-term maintainability, so I really lean towards having less than more. Let me know if I cut too deep :)

rozbb:
@tarcieri re performance: I see no statistically significan't changes on my M1 Air, EXCEPT I get an 11% slowdown on strict signature verification. Regular signature verification is unaffected though. I'm baffled at that

ed25519-dalek/src/hazmat.rs
robjtede: I'm personally happy to remove this as it was added as part of [the second iteration](https://github.com/dalek-cryptography/ed25519-dalek/pull/304) by @mkj

ed25519-dalek/src/hazmat.rs
robjtede: happy to remove this one too as my main interest is the new StreamVerifier which works fine without these new functions

robjtede: removed in 9a8fd748a686f7edf0fd573f9fd075179f3b3380

ed25519-dalek/src/hazmat.rs
robjtede: removed in 9a8fd748a686f7edf0fd573f9fd075179f3b3380

ed25519-dalek/src/hazmat.rs
rozbb: Ty! Can you also remove the second line of the changelog entry and `raw_sign_byupdate`?

ed25519-dalek/tests/ed25519.rs
robjtede: added just above this one

ed25519-dalek/src/hazmat.rs
rozbb: Sorry, should have been clearer. All of the `raw_sign_byupdate` methods. Including [this one](https://github.com/dalek-cryptography/curve25519-dalek/pull/735/files#diff-c2b0bd5917ba51ad0796423378cdd0b0689a299a40d1adca005a15f3235a987fR861), and revert `raw_sign` to the original

mkj: The rationale is in https://github.com/dalek-cryptography/curve25519-dalek/pull/556#issuecomment-1707656494

The signature is over a message made up of various parts, I'm using the serialiser to feed directly into the hash function. This is an example of using it, instead I've had to allocate a few hundred byte buffer to write into then sign (and it has fixed username length limitation assumptions). A `&[&[u8]]` wouldn't work for my case.

https://github.com/mkj/sunset/blob/3c7d6879da93684832f8ec19df4191523999a2f3/src/sign.rs#L324-L338

rozbb: Ahh ok I’m open to this. I feel bad for requesting so many changes. I can break this into a separate PR today

tarcieri:
See also: #763

rozbb:
Ok I tested on an M3 MBP and can't observe a performance difference at all. I'm willing to believe that the regression is spurious.

PR #739, opened 2025-04-03
chore: fix quotation marks

wgyt:
Some editors could have trouble accurately rendering these characters, with these characters showing up as gibberish or causing formatting problems. It will confuse readers.


robjtede (approved):

tarcieri:
Unless you can show a specific editor which is having trouble with these characters, I don't think this PR is justified

PR #746, opened 2025-04-27
Implementation of `elliptic-curve` traits for curve25519 and ed25519.

carloskiki:
The driving reason for implementing this is to eventually implement the `hash2curve` traits   for elligator2.

See also: #612 implements elligator2 without `hash2curve` traits.

## Main Concern for `Ed25519`

The `CurveArithmetic` trait requires an `AffinePoint` for which this crate only has `CompressedEdwardsY`. The problem is that `CurveArithmetic` requires an Infallible conversion from `AffinePoint` (`CompressedEdwardsY`) to `ProjectivePoint` (`EdwardsPoint`).

Currently, this implementation skips the fallible implementation of `CompressedEdwardsY` -> `EdwardsPoint` and creates an invalid `EdwardsPoint` if `CompressedEdwardsY` is invalid (we probably don't want this).

I can think of 2 other solutions:
- We introduce a new `pub AffineEdwardsPoint { x: FieldElement, y: FieldElement }` (with private internals) to implement `Curve` & `CurveArithmetic`. I believe this would not be a breaking change.
- We make the internals of `CompressedEdwardsY` private (__Breaking change__), and introduce a type alias for the the wire representation of `CompressedEdwardsY` e.g.:
  ```rust
  type EdwardsYBytes = [u8; 32];
  ```
  Then the creation of `CompressedEdwardsY` would require validation and would be fallible.

## Main Concern for `Curve25519`:

It does not seem like there is a public `ProjectivePoint` for `curve25519`. One solution would be to make the current one public.

## Note

Only `Ed25519` has the traits implemented (hence why this is a draft), but I would like feedback on this implementation, as I feel the implementation for `Curve25519` will have similar issues. The code contains `QUESTION`s that explain the issues faced.

Both `Ed25519` and `Curve25519` will be housed in the crate `curve25519`, since this is where the arithmetic is implemented for both curves.

### DIFF
--- a/curve25519-dalek/src/ed25519.rs
+++ b/curve25519-dalek/src/ed25519.rs
@@ -1,0 +1,24 @@
+use elliptic_curve::{bigint::U256, consts::U32, Curve, CurveArithmetic, FieldBytesEncoding};
+
+use crate::{constants::BASEPOINT_ORDER_PRIVATE, edwards::CompressedEdwardsY, EdwardsPoint, Scalar};
+
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, PartialOrd, Ord, Hash)]
+pub struct Ed25519;
+
+impl Curve for Ed25519 {
+    type FieldBytesSize = U32;
+
+    type Uint = U256;
+
+    const ORDER: Self::Uint = U256::from_le_slice(&BASEPOINT_ORDER_PRIVATE.bytes);
+}
+
+impl CurveArithmetic for Ed25519 {
+    type AffinePoint = CompressedEdwardsY;
+
+    type ProjectivePoint = EdwardsPoint;
+
+    type Scalar = Scalar;
+}
+
+impl FieldBytesEncoding<Ed25519> for U256 {}
--- a/curve25519-dalek/src/ed25519.rs
+++ b/curve25519-dalek/src/ed25519.rs
@@ -2,6 +2,8 @@ use elliptic_curve::{bigint::U256, consts::U32, Curve, CurveArithmetic, FieldByt
 
 use crate::{constants::BASEPOINT_ORDER_PRIVATE, edwards::CompressedEdwardsY, EdwardsPoint, Scalar};
 
+/// QUESTION: I don't know where to put this singleton. Maybe in the crate's root?
+/// Otherwise, I thought of something like "singleton" or "curve".
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq, PartialOrd, Ord, Hash)]
 pub struct Ed25519;
 
--- a/curve25519-dalek/Cargo.toml
+++ b/curve25519-dalek/Cargo.toml
@@ -54,6 +54,7 @@ digest = { version = "0.10", default-features = false, optional = true }
 subtle = { version = "2.6.0", default-features = false, features = ["const-generics"]}
 serde = { version = "1.0", default-features = false, optional = true, features = ["derive"] }
 zeroize = { version = "1", default-features = false, optional = true }
+elliptic-curve = { version = "0.13", optional = true }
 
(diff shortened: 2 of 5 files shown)


tarcieri:
@carloskiki there will hopefully be a v0.14 release of `elliptic-curve` out soon along with corresponding v0.14 releases of `ff`/`group`. It would be good to target those (at least the prereleases currently available), though that may need other upgrades elsewhere in the project.

> The `CurveArithmetic` trait requires an `AffinePoint` for which this crate only has `CompressedEdwardsY`

`CompressedEdwardsY` is an encoding which must first be decompressed before you have an affine representation. It's more analogous to the SEC1 encoding which is the main one supported by `elliptic-curve` currently.

> We introduce a new pub AffineEdwardsPoint { x: FieldElement, y: FieldElement } (with private internals) to implement Curve & CurveArithmetic. I believe this would not be a breaking change.

This would probably make the most sense as a way to support the current `elliptic-curve` API. It's already computed inside of e.g. `EdwardsPoint::compress`, though it's less helpful in the point decompression case.

Perhaps the `elliptic-curve` API could be better generalized so as not to expose an `AffinePoint`/`ProjectivePoint` distinction, which isn't particularly helpful with Curve25519, and instead just having a single type for elements of the curve group. In the https://github.com/RustCrypto/elliptic-curves crates there could potentially instead be a `Point` enum that picks the appropriate formula based on the given variant which would also simplify the API to end users, but might require a somewhat difficult refactoring. We're making breaking changes now though, so it's something to consider.

curve25519-dalek/src/ed25519.rs
```diff
@@ -0,0 +1,25 @@
+use elliptic_curve::{bigint::U256, consts::U32, Curve, CurveArithmetic, FieldBytesEncoding};
+
+use crate::{constants::BASEPOINT_ORDER_PRIVATE, edwards::affine::AffinePoint, EdwardsPoint, Scalar};
+
+/// QUESTION: I don't know where to put this singleton. Maybe in the crate's root?
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, PartialOrd, Ord, Hash)]
+pub struct Ed25519;
```
tarcieri, line 7: Ed25519 is the name of the EdDSA variant.

RFC7748 calls the Edwards form of the curve "edwards25519"

carloskiki:
I opened https://github.com/RustCrypto/traits/issues/1900 to see if there would be interest in using a single `Point` type instead of the affine and projective variants. Since elliptic-curve v0.14.0 is already in release candidate phase this may have to wait to v0.15.0 if we decide this is a good idea. I will be drafting some traits to see how it would work soon.

In the mean time we may want to wait to implement these traits, especially for `curve25519`, since none of the required mathematical operations for `montgomery::ProjectivePoint` are implemented. I hardly see a use in implementing those other than to satisfy the trait bounds.

curve25519-dalek/src/ed25519.rs
```diff
@@ -0,0 +1,25 @@
+use elliptic_curve::{bigint::U256, consts::U32, Curve, CurveArithmetic, FieldBytesEncoding};
+
+use crate::{constants::BASEPOINT_ORDER_PRIVATE, edwards::affine::AffinePoint, EdwardsPoint, Scalar};
+
+/// QUESTION: I don't know where to put this singleton. Maybe in the crate's root?
```
tarcieri, line 5: I'd probably put it in `edwards.rs`

tarcieri:
@carloskiki I think it's fine to impl them just for `EdwardsPoint` and `RistrettoPoint` for now

carloskiki:
Closing this for now. I may try again after the new stable releases are out.

tarcieri:
Yeah, would definitely be good to target v0.14 after it's been released

PR #747, opened 2025-04-29
Implement `Zeroize` for assorted types

daxpedda:
This PR implements `Zeroize` for `ExpandedSecretKey`, `SigningKey` and `VerifiyingKey`.

I initially implemented them as zeroizing to scalar `1` or the base point, but noticed that these types don't uphold these invariants themselves.

### DIFF
--- a/ed25519-dalek/src/hazmat.rs
+++ b/ed25519-dalek/src/hazmat.rs
@@ -42,10 +42,17 @@ pub struct ExpandedSecretKey {
 }
 
 #[cfg(feature = "zeroize")]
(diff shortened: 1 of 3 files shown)


tarcieri:
Having them exclusively zeroize-on-drop as opposed to impl'ing `Zeroize` avoids having to deal with invalid states and also prevents an insecure "use-after-zeroize" condition

rozbb:
Never thought about that, but I agree with the reasoning here. I'll close for now. Thank you regardless!

Issue #743, opened 2025-04-24
some recent change broke Fiat Cryptography's CI


JasonGross:
Any idea what's going on with https://github.com/mit-plv/fiat-crypto/actions/runs/14630956505/job/41053013685?pr=2075#step:5:437
```
running 7 tests
test curve25519-dalek-derive/src/lib.rs - (line 41) ... ok
test curve25519-dalek-derive/src/lib.rs - (line 121) ... ok
test curve25519-dalek-derive/src/lib.rs - (line 108) - compile fail ... ok
test curve25519-dalek-derive/src/lib.rs - (line 140) ... ok
test curve25519-dalek-derive/src/lib.rs - (line 90) ... ok
test curve25519-dalek-derive/src/lib.rs - (line 97) ... ok
test curve25519-dalek-derive/src/lib.rs - (line 84) - compile fail ... FAILED

failures:

---- curve25519-dalek-derive/src/lib.rs - (line 84) stdout ----
Test compiled successfully, but it's marked `compile_fail`.

failures:
    curve25519-dalek-derive/src/lib.rs - (line 84)

test result: FAILED. 6 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.37s

error: doctest failed, to rerun pass `-p curve25519-dalek-derive --doc`
+ exit 101
```

tarcieri:
Can you post a compiler error message for why compilation failed?

JasonGross:
How do I get a compiler error message from `cargo test`?

<details><summary>full log</summary>

```
+ RUSTFLAGS='--cfg curve25519_dalek_backend="fiat" --cfg curve25519_dalek_bits="64"'
+ cargo test
    Updating crates.io index
     Locking 125 packages to latest compatible versions
      Adding bincode v1.3.3 (available: v2.0.1)
      Adding hex-literal v0.4.1 (available: v1.0.0)
      Adding rand v0.8.5 (available: v0.9.1)
      Adding rand_core v0.6.4 (available: v0.9.3)
      Adding toml v0.7.8 (available: v0.8.20)
 Downloading crates ...
  Downloaded toml v0.7.8
  Downloaded version_check v0.9.5
  Downloaded clap_lex v0.7.4
  Downloaded cpufeatures v0.2.17
  Downloaded autocfg v1.4.0
  Downloaded crossbeam-epoch v0.9.18
  Downloaded ciborium v0.2.2
[...]

PR #751, opened 2025-05-07
nightly and stable disagree on usage of unsafe

baloo:
```
error: unnecessary `unsafe` block
492
   --> curve25519-dalek/src/backend/vector/avx2/field.rs:479:28
493
    |
494
479 |         let c9_19: u32x8 = unsafe {
495
    |                            ^^^^^^ unnecessary `unsafe` block
```

This would only happen on nightly.


tarcieri:
This seems like a good time to take care of #743 too

baloo:
Clippy CI should be fixed by #752, let me know if you want me to merge those two.

tarcieri:
One PR to get the build green again would be good

Issue #705, opened 2024-09-23
Help out for batched random key generation?


stevefan1999-personal:
I'm trying to generate vanity wallet addresses, which are originally in ed25519, so I start by generating random bytes from a PRNG enough to cast into a secret key and then generate key pairs, but I ran into performance issue so I want to debug it:
```rust
for _ in 0..cli.threads {
    let thr = {
        let counter = counter.clone();
        move || {
            let mut csprng = fastrand::Rng::new();
            let mut pair = SecretKey::default();

            loop {
                let key: SigningKey = SigningKey::from_bytes(&pair);
                counter.fetch_add(1, Ordering::Relaxed);
            }
        }
    };

    thread::spawn(thr);
}
```
On my 3700X PC, and running in full 16 threads, it only runs 300K keypairs/second with a **precise** monotonic clock. However, the C vanity address generator on my same PC, with same parameters can generate 30M keypairs/second with random data, so I think there could be something wrong here

So I started digging, which means we start from here:
https://github.com/dalek-cryptography/curve25519-dalek/blob/d5ef57a3c2c4570720cafe8a112f5dd3ff56905e/ed25519-dalek/src/signing.rs#L102-L108

And then here:
https://github.com/dalek-cryptography/curve25519-dalek/blob/d5ef57a3c2c4570720cafe8a112f5dd3ff56905e/ed25519-dalek/src/signing.rs#L788-L794

And then here:
https://github.com/dalek-cryptography/curve25519-dalek/blob/d5ef57a3c2c4570720cafe8a112f5dd3ff56905e/ed25519-dalek/src/hazmat.rs#L57-L76

And then here:
https://github.com/dalek-cryptography/curve25519-dalek/blob/d5ef57a3c2c4570720cafe8a112f5dd3ff56905e/curve25519-dalek/src/scalar.rs#L237-L246

And eventually here:
https://github.com/dalek-cryptography/curve25519-dalek/blob/d5ef57a3c2c4570720cafe8a112f5dd3ff56905e/curve25519-dalek/src/scalar.rs#L1127-L1128

I noticed that despite I'm using AVX512, the flame graph shows that those `UnpackedScalar::mul_internal` and `UnpackedScalar::montgomery_reduce` are still using generic serial implementation and took a large group of execution time, I'm not sure if this can be sped-up and improved? Or maybe I should look for something else for address generation?

stevefan1999-personal:
[...]

PR #759, opened 2025-05-26
curve: add `EdwardsPoint::compress_batch` and inherent `::random`

tarcieri:
We've had various requests to implement batch point compression for `EdwardsPoint`, e.g. #705.

We can leverage `FieldElement::batch_invert` to implement it, which results in a fairly significant speedup.

The name `EdwardsPoint::compress_batch` has been chosen to match `RistrettoPoint::double_and_compress_batch`.

For benchmarking, randomized `EdwardsPoint`s have been used. To obtain these, an inherent `EdwardsPoint::random` has been extracted from the existing `Group::random` implementation, which uses rejection sampling. `Group::random` has been updated to call the inherent `EdwardsPoint::random`. This avoids a `group` dependency just to run the batch compression benchmarks.

The following benchmark results have been obtained:

```
edwards benches/EdwardsPoint compression
                        time:   [3.5029 µs 3.5098 µs 3.5171 µs]

edwards benches/Batch EdwardsPoint compression/1
                        time:   [3.6698 µs 3.6758 µs 3.6817 µs]
edwards benches/Batch EdwardsPoint compression/2
                        time:   [3.8410 µs 3.8461 µs 3.8516 µs]
edwards benches/Batch EdwardsPoint compression/4
                        time:   [4.1534 µs 4.1961 µs 4.2558 µs]
edwards benches/Batch EdwardsPoint compression/8
                        time:   [4.8466 µs 4.8533 µs 4.8600 µs]
edwards benches/Batch EdwardsPoint compression/16
                        time:   [6.1216 µs 6.1315 µs 6.1410 µs]
```

As you can see, it affords a fairly significant speedup, batch compressing 16 points in less time than the standard point compression algorithm would take to compress 2 in a row.

### DIFF
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -1342,6 +1342,7 @@ impl group::Group for EdwardsPoint {
     type Scalar = Scalar;
 
     fn random(rng: impl RngCore) -> Self {
+        // Call the inherent `pub fn random` defined above
         Self::random(rng)
     }
 
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -111,7 +111,7 @@ use {
     subtle::CtOption,
 };
 
-#[cfg(feature = "rand_core")]
+#[cfg(any(test, feature = "rand_core"))]
 use rand_core::RngCore;
 
 use subtle::Choice;
@@ -643,7 +643,7 @@ impl EdwardsPoint {
     ///
     /// Uses rejection sampling, generating a random `CompressedEdwardsY` and then attempting point
     /// decompression, rejecting invalid points.
-    #[cfg(feature = "rand_core")]
+    #[cfg(any(test, feature = "rand_core"))]
     pub fn random(mut rng: impl RngCore) -> Self {
         let mut repr = CompressedEdwardsY([0u8; 32]);
         loop {
@@ -2072,12 +2072,17 @@ mod test {
     #[cfg(feature = "alloc")]
     #[test]
     fn compress_batch() {
+        let mut rng = rand::thread_rng();
+
         // TODO(tarcieri): proptests?
-        let points = (1u64..16)
+        // Make some points deterministically then randomly
+        let mut points = (1u64..16)
             .map(|n| constants::ED25519_BASEPOINT_POINT * Scalar::from(n))
             .collect::<Vec<_>>();
+        points.extend(core::iter::repeat_with(|| EdwardsPoint::random(&mut rng)).take(100));
         let compressed = EdwardsPoint::compress_batch(&points);
 
+        // Check that the batch-compressed points match the individually compressed ones
         for (point, compressed) in points.iter().zip(&compressed) {
             assert_eq!(&point.compress(), compressed);
         }
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -585,8 +585,8 @@ impl EdwardsPoint {
             .iter()
             .zip(&zs)
             .map(|(input, recip)| {
-                let x = &input.X * &recip;
-                let y = &input.Y * &recip;
+                let x = &input.X * recip;
+                let y = &input.Y * recip;
 
                 let mut s = y.as_bytes();
                 s[31] ^= x.is_negative().unwrap_u8() << 7;
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -568,14 +568,11 @@ impl EdwardsPoint {
         let recip = self.Z.invert();
         let x = &self.X * &recip;
         let y = &self.Y * &recip;
-        let mut s: [u8; 32];
-
-        s = y.as_bytes();
-        s[31] ^= x.is_negative().unwrap_u8() << 7;
-        CompressedEdwardsY(s)
+        Self::compress_affine(x, y)
     }
 
-    /// Compress several `EdwardsPoint`s into `CompressedEdwardsY`.
+    /// Compress several `EdwardsPoint`s into `CompressedEdwardsY` format, using a batch inversion
+    /// for a significant speedup.
     #[cfg(feature = "alloc")]
     pub fn compress_batch(inputs: &[EdwardsPoint]) -> Vec<CompressedEdwardsY> {
         let mut zs = inputs.iter().map(|input| input.Z).collect::<Vec<_>>();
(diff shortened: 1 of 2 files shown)


curve25519-dalek/src/edwards.rs
```diff
@@ -574,6 +575,26 @@ impl EdwardsPoint {
...
+        inputs
+            .iter()
+            .zip(&zs)
+            .map(|(input, recip)| {
+                let x = &input.X * recip;
+                let y = &input.Y * recip;
+
+                let mut s = y.as_bytes();
+                s[31] ^= x.is_negative().unwrap_u8() << 7;
+                CompressedEdwardsY(s)
```
tarcieri, line 593: This much could potentially be DRYed out with `EdwardsPoint::compress`

curve25519-dalek/src/edwards.rs
```diff
@@ -574,6 +575,26 @@ impl EdwardsPoint {
         CompressedEdwardsY(s)
     }
 
+    /// Compress several `EdwardsPoint`s into `CompressedEdwardsY`.
+    #[cfg(feature = "alloc")]
+    pub fn compress_batch(inputs: &[EdwardsPoint]) -> Vec<CompressedEdwardsY> {
```
tarcieri, line 580: Note: this and `FieldElement::batch_invert` could avoid the `alloc` requirement by using const generic-sized arrays, which allows storing intermediates in a same-sized array rather than a `Vec`.

In the @RustCrypto crates we support both a no-alloc const generic and alloc-backed slice-based inputs for these sorts of APIs.

curve25519-dalek/src/edwards.rs
tarcieri: Done in 5263f7f

curve25519-dalek/src/edwards.rs
rozbb: Agreed we could probably very easily provide this (assuming we're okay exposing a generic array crate, which I don't think we're doing now)

curve25519-dalek/src/edwards.rs
```diff
@@ -567,9 +568,31 @@ impl EdwardsPoint {
...
+        Self::compress_affine(x, y)
+    }
 
-        s = y.as_bytes();
+    /// Compress several `EdwardsPoint`s into `CompressedEdwardsY` format, using a batch inversion
+    /// for a significant speedup.
+    #[cfg(feature = "alloc")]
+    pub fn compress_batch(inputs: &[EdwardsPoint]) -> Vec<CompressedEdwardsY> {
+        let mut zs = inputs.iter().map(|input| input.Z).collect::<Vec<_>>();
+        FieldElement::batch_invert(&mut zs);
```
rozbb, line 579: Question, more for this entire file: is `Z` ever zero? Why/why not? If it is zero, won't some of these get mapped to `(0, 0)`, which is the 2-torsion point, not the identity?

curve25519-dalek/src/edwards.rs
```diff
@@ -2019,6 +2059,26 @@ mod test {
...
+        {
+            let compressed = EdwardsPoint::compress_batch(&[EdwardsPoint::identity()]);
+            assert_eq!(&compressed, &[CompressedEdwardsY::identity()]);
+        }
+    }
+
+    #[cfg(feature = "alloc")]
+    #[test]
+    fn compress_batch() {
+        // TODO(tarcieri): proptests?
```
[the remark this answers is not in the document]
rozbb: I'm surprised I've never heard of proprtest until now. Looks very promising

curve25519-dalek/src/edwards.rs
```diff
@@ -1291,16 +1341,8 @@ impl Debug for EdwardsPoint {
...
-        loop {
-            rng.fill_bytes(&mut repr.0);
-            if let Some(p) = repr.decompress() {
-                if !IsIdentity::is_identity(&p) {
-                    break p;
-                }
-            }
-        }
+    fn random(rng: impl RngCore) -> Self {
+        Self::random(rng)
```
rozbb, line 1345: Note to self: had to check this doesn't cause infinite recursion. It doesn't.

rozbb (commented):
Looks good! Just 1 question in there. I'm gonna push some changes as well, since I noticed we can use random points in our tests

curve25519-dalek/src/edwards.rs
```diff
@@ -112,7 +111,7 @@ use {
     subtle::CtOption,
 };
 
-#[cfg(feature = "group")]
+#[cfg(any(test, feature = "rand_core"))]
```
rozbb, line 114: This matches how we use `rand_core` in [`scalar.rs`](https://github.com/dalek-cryptography/curve25519-dalek/blob/e3b5328202f4fae269daa88553bea7b4d3f86346/curve25519-dalek/src/scalar.rs#L574)

curve25519-dalek/src/edwards.rs
tarcieri: You don't need a generic array crate. It works fine with existing const generics, e.g.:

```rust
fn batch_invert<const N: usize>(elems: &mut [Self; N]) {
    let scratch = [Self::ONE; N];
    ...
}
```

Functions like that can infer the stack size needed for temporaries based on the size of the input, avoiding a heap allocation

curve25519-dalek/src/edwards.rs
tarcieri: I believe it's not allowed to ever be zero. It's `1` for the additive identity (X:0, Y:1, Z:1, T:0).

https://en.wikipedia.org/wiki/Homogeneous_coordinates suggests `Z` may be `0` for the point at infinity / additive identity, but that isn't the case here.

curve25519-dalek/src/edwards.rs
tarcieri: Yeah, it's a little confusing calling a same-named inherent method from a trait method, but Rust prefers the inherent method to recursing

tarcieri: Added a comment in [18cd950](https://github.com/dalek-cryptography/curve25519-dalek/pull/759/commits/18cd95024da1e507ce06f6af7f76071f9174753f)

curve25519-dalek/src/edwards.rs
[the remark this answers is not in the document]
tarcieri: Yeah, I can write some basic proptests based on some similar ones in the @RustCrypto elliptic curve crates

curve25519-dalek/src/edwards.rs
tarcieri: Even if it's identity, `Z` will still be `FieldElement::ONE`.

There's also a test for identity here: https://github.com/dalek-cryptography/curve25519-dalek/pull/759/files#diff-8dd3d09d222437a2e1efde6f6ea33a387ba97798d8dd7db34ee548e33a79587eR2068-R2069

PR #761, opened 2025-05-28
Don't implicitly enable zeroize for ed25519-dalek alloc feature

stevefan1999-personal:
I tried to compile `ed25519-dalek` using the new https://github.com/Rust-GPU/Rust-CUDA reboot and they mentioned they fixed a lot of issue, and I tried to include curve25519-dalek and it does miraculously compile, except it has to be without `zeroize` which is complaining about atomic fence not supported on PTX, and I have to use the equivalent intrinsic in cuda_std instead. 

As a quick demo for an internal finite field arithmetic + SIMD project and I rather need performance, I don't need zeroize at all. So a quick solution is to just disable it.

Anyway, I noticed with ed25519-dalek, when you enable the alloc feature, zeroize is implicitly enabled. This PR just add a single character to fix that

### DIFF
--- a/ed25519-dalek/Cargo.toml
+++ b/ed25519-dalek/Cargo.toml
@@ -60,7 +60,7 @@ required-features = ["rand_core"]
 
 [features]
 default = ["fast", "std", "zeroize"]
-alloc = ["curve25519-dalek/alloc", "ed25519/alloc", "serde?/alloc", "zeroize/alloc"]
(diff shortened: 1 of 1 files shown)


rozbb:
@tarcieri is this technically breaking? It could make existing code cease to zeroize structs

tarcieri:
@rozbb it might technically be considered breaking, yes. Would be safer to do at a major version boundary

PR #763, opened 2025-06-01
Update to `&[&[u8]]` messages

daxpedda:
Just showing off https://github.com/RustCrypto/traits/pull/1880. The current plan is to move this to a separate trait, which I'm, as of yet, unsure how it will integrate into `ed25519-dalek`.


daxpedda:
So the new trait, `MultiPartSigner` and `MultiPartVerifier` (see https://github.com/RustCrypto/traits/pull/1880), would still require all internal functions to take a `&[&[u8]]` message.

In any case, I think this PR showed that doing that shouldn't be a problem. I will go ahead and re-open this when we are ready to upgrade to the new RustCrypto versions and after #735 is in. Maybe `MultiPartSigner/Verifier` will be built on top of the streaming API.

tarcieri:
@daxpedda you should be able to prototype on top of #676

PR #765, opened 2025-06-03
ed: Expose `hazmat::raw_sign_byupdate()` for streamed signing

rozbb:
Currently, if you want to sign a serialized object, you have to serialize it in full, then sign it. When memory is constrained, this is hard.

This PR defines a function, `hazmat::raw_sign_byupdate`, which, rather than taking a single contiguous message slice, takes a `Fn(&mut Digest)` that knows how to update a hasher with the parts of the message. So, in the case of serialization, the closure will simply serialize each component to a small buffer, do `h.update(buf)`, then fill the buffer with the next serialized component, etc.

This comes from discussion [here](https://github.com/dalek-cryptography/curve25519-dalek/pull/735#discussion_r2119269994), [here](https://github.com/dalek-cryptography/curve25519-dalek/pull/556#issuecomment-1707656494), and hypothetical impl [here](https://github.com/mkj/sunset/blob/3c7d6879da93684832f8ec19df4191523999a2f3/src/sign.rs#L324-L338).

This is feature-gated behind `hazmat` because it takes an expanded secret key, which is possible to catastrophically misuse.

@mkj

### DIFF
--- a/ed25519-dalek/src/hazmat.rs
+++ b/ed25519-dalek/src/hazmat.rs
@@ -169,6 +169,32 @@ where
     esk.raw_sign_prehashed::<CtxDigest, MsgDigest>(prehashed_message, verifying_key, context)
 }
 
+/// Compute an ordinary Ed25519 signature, with the message contents provided incrementally by
+/// updating a digest instance.
+///
+/// The `msg_update` closure provides the message content, updating a hasher argument. It will be
+/// called twice.
+///
+/// `CtxDigest` is the digest used to calculate the pseudorandomness needed for signing. According
+/// to the Ed25519 spec, `CtxDigest = Sha512`.
+///
+/// # ⚠️  Unsafe
+///
+/// Do NOT use this function unless you absolutely must. Using the wrong values in
+/// `ExpandedSecretKey` can leak your signing key. See
+/// [here](https://github.com/MystenLabs/ed25519-unsafe-libs) for more details on this attack.
+pub fn raw_sign_byupdate<CtxDigest, F>(
+    esk: &ExpandedSecretKey,
+    msg_update: F,
+    verifying_key: &VerifyingKey,
+) -> Result<Signature, SignatureError>
+where
+    CtxDigest: Digest<OutputSize = U64>,
+    F: Fn(&mut CtxDigest) -> Result<(), SignatureError>,
+{
+    esk.raw_sign_byupdate::<CtxDigest, F>(msg_update, verifying_key)
+}
+
 /// The ordinary non-batched Ed25519 verification check, rejecting non-canonical R
 /// values.`CtxDigest` is the digest used to calculate the pseudorandomness needed for signing.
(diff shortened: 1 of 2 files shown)


rozbb:
I don't see any performance degredation

mkj:
Thanks. I've tested this PR `raw_sign_byupdate()` works in my codebase, ~~can the `raw_verify_byupdate()` also be added?~~ Edit: not thinking - I should just be able to use `StreamVerifier`.

mkj:
I guess `raw_verify_byupdate()` doesn't have any use, I _think_ `StreamVerifier` covers all those cases.

mkj:
Tested `StreamVerifier` works well for my purposes here. Thanks!

ed25519-dalek/src/signing.rs
```diff
@@ -841,24 +842,48 @@ impl ExpandedSecretKey {
...
+                h.update(message);
+                Ok(())
+            },
+            verifying_key,
+        )
+        .unwrap()
+    }
+
+    /// Sign a message provided in parts. The `msg_update` closure
+    /// will be called twice to hash the message parts.
```
tarcieri: The closure being called twice seems a little awkward, and it's unclear from this documentation each invocation needs to act deterministically compared to the previous ones. Someone not paying attention to the documentation might assume they're free to update the digest every time the callback is invoked.

tarcieri:
I think an `&[&[u8]]` API is a lot more foolproof, though this is more flexible

ed25519-dalek/src/signing.rs
rozbb: Agreed, just added some docs to explain what's needed.

ed25519-dalek/src/hazmat.rs
```diff
@@ -169,6 +169,32 @@ where
...
+/// Compute an ordinary Ed25519 signature, with the message contents provided incrementally by
+/// updating a digest instance.
+///
+/// The `msg_update` closure provides the message content, updating a hasher argument. It will be
+/// called twice.
+///
+/// `CtxDigest` is the digest used to calculate the pseudorandomness needed for signing. According
+/// to the Ed25519 spec, `CtxDigest = Sha512`.
+///
+/// # ⚠️  Unsafe
```
tarcieri: The word "unsafe" is pretty overloaded in Rust:

```suggestion
/// # ⚠️  Cryptographically Unsafe
```

tarcieri: Also seems like this needs similar docs to these regarding each invocation of the callback being deterministic: https://github.com/dalek-cryptography/curve25519-dalek/pull/765#pullrequestreview-2890403083

rozbb: "Unsafe" is used a lot in the hazmat file. I agree tho it's overloaded. I'll change it everywhere then

rozbb: Done

PR #766, opened 2025-06-04
Add `typos` action and correct typos

tarcieri:
https://github.com/crate-ci/typos

### DIFF
--- a/ed25519-dalek/src/signature.rs
+++ b/ed25519-dalek/src/signature.rs
@@ -73,7 +73,7 @@ fn check_scalar(bytes: [u8; 32]) -> Result<Scalar, SignatureError> {
     // potential non-reduced scalars is performed.
     //
     // This is compatible with ed25519-donna and libsodium when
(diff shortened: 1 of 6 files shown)


ed25519-dalek/src/signature.rs
```diff
@@ -73,7 +73,7 @@ fn check_scalar(bytes: [u8; 32]) -> Result<Scalar, SignatureError> {
     // potential non-reduced scalars is performed.
     //
     // This is compatible with ed25519-donna and libsodium when
-    // -DED25519_COMPAT is NOT specified.
+    // `-D ED25519_COMPAT` is NOT specified.
```
tarcieri, line 76: Couldn't figure out another way this was happy with but we could play with the config. I think it's equivalent?

rozbb: Yeah I think it's fine, and it doesn't have to be syntactically accurate, just legible

PR #767, opened 2025-06-06
curve: rename `FieldElement*::as_bytes` => `::to_bytes`

tarcieri:
Methods named `as_*` should perform a zero-cost  borrowing conversion:

https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

Methods named `to_*` can perform an expensive owned conversion.

Since the `FieldElement*` types are technically part of the public API (but feature gated), this also preserves the old names with a deprecation. We can remove them in the next breaking release.

The same change was also made to the backend `Scalar*` types, however these types are not a part of the public API.

### DIFF
--- a/curve25519-dalek-derive/src/lib.rs
+++ b/curve25519-dalek-derive/src/lib.rs
@@ -203,10 +203,7 @@ fn process_mod(
                                         };
 
                                         let feature = feature.value();
-                                        if !spec_features
-                                            .iter()
-                                            .any(|enabled_feature| feature == *enabled_feature)
-                                        {
+                                        if !spec_features.contains(&feature) {
                                             *item = syn::Item::Verbatim(Default::default());
                                             continue 'next_item;
                                         }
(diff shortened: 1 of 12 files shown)


tarcieri:
This is odd: https://github.com/dalek-cryptography/curve25519-dalek/actions/runs/15495347219/job/43630615714?pr=767#step:4:104

```
warning: using `contains()` instead of `iter().any()` is more efficient
   --> curve25519-dalek-derive/src/lib.rs:206:45
    |
206 |   ...                   if !spec_features
    |  ___________________________^
207 | | ...                       .iter()
208 | | ...                       .any(|enabled_feature| feature == *enabled_feature)
    | |_____________________________________________________________________________^ help: try: `spec_features.contains(&feature)`
```

Seems completely unrelated to this PR, but I'm going to fix it anyway.

rozbb (approved):

tarcieri:
Not seeing any performance regressions with this, so I'm going to go ahead and merge.

@rozbb this change isn't breaking because it preserves but deprecates the old method names. However, before we cut the next breaking release we should remove those deprecated methods.

PR #768, opened 2025-06-06
CI: bump `clippy` to Rust 1.87

tarcieri:
Performs a `cargo clippy --fix`


curve25519-dalek/src/backend/serial/curve_models/mod.rs
```diff
@@ -408,7 +408,7 @@ impl ProjectivePoint {
 //
 // upstream rust issue: https://github.com/rust-lang/rust/issues/46380
 //#[doc(hidden)]
-impl<'a, 'b> Add<&'b ProjectiveNielsPoint> for &'a EdwardsPoint {
+impl<'b> Add<&'b ProjectiveNielsPoint> for &EdwardsPoint {
```
tarcieri: The lifetime should probably be renamed to `'a` in these cases. Will do

PR #769, opened 2025-06-06
curve: extract `AffinePoint` type

tarcieri:
Based on discussions about `elliptic-curve` trait impls in #746, and observing a similar type in `ed448-goldilocks` which inspired this one (not to mention in all of the @RustCrypto elliptic curve crates), adds an `AffinePoint` type with `x` and `y` coordinates.

For now, the type is kept out of the public API, and used as an implementation detail for point compression. However, it's been written with the intent of eventually stabilizing and exposing it. It's been marked `pub` so unused functionality doesn't automatically trigger dead code lints.

Further work could include refactoring point decompression to first produce an `AffinePoint` and then convert to extended twisted Edwards coordinates (i.e. `EdwardsPoint`), which is more or less what the existing `step_1` and `step_2` functions do (`step_1` technically produces projective coordinates, but `Z` is always set to `ONE`).


curve25519-dalek/src/edwards.rs
```diff
@@ -542,6 +546,14 @@ impl EdwardsPoint {
         }
     }
 
+    /// Convert to `AffinePoint`.
+    pub(crate) fn to_affine(self) -> AffinePoint {
```
rozbb, line 550: This can't possibly be a total function, right? The identity point doesn't have an affine representation.

tarcieri: My understanding is that unlike e.g. Weierstrass (where in @RustCrypto we effectively turn every `AffinePoint` into a `CtOption` to be able to represent the point at infinity), in Edwards curves the identity actually has affine coordinates, namely `(0,1)`. See also `impl Identity for AffinePoint` below. This also avoids the case `Z=0` we discussed earlier.

curve25519-dalek/src/edwards.rs
```diff
@@ -542,6 +546,14 @@ impl EdwardsPoint {
         }
     }
 
+    /// Convert to `AffinePoint`.
```
tarcieri: I noticed several of the other comments to this effect use this terminology:

```suggestion
    /// Dehomogenize to `AffinePoint`.
```

curve25519-dalek/src/edwards.rs
rozbb: Ah, you're right, I completely forgot the distinction

kayabaNerve:
I was considering a PR to support deserialization of _uncompressed_ points, as I'm currently paying a _notable_ performance penalty reading points from a binary (generated at compile-time) for no reason when I found this PR. As a note, I'd love for a public AffinePoint I can load an x/y coordinate into and convert into a EdwardsPoint.

tarcieri:
@kayabaNerve you're fine with it still checking the coordinates are a valid solution to the curve equation, right? (it would still avoid the `sqrt` decompression requires)

I would be okay with a public checked constructor like that, when and if we do decide to make `AffinePoint` public.

kayabaNerve:
Of course. The few multiplications there are <5% of the sqrt though :sweat_smile:

tarcieri:
@kayabaNerve want to open a PR for discussion to export `AffinePoint` as `pub` and add a public constructor which takes the coordinates?

kayabaNerve:
Working on this after #787. It presumably will have to build upon that, unless we define the coordinates as `[u8; 32]`...

PR #774, opened 2025-06-20
ed25519: vendor `merlin` dependency

tarcieri:
`merlin` is currently a blocker for upgrading to `rand_core` v0.9 by way of the `transcript.build_rng().finalize()` function (which we only pass `ZeroRng` to).

There is an open PR to update `rand_core` in `merlin` and I have pinged the relevant people to take a look, hopefully: zkcrypto/merlin#11

However, in the event we can't get `merlin` updated, this at least unblocks the `rand_core` upgrade, and is being opened as a contingency plan for that case.

The PR has been implemented in a way that it should be easy to switch back to upstream `merlin` in the event they upgrade `rand_core`.

### DIFF
--- a/ed25519-dalek/Cargo.toml
+++ b/ed25519-dalek/Cargo.toml
@@ -33,7 +33,7 @@ sha2 = { version = "0.10", default-features = false }
 subtle = { version = "2.3.0", default-features = false }
 
 # optional features
-merlin = { version = "3", default-features = false, optional = true }
+keccak = { version = "0.1", default-features = false, optional = true }
 rand_core = { version = "0.6.4", default-features = false, optional = true }
 serde = { version = "1.0", default-features = false, optional = true }
 zeroize = { version = "1.5", default-features = false, optional = true }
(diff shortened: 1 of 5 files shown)


rozbb (approved):
Nice! I also don’t love it but I agree it’s the best way forward. Thanks for making this

burdges:
Around transcripts, Arkworks' ark_serialize uses rust's std::io interface adapted to be no-std, but not no-alloc, although rust should provide a no-alloc io interface someday.  All this they inherited from zcash.

Arkworks itself mostly focuses upon field element based transcript schemes for recursion, but if you want a top level transcript that likes this std::io interface then for domain seperation you want the lengths to be hashed postfix after the data, becaue you've no idea how io::Write nests its calls.  I wrote [ark_transcript](https://docs.rs/ark-transcript/latest/ark_transcript/) for this. 

All this is mostly useless to you guys, but if you ever get sick of merlin and do your own general thing then postfix lengths fit rust somewhat better.  I doubt a general transcript thing is required here of course.

PR #762, opened 2025-05-31
Bump rand_core to 0.9 + MSRV 1.85

baloo:
This is #749 and this also bumps the MSRV to 1.85 to fix CI.

PR #775, opened 2025-07-07
Bump edition to 2024 and crate versions to prereleases

tarcieri:
This represents the first breaking change in a new release series, bumping all crates to the 2024 edition of Rust.

As such, the version numbers of all crates have been incremented to represent a new prerelease series:

- `curve25519-dalek`: v5.0.0-pre
- `ed25519-dalek`: v3.0.0-pre
- `x25519-dalek`: v3.0.0-pre

Note that this commit isn't intended to cut associated crate releases of these on crates.io, but is merely bumping the version numbers to denote there are pending breaking changes.


tarcieri:
~~Huh, looks like a possible 2024 edition incompatibility with `zeroize_derive`. I've run into similar problems with other custom derive crates I maintain~~.

Perhaps I'll just get rid of the derive and write the impl manually.

Edit: aah, it's because of `-Z minimal-versions` and it's pulling in an old `zeroize_derive`. Let me bump the `zeroize` requirement and see if that fixes the issue.

rozbb (approved):
One Q: is the x25519 version bump just bc of the edition bump? Otherwise there’s nothing breaking right

tarcieri:
Technically the edition bump itself is not breaking per our SemVer policy, however subsequent changes we plan on making (e.g. #762) will be.

If you'd prefer I can extract the version bumps out into a separate PR.

rozbb:
I think it's fine, ty

PR #776, opened 2025-07-07
x25519: use manual impls instead of `zeroize_derive`

tarcieri:
The types involved are all simple 1-tuple newtypes where zeroization only involves calling `zeroize` on the inner type, making all of the involved impls relatively trivial.

Avoiding custom derive arguably improves auditability as you don't need to expand a proc macro to see the resulting code. It decreases the number of required dependencies in order for the `zeroize` feature to work, where some of those dependencies are incredibly heavy (particularly `syn`).

### DIFF
--- a/x25519-dalek/src/x25519.rs
+++ b/x25519-dalek/src/x25519.rs
@@ -29,7 +29,6 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 /// (in this crate) does *not* automatically happen, but either must be derived
 /// for Drop or explicitly called.
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[cfg_attr(feature = "zeroize", derive(Zeroize))]
 #[derive(PartialEq, Eq, Hash, Copy, Clone, Debug)]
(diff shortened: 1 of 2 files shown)


x25519-dalek/src/x25519.rs
```diff
@@ -111,6 +116,23 @@ impl<'a> From<&'a EphemeralSecret> for PublicKey {
     }
 }
 
+impl Drop for EphemeralSecret {
+    fn drop(&mut self) {
+        #[cfg(feature = "zeroize")]
+        self.zeroize();
+    }
+}
```
tarcieri, line 124: The reason for gating the inner `zeroize` invocation on the feature as opposed to the whole `Drop` impl is so enabling the `zeroize` feature doesn't change the `Drop` bounds / presence of drop glue. There are some weird edge cases around this approach avoids.

tarcieri:
Note: this is largely a workaround for the `x25519` build failure in #775, but I think the case for improved auditability and fewer dependencies is also valid.

I *think* this upstream change should take care of the problem going forward, as well: https://github.com/RustCrypto/utils/pull/1190

rozbb (approved):
LGTM thanks!

PR #749, opened 2025-05-07
chore(deps): bump rand_core to 0.9

baloo:
This includes https://github.com/dalek-cryptography/curve25519-dalek/pull/729

Depends:
 - #753 
 - #750 
 - https://github.com/RustCrypto/hashes/pull/678
 - https://github.com/RustCrypto/traits/pull/1799

PR #777, opened 2025-07-07
Bump rand_core to 0.9 

baloo:
Supersedes #676, #749, and #762


tarcieri:
@baloo one more rebase now that #676 is merged, sorry

rozbb:
Looks great, thank you so much! This seems like it was a pain to get all the API uses

Issue #760, opened 2025-05-27
ed25519-dalek: problematic `legacy_compatibility` feature


tarcieri:
Due to feature unification, one crate dependency activating it causes it to be active globally for all dependencies, allowing malleable signature verification everywhere in a program when maybe you just want it for something like VRFs or batch verification.

But it's similarly problematic in `curve25519-dalek` where one crate can use `Scalar::from_bits` and pass such a scalar to another crate's API that accepts `Scalar`s.

The feature does appear to have real-world users, in both `curve25519-dalek` and `ed25519-dalek`:

https://github.com/search?q=ed25519-dalek+legacy_compatibility+language%3Atoml&type=code

Some notable examples:
- https://github.com/facebook/akd/blob/c509e010fa10857f4e18cf01f4ec80b78fd102e7/akd_core/src/ecvrf/ecvrf_impl.rs#L309
- https://github.com/openmina/openmina/blob/6d88b70d68dbb439906004acd499a7d600d61487/p2p/src/network/noise/p2p_network_noise_state.rs#L599
- Possibly unnecessary? https://github.com/protecc-labs/smol-solana/blob/e46f9e838f63fa11e90c08c7d54f453716e2958e/Cargo.toml#L125

One option for `ed25519-dalek` could be to introduce a `MalleableSignature` type. To make this work, `ed25519-dalek` would still need the `curve25519-dalek`'s `legacy_compatibility` feature. The `MalleableSignature` type could still be gated under `legacy_compatibility`, but it would be purely additive the way features are supposed to work.

That's the only real suggestion I have. It sure would be nice to get rid of the feature(s), but they seem to have legitimate real-world users.

tarcieri:
We could also ask some of those downstream users how much they really need the feature

burdges:
It's not a big change for downstream users, just a `use ..::{.., MalleableSignature as Signature};`, but maybe turn the feature into an error with some explanation.  If you want a feature, then give it a new name like `legacy_malleable`.

Oh wait, you need maximum malleablity for batch verifications, no?

tarcieri:
@burdges the `batch` feature in `ed25519-dalek` works without `legacy_compatibility` enabled

PR #778, opened 2025-07-07
curve,x: Clear deprecated functions

rozbb:
One thing that isn't clear is how to deprecated `Scalar::from_bits`. Ed uses this public function in `legacy_compatability` mode. We can't remove it because it's a direct dependency. And we can't vendor it in Ed because Ed doesn't have access to the type's fields. Should this just be a perma-visible deprecation?

### DIFF
--- a/curve25519-dalek/src/constants.rs
+++ b/curve25519-dalek/src/constants.rs
@@ -69,10 +69,7 @@ pub const RISTRETTO_BASEPOINT_POINT: RistrettoPoint = RistrettoPoint(ED25519_BAS
 /// $$
 /// \ell = 2^\{252\} + 27742317777372353535851937790883648493.
 /// $$
-#[deprecated(since = "4.1.1", note = "Should not have been in public API")]
-pub const BASEPOINT_ORDER: Scalar = BASEPOINT_ORDER_PRIVATE;
-
-pub(crate) const BASEPOINT_ORDER_PRIVATE: Scalar = Scalar {
+pub(crate) const BASEPOINT_ORDER: Scalar = Scalar {
     bytes: [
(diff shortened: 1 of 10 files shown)


tarcieri:
I opened an issue about that, FWIW: #760 

I agree we can't remove it, and it should simply remain perma-deprecated.

tarcieri:
@rozbb looks like it needs a rebase

rozbb:
Just throwing out another option re from_bits: we could also undeprecate it and move it to hazmat. That way we don’t have the awkward situation of keeping a deprecated func around forever. Also hazmat is appropriate because it’s super not safe

tarcieri:
@rozbb that sounds fine to me, but perhaps in a separate PR to track it more easily?

rozbb:
Sounds good

curve25519-dalek/src/constants.rs
```diff
@@ -69,10 +69,7 @@ pub const RISTRETTO_BASEPOINT_POINT: RistrettoPoint = RistrettoPoint(ED25519_BAS
 /// $$
 /// \ell = 2^\{252\} + 27742317777372353535851937790883648493.
 /// $$
-#[deprecated(since = "4.1.1", note = "Should not have been in public API")]
-pub const BASEPOINT_ORDER: Scalar = BASEPOINT_ORDER_PRIVATE;
-
-pub(crate) const BASEPOINT_ORDER_PRIVATE: Scalar = Scalar {
+pub(crate) const BASEPOINT_ORDER: Scalar = Scalar {
```
tarcieri, line 72: This is sure nice to fix

tarcieri (approved):

PR #783, opened 2025-07-08
main: Update all READMEs and CHANGELOGs

rozbb:
Sorry this got a bit messy. This updates the READMEs and CHANGELOGs to include everything since v4.1.3 (copying stuff from the `curve-v4` branch). While I was at it, I noticed a few things:
1. ed doesn't have an `asm` feature anymore. Removed.
2. ed has a `fast` feature that wasn't documented. Added (will backport)
3. ed was not using `std` anywhere other than testing. Removed the feature.
4. curve has a `group-bits` feature that wasn't documented. Added (will backport)
5. ed's benches use `thread_rng()` which is deprecated. Fixed.

Lmk if I missed anything


curve25519-dalek/README.md
```diff
@@ -35,27 +35,28 @@ cofactor-related abstraction mismatches.
...
 ```toml
-curve25519-dalek = "4"
+curve25519-dalek = "5.0.0-pre"
 ```
 
 If opting into [SemVer-exempted features](#public-api-semver-exemptions) a range
 can be used to scope the tested compatible version range e.g.:
 ```toml
-curve25519-dalek = ">= 4.0, < 4.2"
+curve25519-dalek = ">= 5.0, < 5.2"
```
tarcieri, line 44: Do we actually want to update these yet? This is what people see when they go to the `main` branch of the repo

rozbb: Eh I think it’s fine. Comes with main being dev. I think there’s a decent chance that the few newcomers who discover this via our GitHub people will recognize that this is a prerelease. Wdyt?

tarcieri (approved):

PR #786, opened 2025-07-10
Add proper hash_to_curve. 

### LINKED ISSUE #785: curve25519: encode vs hash to curve mismatch.
In [Section 3 of RFC9380](https://www.rfc-editor.org/rfc/rfc9380.html#section-3) , there is a distinction between `hash_to_curve` and `encode_to_curve`. 

The [hash_to_curve](https://docs.rs/curve25519-dalek/4.2.0/curve25519_dalek/edwards/struct.EdwardsPoint.html#method.hash_to_curve) rust method actually implements `encode_to_curve`, so it is not safe to be used (and named) as a hash.

My recommendation is to define two functions: 
**encode_to_curve**: implementing the suite `edwards25519_XMD:SHA-512_ELL2_NU_`.
**hash_to curve**: implementing the suite `edwards25519_XMD:SHA-512_ELL2_RO_`.

cc: @rozbb

iquerejeta:
Sorry for the misunderstanding - it was a long time I did the original PR 🙏  Here is a PR with the proper hash_to_curve implementation. 

When implementing, I did a stupid mistake that got me debugging for a couple of hours, and just to make sure that the issue was not the elligator (which it was not) I implemented the inline description of the draft. I found the function `pow_p58`, which is the one I thought was missing originally. Let me know if you prefer this inline version or the one we had before. Both work (turned out that the mistake was me hardcoding incorrect test vectors 🥲 )

The other important mention is that I've used `unsafe` code to be able to not use `alloc` in `hash_to_field`. I was trying several ways, but couldn't get anything better. I'm sure you've found similar issues in the past and you have a preferred way to handle these scenarios - let me know what you prefer.  

I'll be off for a couple of days, so if this is a rush, feel free to take over the PR. 

This fixes #785.

### DIFF
--- a/curve25519-dalek/benches/dalek_benchmarks.rs
+++ b/curve25519-dalek/benches/dalek_benchmarks.rs
@@ -89,6 +89,21 @@ mod edwards_benches {
         );
     }
 
+    #[cfg(feature = "digest")]
+    fn hash_to_curve<M: Measurement>(c: &mut BenchmarkGroup<M>) {
+        let mut rng = rng();
+
+        let mut msg = [0u8; 32];
+        let mut domain_sep = [0u8; 32];
+        rng.fill_bytes(&mut msg);
+        rng.fill_bytes(&mut domain_sep);
+
+        c.bench_function(
+            "Elligator2 hash to curve (SHA-512, input size 32 bytes)",
+            |b| b.iter(|| EdwardsPoint::hash_to_curve::<Sha512>(&[&msg], &[&domain_sep])),
+        );
+    }
+
     pub(crate) fn edwards_benches() {
         let mut c = Criterion::default();
         let mut g = c.benchmark_group("edwards benches");
@@ -101,6 +116,7 @@ mod edwards_benches {
         consttime_variable_base_scalar_mul(&mut g);
         vartime_double_base_scalar_mul(&mut g);
         encode_to_curve(&mut g);
+        hash_to_curve(&mut g);
     }
 }
 
--- a/curve25519-dalek/benches/dalek_benchmarks.rs
+++ b/curve25519-dalek/benches/dalek_benchmarks.rs
@@ -84,7 +84,7 @@ mod edwards_benches {
         rng.fill_bytes(&mut domain_sep);
 
         c.bench_function(
-            "Elligator2 hash to curve (SHA-512, input size 32 bytes)",
+            "Elligator2 encode to curve (SHA-512, input size 32 bytes)",
             |b| b.iter(|| EdwardsPoint::encode_to_curve::<Sha512>(&[&msg], &[&domain_sep])),
         );
     }
--- a/curve25519-dalek/benches/dalek_benchmarks.rs
+++ b/curve25519-dalek/benches/dalek_benchmarks.rs
@@ -75,7 +75,7 @@ mod edwards_benches {
     }
 
     #[cfg(feature = "digest")]
-    fn hash_to_curve<M: Measurement>(c: &mut BenchmarkGroup<M>) {
+    fn encode_to_curve<M: Measurement>(c: &mut BenchmarkGroup<M>) {
         let mut rng = rng();
 
         let mut msg = [0u8; 32];
@@ -85,7 +85,7 @@ mod edwards_benches {
 
         c.bench_function(
             "Elligator2 hash to curve (SHA-512, input size 32 bytes)",
-            |b| b.iter(|| EdwardsPoint::hash_to_curve::<Sha512>(&[&msg], &[&domain_sep])),
+            |b| b.iter(|| EdwardsPoint::encode_to_curve::<Sha512>(&[&msg], &[&domain_sep])),
         );
     }
 
@@ -100,7 +100,7 @@ mod edwards_benches {
         consttime_fixed_base_scalar_mul(&mut g);
         consttime_variable_base_scalar_mul(&mut g);
         vartime_double_base_scalar_mul(&mut g);
-        hash_to_curve(&mut g);
+        encode_to_curve(&mut g);
     }
 }
 
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -674,11 +674,15 @@ impl EdwardsPoint {
     }
 
     #[cfg(feature = "digest")]
-    /// Perform encode to curve per RFC 9380, with explicit hash function and domain separator, `domain_sep`,
-    /// using the suite `edwards25519_XMD:SHA-512_ELL2_NU_`. The input is the concatenation of the
-    /// elements of `bytes`. Likewise for the domain separator with `domain_sep`. At least one
-    /// element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed
-    /// 255 bytes.
+    /// Perform encode to curve per RFC 9380, with explicit hash function and domain separator
+    /// `domain_sep`, using the Twisted Edwards Elligator 2 method. The input is the concatenation
+    /// of the elements of `bytes`. Likewise for the domain separator with `domain_sep`. At least
+    /// one element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed 255
+    /// bytes.
+    /// 
+    /// The specification names SHA-512 as an example of a secure hash to use with this function,
+    /// but you may use any 512-bit hash within reason (see the
+    /// [`spec`](https://www.rfc-editor.org/rfc/rfc9380.html#section-5.2) for details).
     ///
     /// # Warning
     /// `encode_to_curve` is a nonuniform encoding from byte strings to points in `G`. That is,
@@ -705,11 +709,15 @@ impl EdwardsPoint {
     }
 
     #[cfg(feature = "digest")]
-    /// Perform a hash to curve per RFC 9380, with explicit hash function and domain separator, `domain_sep`,
-    /// using the suite `edwards25519_XMD:SHA-512_ELL2_RO_`. The input is the concatenation of the
-    /// elements of `bytes`. Likewise for the domain separator with `domain_sep`. At least one
-    /// element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed
+    /// Perform a hash to curve per RFC 9380, with explicit hash function and domain separator
+    /// `domain_sep`, using the Twisted Edwards Elligator 2 method. The input is the concatenation
+    /// of the elements of `bytes`. Likewise for the domain separator with `domain_sep`. At least
+    /// one element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed
     /// 255 bytes.
+    ///
+    /// The specification names SHA-512 as an example of a secure hash to use with this function,
+    /// but you may use any 512-bit hash within reason (see the
+    /// [`spec`](https://www.rfc-editor.org/rfc/rfc9380.html#section-5.2) for details).
     ///
     /// # Panics
     /// Panics if `domain_sep.collect().len() == 0` or `> 255`
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -2472,6 +2472,7 @@ mod test {
         )
     ];
 
+    #[cfg(all(feature = "alloc", feature = "digest"))]
     fn hex_str_to_fe(hex_str: &str) -> FieldElement {
         let mut bytes = hex::decode(hex_str).unwrap().to_vec();
         bytes.reverse();
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -665,10 +665,12 @@ impl EdwardsPoint {
         let yd = FieldElement::conditional_select(&yd, &FieldElement::ONE, e);
         // 13. return (xn, xd, yn, yd)
 
-        let x = &xn * &xd.invert();
-        let y = &yn * &yd.invert();
-
-        AffinePoint { x, y }.to_edwards()
+        EdwardsPoint {
+            X: &xn * &yd,
+            Y: &xd * &yn,
+            Z: &xd * &yd,
+            T: &xn * &yn,
+        }
     }
 
     #[cfg(feature = "digest")]
@@ -2433,7 +2435,7 @@ mod test {
     }
 
     // Hash-to-curve test vectors from
-    // https://www.rfc-editor.org/rfc/rfc9380.html#name-edwards25519_xmdsha-512_ell2
+    // https://www.rfc-editor.org/rfc/rfc9380.html#appendix-J.5.2
     // These are of the form (input_msg, output_x, output_y)
     #[cfg(all(feature = "alloc", feature = "digest"))]
     const RFC_HASH_TO_CURVE_KAT_NU: &[(&[u8], &str, &str)] = &[
@@ -2472,7 +2474,7 @@ mod test {
 
     #[test]
     #[cfg(all(feature = "alloc", feature = "digest"))]
-    fn elligator_map_to_curve_test_vectors() {
+    fn elligator_encode_to_curve_test_vectors() {
         let dst = b"QUUX-V01-CS02-with-edwards25519_XMD:SHA-512_ELL2_NU_";
         for (index, vector) in RFC_HASH_TO_CURVE_KAT_NU.iter().enumerate() {
             let input = vector.0;
@@ -2500,7 +2502,7 @@ mod test {
     }
 
     // Hash-to-curve test vectors from
-    // https://www.rfc-editor.org/rfc/rfc9380.html#name-edwards25519_xmdsha-512_ell
+    // https://www.rfc-editor.org/rfc/rfc9380.html#appendix-J.5.1
     // These are of the form (input_msg, output_x, output_y, q0_x, q0_y, q1_x, q1_y, u0, u1)
     #[cfg(all(feature = "alloc", feature = "digest"))]
     const RFC_HASH_TO_CURVE_KAT_RO: &[(&[u8], &str, &str)] = &[
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -678,6 +678,15 @@ impl EdwardsPoint {
     /// element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed
     /// 255 bytes.
     ///
+    /// # Warning
+    /// `encode_to_curve` is a nonuniform encoding from byte strings to points in `G`. That is,
+    /// the distribution of its output is not uniformly random in `G`: the set of possible outputs
+    /// of encode_to_curve is only a fraction of the points in `G`, and some points in this set
+    /// are more likely to be output than others.
+    ///
+    /// If your application needs the distribution of the output to be statistically close to
+    /// uniform in `G`, use [hash_to_curve] instead.
+    ///
     /// # Panics
     /// Panics if `domain_sep.collect().len() == 0` or `> 255`
     pub fn encode_to_curve<D>(bytes: &[&[u8]], domain_sep: &[&[u8]]) -> EdwardsPoint
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -635,7 +635,7 @@ impl EdwardsPoint {
     }
 
     #[cfg(feature = "digest")]
-    /// Perform hashing to curve, with explicit hash function and domain separator, `domain_sep`,
+    /// Perform map to curve, with explicit hash function and domain separator, `domain_sep`,
     /// using the suite `edwards25519_XMD:SHA-512_ELL2_NU_`. The input is the concatenation of the
     /// elements of `bytes`. Likewise for the domain separator with `domain_sep`. At least one
     /// element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed
@@ -643,7 +643,7 @@ impl EdwardsPoint {
     ///
     /// # Panics
     /// Panics if `domain_sep.collect().len() == 0` or `> 255`
-    pub fn hash_to_curve<D>(bytes: &[&[u8]], domain_sep: &[&[u8]]) -> EdwardsPoint
+    pub fn encode_to_curve<D>(bytes: &[&[u8]], domain_sep: &[&[u8]]) -> EdwardsPoint
     where
         D: BlockSizeUser + Default + FixedOutput<OutputSize = U64> + HashMarker,
         D::BlockSize: IsGreater<D::OutputSize, Output = True>,
@@ -2448,7 +2448,7 @@ mod test {
                 }
             };
 
-            let computed = EdwardsPoint::hash_to_curve::<sha2::Sha512>(&[&input], &[dst]);
+            let computed = EdwardsPoint::encode_to_curve::<sha2::Sha512>(&[&input], &[dst]);
             assert_eq!(computed, expected_output, "Failed in test {}", index);
         }
     }
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -2438,7 +2438,7 @@ mod test {
     // https://www.rfc-editor.org/rfc/rfc9380.html#appendix-J.5.2
     // These are of the form (input_msg, output_x, output_y)
     #[cfg(all(feature = "alloc", feature = "digest"))]
-    const RFC_HASH_TO_CURVE_KAT_NU: &[(&[u8], &str, &str)] = &[
+    const RFC_ENCODE_TO_CURVE_KAT: &[(&[u8], &str, &str)] = &[
         (
             b"",
             "1ff2b70ecf862799e11b7ae744e3489aa058ce805dd323a936375a84695e76da",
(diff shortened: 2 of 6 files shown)


curve25519-dalek/src/edwards.rs
```diff
@@ -635,50 +635,96 @@ impl EdwardsPoint {
...
+    /// of encode_to_curve is only a fraction of the points in `G`, and some points in this set
+    /// are more likely to be output than others.
+    ///
+    /// If your application needs the distribution of the output to be statistically close to
+    /// uniform in `G`, use [hash_to_curve] instead.
+    ///
     /// # Panics
     /// Panics if `domain_sep.collect().len() == 0` or `> 255`
-    pub fn hash_to_curve<D>(bytes: &[&[u8]], domain_sep: &[&[u8]]) -> EdwardsPoint
+    pub fn encode_to_curve<D>(bytes: &[&[u8]], domain_sep: &[&[u8]]) -> EdwardsPoint
```
tarcieri, line 698: We did already release this in v4.2.0, and this is a breaking change.

It becomes a question of yanking that release and making a breaking change, or keep the old name, document the problem, and fix it in v5

rozbb: I think this is worth a yank. Calling encode-to-curve hash-to-curve is a significant security issue imo

rozbb:
Thank you so much for the quick turnaround! I will make time to review this in the next few days

curve25519-dalek/src/edwards.rs
tarcieri: Okay, we should probably yank it then, before people start using it

curve25519-dalek/benches/dalek_benchmarks.rs
```diff
@@ -85,7 +85,7 @@ mod edwards_benches {
 
         c.bench_function(
             "Elligator2 hash to curve (SHA-512, input size 32 bytes)",
```
armfazh: I suggest adding benchmarks for both encode_to_curve and hash_to_curve.

```suggestion
            "Elligator2 encode to curve (SHA-512, input size 32 bytes)",
```

curve25519-dalek/src/edwards.rs
```diff
@@ -2426,9 +2472,9 @@ mod test {
 
     #[test]
     #[cfg(all(feature = "alloc", feature = "digest"))]
-    fn elligator_hash_to_curve_test_vectors() {
+    fn elligator_map_to_curve_test_vectors() {
```
armfazh:
```suggestion
    fn elligator_encode_to_curve_test_vectors() {
```

curve25519-dalek/src/edwards.rs
```diff
@@ -2448,8 +2494,75 @@ mod test {
...
             };
 
-            let computed = EdwardsPoint::hash_to_curve::<sha2::Sha512>(&[&input], &[dst]);
+            let computed = EdwardsPoint::encode_to_curve::<sha2::Sha512>(&[&input], &[dst]);
             assert_eq!(computed, expected_output, "Failed in test {}", index);
         }
     }
+
+    // Hash-to-curve test vectors from
+    // https://www.rfc-editor.org/rfc/rfc9380.html#name-edwards25519_xmdsha-512_ell
```
armfazh:
```suggestion
    // https://www.rfc-editor.org/rfc/rfc9380.html#appendix-J.5.1
```

curve25519-dalek/src/edwards.rs
```diff
@@ -2390,7 +2436,7 @@ mod test {
     // https://www.rfc-editor.org/rfc/rfc9380.html#name-edwards25519_xmdsha-512_ell2
```
armfazh:
```suggestion
    // https://www.rfc-editor.org/rfc/rfc9380.html#appendix-J.5.2
```

curve25519-dalek/src/field.rs
```diff
@@ -721,6 +770,13 @@ mod test {
...
 
+    #[cfg(feature = "digest")]
+    fn fe_from_test_vector(expected_hex: &str) -> FieldElement {
+        let mut expected_hash = hex::decode(expected_hex).unwrap();
+        expected_hash.reverse();
+        FieldElement::from_bytes(&expected_hash.try_into().unwrap())
+    }
+
     /// Hash to field test vectors from
     /// https://www.rfc-editor.org/rfc/rfc9380.html#name-edwards25519_xmdsha-512_ell2
```
armfazh:
```suggestion
    /// https://www.rfc-editor.org/rfc/rfc9380.html#appendix-J.5.2
```

curve25519-dalek/src/field.rs
```diff
@@ -761,14 +817,67 @@ mod test {
...
-            };
+            let fe = FieldElement::hash_to_field::<Sha512, 1>(&[msg], &[dst.as_bytes()])[0];
+            let expected_fe = fe_from_test_vector(expected_hash_hex);
 
             assert_eq!(fe, expected_fe);
         }
     }
+
+    /// Hash to field test vectors from
+    /// https://www.rfc-editor.org/rfc/rfc9380.html#name-edwards25519_xmdsha-512_ell
```
armfazh:
```suggestion
    /// https://www.rfc-editor.org/rfc/rfc9380.html#appendix-J.5.1
```

curve25519-dalek/src/edwards.rs
```diff
@@ -635,50 +635,96 @@ impl EdwardsPoint {
     }
 
     #[cfg(feature = "digest")]
-    /// Perform hashing to curve, with explicit hash function and domain separator, `domain_sep`,
+    // The function `map_to_curve` calculates an [EdwardsPoint] from a [FieldElement].
+    fn map_to_curve(fe: FieldElement) -> EdwardsPoint {
+        let c1 = ED25519_SQRTAM2;
+
+        // 1.  (xMn, xMd, yMn, yMd) = map_to_curve_elligator2_curve25519(u)
+        let (xMn, xMd, yMn, yMd) = crate::montgomery::elligator_encode(&fe);
```
armfazh, line 643: This code can be simplified since `yMd=1`.

```suggestion
        let (xMn, xMd, yMn, _) = crate::montgomery::elligator_encode(&fe);
```

curve25519-dalek/src/edwards.rs
```diff
@@ -635,50 +635,96 @@ impl EdwardsPoint {
...
+        // 11. yn = CMOV(yn, 1, e)
+        let yn = FieldElement::conditional_select(&yn, &FieldElement::ONE, e);
+        // 12. yd = CMOV(yd, 1, e)
+        let yd = FieldElement::conditional_select(&yd, &FieldElement::ONE, e);
+        // 13. return (xn, xd, yn, yd)
+
+        let x = &xn * &xd.invert();
+        let y = &yn * &yd.invert();
+
+        AffinePoint { x, y }.to_edwards()
```
armfazh: No inverse are required

```suggestion
        EdwardsPoint {
            X: &xn * &yd,
            Y: &xd * &yn,
            Z: &xd * &yd,
            T: &xn * &yn,
        }
```

curve25519-dalek/src/edwards.rs
```diff
@@ -635,50 +635,96 @@ impl EdwardsPoint {
...
+        // 8.   e = tv1 == 0
+        let e = tv1.ct_eq(&FieldElement::ZERO);
+        // 9.  xn = CMOV(xn, 0, e)
+        let xn = FieldElement::conditional_select(&xn, &FieldElement::ZERO, e);
+        // 10. xd = CMOV(xd, 1, e)
+        let xd = FieldElement::conditional_select(&xd, &FieldElement::ONE, e);
+        // 11. yn = CMOV(yn, 1, e)
+        let yn = FieldElement::conditional_select(&yn, &FieldElement::ONE, e);
+        // 12. yd = CMOV(yd, 1, e)
+        let yd = FieldElement::conditional_select(&yd, &FieldElement::ONE, e);
```
daxpedda, line 665: Personally I tend to simplify to this:
```suggestion
        // 9.  xn = CMOV(xn, 0, e)
        xn.conditional_assign(&FieldElement::ZERO, e);
        // 10. xd = CMOV(xd, 1, e)
        xd.conditional_assign(&FieldElement::ONE, e);
        // 11. yn = CMOV(yn, 1, e)
        yn.conditional_assign(&FieldElement::ONE, e);
        // 12. yd = CMOV(yd, 1, e)
        yd.conditional_assign(&FieldElement::ONE, e);
```

curve25519-dalek/src/field.rs
```diff
@@ -361,13 +361,31 @@ impl FieldElement {
     /// element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed 255 bytes.
     ///
     /// # Panics
-    /// Panics if `domain_sep.collect().len() == 0` or `> 255`
-    pub fn hash_to_field<D>(bytes: &[&[u8]], domain_sep: &[&[u8]]) -> Self
+    /// Panics if `domain_sep.collect().len() == 0` or `> 255`.
+    /// If COUNT > 2.
+    pub fn hash_to_field<D, const COUNT: usize>(
```
daxpedda, line 369:
```suggestion
    pub fn hash_to_field<D, const COUNT: u16>(
```
Should probably be more convenient.

curve25519-dalek/src/field.rs
```diff
@@ -391,6 +409,8 @@ impl FieldElement {
             "Domain separator MUST have nonzero length."
         );
 
+        // We statically allocate `hash_outputs` to its maximum permitted length.
+        let mut hash_outputs = [Array::<u8, D::OutputSize>::default(); 48 * 2];
```
daxpedda:
```suggestion
        let mut hash_outputs = [Array::<u8, D::OutputSize>::default(); 2];
```

curve25519-dalek/src/field.rs
```diff
@@ -399,14 +419,43 @@ impl FieldElement {
             hasher = hasher.chain_update(slice)
         }
 
-        let b_1 = hasher.chain_update([domain_sep_len]).finalize();
+        hash_outputs[0] = hasher.chain_update([domain_sep_len]).finalize();
+
+        for i in 2..=ell {
```
daxpedda: This for-loop is a bit strange, because in practice it can only run once.
It might be easier to optimize by the compiler when its just an `if`.

On the other hand this is much easier to compare to the specification.

curve25519-dalek/src/field.rs
```diff
@@ -399,14 +419,43 @@ impl FieldElement {
...
+                hasher = hasher.chain_update(slice)
+            }
+
+            hash_outputs[i - 1] = hasher.chain_update([domain_sep_len]).finalize();
+        }
+
+        let concatenated_outputs: &[u8] = unsafe {
+            core::slice::from_raw_parts(
+                hash_outputs.as_ptr() as *const u8,
+                48 * 2 * D::output_size(),
```
daxpedda:
```suggestion
                2 * D::output_size(),
```
Adjusting to the fix above.

curve25519-dalek/src/field.rs
```diff
@@ -399,14 +419,43 @@ impl FieldElement {
...
+            )
+        };
```
daxpedda: Considering this is all fixed size here, you could just concat those.
```suggestion
        let concatenated_outputs = hash_outputs[0].concat(hash_outputs[1]);
```

curve25519-dalek/src/montgomery.rs
```diff
@@ -252,33 +254,98 @@ impl MontgomeryPoint {
...
-    let (eps_is_sq, _eps) = FieldElement::sqrt_ratio_i(&eps, &one);
-
-    let zero = FieldElement::ZERO;
-    let Atemp = FieldElement::conditional_select(&MONTGOMERY_A, &zero, eps_is_sq); /* 0, or A if nonsquare*/
-    let mut u = &d + &Atemp; /* d, or d+A if nonsquare */
-    u.conditional_negate(!eps_is_sq); /* d, or -d-A if nonsquare */
-
-    (MontgomeryPoint(u.to_bytes()), eps_is_sq)
+    let two = &one + &one;
+    let c2 = &two * &(two.pow_p58());
```
daxpedda: These should be made const.
For your convenience I printed out `c2`:
```rust
// 64-bit
FieldElement::from_limbs([1718705420411057, 234908883556509, 2233514472574048, 2117202627021982, 765476049583133]);
// 32-bit
FieldElement::from_limbs([34513073, 25610706, 9377949, 3500415, 12389472, 33281959, 41962654, 31548777, 326685, 11406482]);
```

Considering `FieldElement` methods could already be made const at the current MSRV, these could actually be calculated in-place during compile-time. But probably best left for a separate PR.

curve25519-dalek/src/montgomery.rs
```diff
@@ -252,33 +254,98 @@ impl MontgomeryPoint {
...
+    let tv2 = &tv2 * &gxd;
+    // 34.  e3 = tv2 == gx1
+    let e3 = tv2.ct_eq(&gx1);
+    // 35.  xn = CMOV(x2n, x1n, e3)
+    let xn = FieldElement::conditional_select(&x2n, &x1n, e3);
+    // 36.   y = CMOV(y2, y1, e3)
+    let y = FieldElement::conditional_select(&y2, &y1, e3);
+    // 37.  e4 = sgn0(y) == 1
+    // TODO: check that this step is correct
+    let e4 = y.is_negative();
```
daxpedda: This is indeed correct, see https://www.rfc-editor.org/rfc/rfc9380.html#section-4-4.5.
I didn't check the implementation of `is_negative()` though.

curve25519-dalek/src/montgomery.rs
```diff
@@ -252,33 +254,98 @@ impl MontgomeryPoint {
...
+    // 38.   y = CMOV(y, -y, e3 XOR e4)
+    let y = FieldElement::conditional_select(&y, &y.neg(), e3 ^ e4);
```
daxpedda, line 359:
```suggestion
    y.conditional_negate(e3 ^ e4);
```

daxpedda (approved):
Just a bunch of nits.
LGTM!

curve25519-dalek/src/field.rs
iquerejeta: I went for `usize` because I use it for defining the length of the output array.

curve25519-dalek/src/montgomery.rs
iquerejeta: Oh, leftover of notes to myself. I also checked, and I believe it is correct 👍  I'll remove the TODO.

curve25519-dalek/src/field.rs
iquerejeta: Good catch. Your suggestion is true if we use SHA512, but not if we use SHA256 (we'd need 3). I see two possible directions:

1. We fix the implementation to SHA512 as we specify in the docs, and change accordingly. Or, 
2. We modify the docs to specify that the implementation is generic to the hash function (and only fix it in the hash_to_curve function), and use hash_outputs with max_length `3` (as a hash function with output size < 256 should not be used, as suggested by the spec).

Which is your preferred fix?

curve25519-dalek/src/field.rs
iquerejeta: I'll change it depending on the decision above.

curve25519-dalek/src/edwards.rs
rozbb: I also think this should not be `pub`. I've never seen a reason to use this function

curve25519-dalek/src/field.rs
rozbb: I've resolve this and some other comments in a refactor. This function only permits hashes with 64-byte digests anyway. Lmk what you think of the new code

curve25519-dalek/src/field.rs
rozbb: Ditto re refactor

curve25519-dalek/src/montgomery.rs
rozbb: Good call. Just added

curve25519-dalek/src/montgomery.rs
rozbb: Fixed

rozbb:
These are fantastic edits, thank you! I just refactored stuff a bit. I'll continue tomorrow and hopefully merge this week

curve25519-dalek/src/edwards.rs
rozbb: True true, tho I'd rather stick to the pseudocode as tightly as possible

curve25519-dalek/src/edwards.rs
rozbb: This requires `let mut xn; let mut xd; let mut yn; let mut yd`, no? If so I think avoiding the `mut` and extra line is a good thing

curve25519-dalek/src/montgomery.rs
rozbb: The def of `conditional_negate` expands to the above, and IMO it's nice to have the same op for all the CMOVs in the comments. I'm inclined to keep this as is

rozbb:
Question: do we need `encode_to_curve` anymore?? I've never actually seen it used in the wild. If we kill it, we get to remove it and all its tests too

iquerejeta:
I think `encode_to_curve` should stay, and IMO be public. It is what we actually use in the VRF function (see [the standard](https://www.rfc-editor.org/rfc/rfc9381.html#name-ecvrf-ciphersuites), particularly the `ECVRF-EDWARDS25519-SHA512-ELL2` suite). Uniformity is not always required.

curve25519-dalek/src/edwards.rs
tarcieri: Hmm, six of one, half dozen of another here. I guess I weakly prefer how it is now with let bindings and assignments just because it more closely matches the pseudocode (in other cases I'd probably use `conditional_assign`)

OtaK:
> Question: do we need `encode_to_curve` anymore?? I've never actually seen it used in the wild. If we kill it, we get to remove it and all its tests too

Yes, as @iquerejeta pointed out, e2c *needs* to stay public in order to implement VRFs.

curve25519-dalek/src/edwards.rs
```diff
@@ -635,50 +635,98 @@ impl EdwardsPoint {
...
+    /// elements of `bytes`. Likewise for the domain separator with `domain_sep`. At least one
+    /// element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed
+    /// 255 bytes.
+    ///
+    /// # Panics
+    /// Panics if `domain_sep.collect().len() == 0` or `> 255`
+    pub fn hash_to_curve<D>(bytes: &[&[u8]], domain_sep: &[&[u8]]) -> EdwardsPoint
+    where
+        D: BlockSizeUser + Default + FixedOutput<OutputSize = U64> + HashMarker,
+        D::BlockSize: IsGreater<D::OutputSize, Output = True>,
```
armfazh, line 727: I understand that letting `D` to be specified by the user allows `hash_to_curve` to use different SHA512 implementations. 
However, nothing prevents to use other functions, e.g. anyone can do

```diff
-  let computed = EdwardsPoint::hash_to_curve::<sha2::Sha512>(&[&input], &[dst]);
+  let computed = EdwardsPoint::hash_to_curve::<sha3::Sha3_512>(&[&input], &[dst]);
```

and get wrong/uncompatible results.

The library must fix SHA512 as the hash function due to this comes from the hash to curve ciphersuite specification: `edwards25519_XMD:SHA-512_ELL2_NU_` and `edwards25519_XMD:SHA-512_ELL2_RO_`.

On the other hand, note that `hash_to_field` accepts Merkle-Damgard (MD) fuctions, such as the SHA2 and SHA3 hash functions, and XOF functions such as Shake128. It's ok, if `hash_to_field` only supports MD functions and COUNT=1 or COUNT=2.

curve25519-dalek/src/field.rs
```diff
@@ -355,59 +355,132 @@ impl FieldElement {
...
     {
-        let l_i_b_str = 48u16.to_be_bytes();
-        let z_pad = Array::<u8, D::BlockSize>::default();
-
-        let mut hasher = D::new().chain_update(z_pad);
-
-        for slice in bytes {
-            hasher = hasher.chain_update(slice);
+        // We only use `hash_to_field` for Elligator2, which uses COUNT <= 2
+        assert!(COUNT <= 2);
```
armfazh:
```suggestion
        assert!(0 < COUNT && COUNT <= 2);
```

armfazh (changes requested):
One comment regarding the specification of a generic SHA2 function.

curve25519-dalek/src/edwards.rs
tarcieri: @armfazh one way to do that is using the `AssociatedOid` trait. You could potentially could check it with the type system using:

```rust
D: AssociatedOid<OID = sha2::Sha512::OID>
```

daxpedda: I don't believe the [specification enforces such limitations](https://www.rfc-editor.org/rfc/rfc9380.html#section-5.3.1-1):
> - H **MAY** be a Merkle-Damgaard hash function like SHA-2. In this case, security holds when the underlying compression function is modeled as a random oracle [[CDMP05](https://www.rfc-editor.org/rfc/rfc9380.html#CDMP05)]. (See [Section 10.6](https://www.rfc-editor.org/rfc/rfc9380.html#security-considerations-expand-xmd) for discussion.)
> - H **MAY** be a sponge-based hash function like SHA-3 or BLAKE2. In this case, security holds when the inner function is modeled as a random transformation or as a random permutation [[BDPV08](https://www.rfc-editor.org/rfc/rfc9380.html#BDPV08)].
> - Otherwise, H **MUST** be a hash function that has been proved indifferentiable from a random oracle [[MRH04](https://www.rfc-editor.org/rfc/rfc9380.html#MRH04)] under a reasonable cryptographic assumption.

The last point should allow us to use any "reasonable" hash here. I believe enforcing `HashMarker` is appropriate enough for this case.

We also don't enforce the suite ID, users can pass whatever they think is right with `dst`. While we are enforcing parts of the "default" suite like Elligator 2 mapping, these aren't incompatible with different hashes according to the specification. E.g. the [Elligator 2 mapping](https://www.rfc-editor.org/rfc/rfc9380.html#name-mappings-for-twisted-edward) only requires certain curves and has no relation to the hash in use.

iquerejeta: I did consider fixing the hash function, but as @daxpedda says, the spec does allows using any hash function as long as is meets certain properties. 

Having the digest generic allows users to use suites other than `edwards25519_XMD:SHA-512_ELL2_NU_` and `edwards25519_XMD:SHA-512_ELL2_RO_`

armfazh: I agree that the generality on choosing the hash function applies to `hash_to_field`, but may or may not apply to `encode_to_curve` and `hash_to_curve`, and this depends on what you want to provide to users.

Option A: 
Strict ciphersuites: only `edwards25519_XMD:SHA-512_ELL2_RO_` and `edwards25519_XMD:SHA-512_ELL2_NU_` suites. In this case, only the SHA512 function must be used, otherwise the test vectors will not match. Users are only allowed to switch between SHA512-compliant implementations.

Option B:   
Strict + other ciphersuites: This is Option A, plus allowing the user to choose any hash function, i.e. any secure or insecure hash function. This passes the responsability to users on choosing an appropriate hash function (assuming they have read the RFC).

I consider the library should go for Option A: strict ciphersuites.
 
More ciphersuites can be defined in future. In fact, the second most attractive option is to switch from Merkle-Damgard functions (SHA2 familiy) to eXtendable Output functions (Shake familiy), however this is not supported in this PR, and more changes are needed.

OtaK: My 2 cents: Option A would cause users to create forks down the line for anyone wanting to do e2c/h2c with custom suites. I don't think we want this, users are already responsible with passing down the correct DST, passing down a correct hash function for the DST you're already responsible for isn't too far off either.

daxpedda: > This passes the responsability to users on choosing an appropriate hash function (assuming they have read the RFC).

I think my earlier point is that we have all the right checks in place so that it can't actually be misused. Do you have a scenario in mind that actually is able to misuse this API?

curve25519-dalek/src/field.rs
rozbb: Good catch, just added ty

curve25519-dalek/src/edwards.rs
armfazh: The scenario I foresee is someone (who didn't read the documentation properly) start using their favorite hash function and then claiming to be compatible with the named ciphersuite. 
This is an error on the user-side, of course. Is this something, the library should prevent from happening? 

I agree that it is easier to use a generic trait for the sake of agility, and to avoid fragmentation.

In any path you decide to take, please add some explanatory text on the documentation about it.

armfazh (commented):
See my comment on at https://github.com/dalek-cryptography/curve25519-dalek/pull/786/files#r2277816872

Removing 'Request Changes' state.

armfazh (approved):
Removing `Request changes` state.

curve25519-dalek/src/edwards.rs
daxpedda: If I'm getting this right, the problem you are describing is that this function doesn't enforce the specific ciphersuite from the spec but users might get confused about that.

Warning against that makes a lot of sense to me.

curve25519-dalek/src/edwards.rs
```diff
@@ -635,50 +635,98 @@ impl EdwardsPoint {
...
+    }
+
+    #[cfg(feature = "digest")]
+    /// Perform encode to curve per RFC 9380, with explicit hash function and domain separator, `domain_sep`,
     /// using the suite `edwards25519_XMD:SHA-512_ELL2_NU_`. The input is the concatenation of the
     /// elements of `bytes`. Likewise for the domain separator with `domain_sep`. At least one
     /// element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed
     /// 255 bytes.
     ///
+    /// # Warning
```
daxpedda:
```suggestion
    /// Perform encode to curve per RFC 9380, with explicit hash function and domain separator, `domain_sep`,
    /// using the Twisted Edwards Elligator 2 method. The input is the concatenation of the
    /// elements of `bytes`. Likewise for the domain separator with `domain_sep`. At least one
    /// element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed
    /// 255 bytes.
    ///
    /// # Warnings
    ///
    /// This function doesn't enforce the `edwards25519_XMD:SHA-512_ELL2_NU_` ciphersuite as per RFC 9380.
    /// It is up to the caller to use this function with a ciphersuite of their choice.
```

curve25519-dalek/src/edwards.rs
```diff
@@ -635,50 +635,98 @@ impl EdwardsPoint {
...
-        // followed by the mapping
-        //     https://www.rfc-editor.org/rfc/rfc9380.html#name-mappings-for-twisted-edward
-        // The only difference is that our `elligator_encode` returns only the Montgomery `u` coordinate,
-        // so we apply this workaround to reconstruct and validate the sign.
+    #[cfg(feature = "digest")]
+    /// Perform a hash to curve per RFC 9380, with explicit hash function and domain separator, `domain_sep`,
+    /// using the suite `edwards25519_XMD:SHA-512_ELL2_RO_`. The input is the concatenation of the
+    /// elements of `bytes`. Likewise for the domain separator with `domain_sep`. At least one
+    /// element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed
+    /// 255 bytes.
```
daxpedda:
```suggestion
    /// Perform a hash to curve per RFC 9380, with explicit hash function and domain separator, `domain_sep`,
    /// using the Twisted Edwards Elligator 2 method. The input is the concatenation of the
    /// elements of `bytes`. Likewise for the domain separator with `domain_sep`. At least one
    /// element of `domain_sep`, MUST be nonempty, and the concatenation MUST NOT exceed
    /// 255 bytes.
    ///
    /// # Warnings
    ///
    /// This function doesn't enforce the `edwards25519_XMD:SHA-512_ELL2_RO_` ciphersuite as per RFC 9380.
    /// It is up to the caller to use this function with a ciphersuite of their choice.
```

daxpedda (changes requested):
Following up from #786 (comment), the documentation is clearly wrong and can easily mislead users.

curve25519-dalek/src/edwards.rs
rozbb: Thank you, these changes make sense. I had some language in the hash-to-field function, so I copied that warning, along with some of your language here

daxpedda (approved):

PR #792, opened 2025-07-17
Use tuple for `MultiscalarMul`

daxpedda:
`MultiscalarMul` currently takes two iterators for scalars and points respectively. This has the downside that if they are differently sized we currently simply panic.

This PR changes that by putting them both into one iterator with `Item = (Point, Scalar)`. This way we get rid of the panic and don't loose any of the flexibility.
Migration for users should be quite straightforward, just add a `zip()` between the two parameters, as shown in the example.

I also switched the order of scalars and points around to align more with `elliptic_curve::LinearCombination` and applied the same tuple-treatment to the fixed-sized array method as well. This does affect its flexibility and maybe we should change `elliptic_curve::LinearCombination` accordingly.

As requested in #790.


tarcieri:
It seems like if we do this, it should be done for `VartimeMultiscalarMul` and `VartimePrecomputedMultiscalarMul` as well

daxpedda:
> It seems like if we do this, it should be done for `VartimeMultiscalarMul` and `VartimePrecomputedMultiscalarMul` as well

Sounds good! Will do it when we get #790 in.

I can also turn this around if you want, rebase #790 on top of this, then we can do this one first.
Let me know!

PR #793, opened 2025-07-17
workspace: use a global patch.crates-io to avoid duplicates

baloo:
When pulling crates via git (to grab non-released yet fixes or otherwise), the `path = "../"` in each crate crates duplicates unless you pull the whole tree in your local `[patch.crates-io]`.

This creates issues in downstream packaging (nix, ...) where a crate version can only appear once.
Those are somewhat difficult to diagnose.

Using a `[patch.crates-io]` in the workspace serves the same purpose but does not create the duplication in consumers' tree.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,3 +10,7 @@ resolver = "2"
 [profile.dev]
 opt-level = 2
 
+[patch.crates-io]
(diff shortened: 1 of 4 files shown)


tarcieri (approved):
I'm a fan of this approach, unless @rozbb objects

PR #804, opened 2025-08-13
Add `Scalar::halve()`

daxpedda:
This adds a new method `Scalar::halve()` which halves the given `Scalar`.

This can facilitate batch encoding the result of scalar-point multiplications by halving the `Scalar` and then using `RistrettoPoint::double_and_compress_batch()`. I've added a test for this.

### DIFF
--- a/curve25519-dalek/src/ristretto.rs
+++ b/curve25519-dalek/src/ristretto.rs
@@ -1868,6 +1868,35 @@ mod test {
     }
 
     #[test]
+    #[cfg(all(feature = "alloc", feature = "rand_core", feature = "group"))]
+    fn multiply_double_and_compress_1024_random_points() {
+        use ff::Field;
+        use group::Group;
+        let mut rng = OsRng;
+
+        let mut scalars: Vec<Scalar> = (0..1024)
+            .map(|_| Scalar::try_from_rng(&mut rng).unwrap())
+            .collect();
+        scalars[500] = Scalar::ZERO;
+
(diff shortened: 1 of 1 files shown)


tarcieri:
It seems like you could implement a much more efficient divide-by-two operation using a right shift over the limbs of the unsaturated 52-bit representation used by `Scalar`

daxpedda:
I originally did exactly that, but that doesn't have the same output and doesn't seem to work; the test will fail.
Not sure exactly why, IANAC.

tarcieri:
@daxpedda I think it might need the same trick that the `div_by_2` implementation in `crypto-bigint` uses, where it conditionally adds the modulus if odd

tarcieri:
#805 should be close, I think

daxpedda:
I see! Seems to work great!
I've boiled down the PR to just adding the relevant test test and targeting your `scalar-div-by-2` branch.

curve25519-dalek/src/ristretto.rs
```diff
@@ -1867,6 +1867,35 @@ mod test {
         }
     }
 
+    #[test]
+    #[cfg(all(feature = "alloc", feature = "rand_core", feature = "group"))]
+    fn multiply_double_and_compress_1024_random_points() {
```
tarcieri, line 1872: We should really get `proptest` set up, this is fine for now though.

cc @rozbb

tarcieri (approved):

PR #805, opened 2025-08-14
[WIP] Scalar::div_by_2

tarcieri:


curve25519-dalek/src/scalar.rs
```diff
@@ -831,6 +831,21 @@ impl Scalar {
...
+    pub fn div_by_2(&self) -> Self {
+        // We are looking for such `b` that `b + b = a mod modulus`.
+        // Two possibilities:
+        // - if `a` is even, we can just divide by 2;
+        // - if `a` is odd, we divide `(a + modulus)` by 2.
+        let is_odd = Choice::from(self.as_bytes()[0] & 1);
+        let mut scalar = self.unpack();
+        scalar.conditional_add_l(is_odd);
+
+        // TODO(tarcieri): propagate carry
```
tarcieri: Please double check me on this, but adding the modulus (which is odd) should ensure the value is even, so the carry bit here is always zero, so we don't need to propagate it.

(I guess I can add a debug assert)

curve25519-dalek/src/scalar.rs
```diff
@@ -1677,6 +1693,23 @@ pub(crate) mod test {
...
+            let dividend = scalar.div_by_2();
+            assert_eq!(scalar, dividend + dividend);
+        }
+
+        // test a range of scalars near the modulus
+        for i in 0u64..32 {
+            let scalar = Scalar::ZERO - Scalar::from(i);
+            let dividend = scalar.div_by_2();
+            assert_eq!(scalar, dividend + dividend);
+        }
```
tarcieri, line 1710: Not sure if I should add a test or leave a comment to this effect, but to double check one other thing:

```
l=0x1000000000000000000000000000000014def9dea2f79cd65812631a5cf5d3ed
x=l-2 # largest odd canonical scalar
frac_2_x=(x + l) >> 1
frac_2_x < l # true
```

Edit: I guess it should be obvious that (l + (l - 2)) / 2 < l, since (2l - 2)/2 = l - 1

curve25519-dalek/src/ristretto.rs
```diff
@@ -1867,6 +1867,35 @@ mod test {
...
+            .iter()
+            .zip(&points)
+            .map(|(scalar, point)| scalar.div_by_2() * point)
+            .collect();
+        let compressed = RistrettoPoint::double_and_compress_batch(&multiplied_points);
+
+        for ((s, P), P2_compressed) in scalars.iter().zip(points).zip(compressed) {
+            assert_eq!(P2_compressed, (s * P).compress());
+        }
+    }
```
tarcieri: This was added from #804, where I suggested converting it to use `proptest`, which I'm fine with doing in this PR

daxpedda: See #806.

PR #806, opened 2025-08-15
`scalar-div-by-2`: test with Proptest

daxpedda:
Should I convert more tests to Proptest right here or in a follow-up?
Alternatively I could make a PR to main adding Proptest and then we can rebase #805 on top of that.


tarcieri:
I'm fine with this as-is, though after we merge #805 a followup to convert any other randomized testing to proptests would be great (as well as adding new proptests where they make sense).

daxpedda:
> though after we merge #805 a followup to convert any other randomized testing to proptests would be great

Will do the PR then after #805 is merged.

tarcieri:
@daxpedda #805 is merged! Feel free to submit a followup with more proptests

daxpedda:
See #809.

PR #809, opened 2025-09-01
Convert more tests to Proptest

daxpedda:
To make the code cleaner I implemented `Arbitrary` for various types where appropriate. The implementation is a bit naive because it simply maps by reducing instead of properly having a min/max implementation via a custom `ValueTree`.

I also switched to the newer `#[property_test]` instead of `proptest!`.

Follow-up from https://github.com/dalek-cryptography/curve25519-dalek/pull/806.

### DIFF
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -128,6 +128,9 @@ use subtle::ConstantTimeEq;
 #[cfg(feature = "zeroize")]
 use zeroize::Zeroize;
 
+#[cfg(test)]
+use proptest::{arbitrary::Arbitrary, strategy::BoxedStrategy};
+
 use crate::constants;
 
 use crate::field::FieldElement;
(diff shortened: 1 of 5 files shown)


curve25519-dalek/src/edwards.rs
```diff
@@ -2066,17 +2081,14 @@ mod test {
     }
 
     /// Check that mul_base_clamped and mul_clamped agree
-    #[test]
-    fn mul_base_clamped() {
-        let mut csprng = rand_core::OsRng;
+    #[property_test(config = ProptestConfig { cases: 50, .. ProptestConfig::default() })]
+    fn mul_base_clamped(b: EdwardsPoint, a: [[u8; 32]; 100]) {
```
daxpedda, line 2085: I think this would be better split up into multiple tests, but I didn't know exactly how.

So here is a suggestion, let me know what you think:
```rust
    /// Check that mul_base_clamped and mul_clamped agree

    #[test]
    fn mul_base_clamped_max() {
        // Test that mul_base_clamped and mul_clamped agree on a large integer. Even after
        // clamping, this integer is not reduced mod l.
        let a_bytes = [0xff; 32];
        assert_eq!(
            EdwardsPoint::mul_base_clamped(a_bytes),
            constants::ED25519_BASEPOINT_POINT.mul_clamped(a_bytes)
        );
    }

    #[property_test]
    fn mul_base_clamped(a_bytes: [u8; 32]) {
        prop_assert_eq!(
            EdwardsPoint::mul_base_clamped(a_bytes),
            constants::ED25519_BASEPOINT_POINT.mul_clamped(a_bytes)
        );
    }

    #[property_test(config = ProptestConfig { cases: 50, .. ProptestConfig::default() })]
    #[cfg(feature = "precomputed-tables")]
    fn mul_base_clamped_precomputed(b: EdwardsPoint, a_bytes: [u8; 32]) {
        // Make a random curve point in the curve. Give it torsion to make things interesting.
        let random_point = b + constants::EIGHT_TORSION[1];
        // Make a basepoint table from the random point. We'll use this with mul_base_clamped
        let random_table = EdwardsBasepointTableRadix256::create(&random_point);

        prop_assert_eq!(
            random_table.mul_base_clamped(a_bytes),
            random_point.mul_clamped(a_bytes)
        );
    }
```

PR #813, opened 2025-09-01
Example verus proofs

TheodoreEhrenborg:
Thanks for the call with BAIF @rozbb. I'm informally reaching out to ask some questions and show you what Verus looks like when applied to your codebase.

# Questions

1. Our current understanding is you don't want any Verus code merged into curve-dalek (i.e. you'd reject this PR). Is this correct?
2. We were considering maintaining a mirror repo which had a duplicate of your code and also contained the verus proofs, and then comparing to your official copy with https://crates.io/crates/cargo-expand to make sure we were in sync. (Verus annotations disappear when macros are expanded.) If we went down that route, we'd make some minor PRs to refactor the code into something Verus supports. How much refactoring would you tolerate? e.g. verus currently doesn't support tuple structs, so you'll notice that causes a lot of changes. My opinion is that removing tuple structs from curve-dalek is too much of a change, and we should just kindly ask the Verus team to add support for them. But for more minor changes, I might lean towards just making a PR to here
3. If we could find a proof engineer willing to maintain the Verus proofs for curve-dalek, would you be open to merging Verus code into curve-dalek? We can't commit to anything yet.

# Summary of changes

I don't intend to merge this PR because we're still working inside the u64 folder.
In particular, my coworker Jure has done a lot of work with `field.rs`, which I haven't included in this branch.

This branch has 8 proved functions
- `m`
- `add`
- `sub`
- `mul_internal`
- `square_internal`
- `montgomery_mul`
- `montgomery_square`
- `from_montgomery`

`add` and `sub` are the most complicated functions and have the most complicated proofs.


.github/workflows/curve25519-dalek.yml
```diff
@@ -43,30 +43,6 @@ jobs:
           RUSTFLAGS: '--cfg curve25519_dalek_backend="fiat"'
         run: cargo test --target ${{ matrix.target }}
 
-  # Default no_std test only tests using serial across all crates
```
TheodoreEhrenborg, line 46: Almost all of the CI still works. The no_std tests annoyingly don't. I haven't looked into this deeply yet---I hope it's something trivial

.github/workflows/curve25519-dalek.yml
```diff
@@ -143,3 +119,23 @@ jobs:
...
+    - name: Install Verus
+      run: |
+        wget https://github.com/verus-lang/verus/releases/download/release%2F0.2025.08.01.33c6cec/verus-0.2025.08.01.33c6cec-x86-linux.zip
+        unzip verus-0.2025.08.01.33c6cec-x86-linux.zip
+        mv verus-x86-linux ~/.cargo/bin
+        cd ~/.cargo/bin
+        ln -s verus-x86-linux/cargo-verus
+    - name: Run test
+      run: | 
+        cargo-verus verify -- --smt-option smt.random_seed=0
```
TheodoreEhrenborg, line 140: Checking the proofs in CI

curve25519-dalek/src/backend/serial/u32/constants.rs
TheodoreEhrenborg, line 1: There are no proofs in the u32 folder. But we had to change the types for compatibility because we got rid of tuple structs in the u64 folder

curve25519-dalek/src/backend/serial/u32/scalar.rs
```diff
@@ -132,38 +136,38 @@ impl Scalar29 {
...
-        s[23] =  (self.0[6] >> 10)                      as u8;
-        s[24] =  (self.0[6] >> 18)                      as u8;
-        s[25] = ((self.0[6] >> 26) | (self.0[7] << 3))  as u8;
-        s[26] =  (self.0[7] >>  5)                      as u8;
-        s[27] =  (self.0[7] >> 13)                      as u8;
-        s[28] =  (self.0[7] >> 21)                      as u8;
-        s[29] =  (self.0[8] >>  0)                      as u8;
-        s[30] =  (self.0[8] >>  8)                      as u8;
-        s[31] =  (self.0[8] >> 16)                      as u8;
+        s[ 0] =  (self.limbs[0] >>  0)                      as u8;
```
TheodoreEhrenborg, line 139: More changes caused by Verus's lack of support for tuple structs

curve25519-dalek/src/backend/serial/u64/scalar_lemmas.rs
TheodoreEhrenborg, line 1: This is where the bulk of the proofs is. All of this is non-executable verus code. When macros are expanded, it shrinks down to

```rust
pub mod scalar_lemmas {
    #[allow(unused_imports)]
    use super::constants;
    #[allow(unused_imports)]
    use super::scalar::Scalar52;
    #[allow(unused_imports)]
    use super::scalar_specs::*;
    #[allow(unused_imports)]
    use vstd::arithmetic::div_mod::*;
    #[allow(unused_imports)]
    use vstd::arithmetic::mul::*;
    #[allow(unused_imports)]
    use vstd::arithmetic::power2::*;
    #[allow(unused_imports)]
    use vstd::bits::*;
    #[allow(unused_imports)]
    use vstd::calc;
    use vstd::prelude::*;
}
```

curve25519-dalek/src/backend/serial/u64/scalar_specs.rs
```diff
@@ -0,0 +1,69 @@
+#[allow(unused_imports)]
+use super::scalar::Scalar52;
+#[allow(unused_imports)]
+use vstd::arithmetic::power2::*;
+use vstd::prelude::*;
+
+verus! {
+pub open spec fn seq_to_nat(limbs: Seq<nat>) -> nat
```
TheodoreEhrenborg, line 8: Conversion function from limbs to a natural number

curve25519-dalek/src/backend/serial/u64/subtle_assumes.rs
```diff
@@ -0,0 +1,34 @@
+//! Tell Verus what Choice does
```
TheodoreEhrenborg, line 1: `subtle` doesn't have verus annotations, so instead we tell verus here what `subtle` does. Hence we wrap `u64::conditional_select` in an annotated function `select`

curve25519-dalek/src/lib.rs
```diff
@@ -23,13 +23,7 @@
 // Linting:
 //------------------------------------------------------------------------
 #![cfg_attr(allow_unused_unsafe, allow(unused_unsafe))]
-#![warn(
-    clippy::unwrap_used,
-    missing_docs,
-    rust_2018_idioms,
-    unused_lifetimes,
-    unused_qualifications
```
TheodoreEhrenborg, line 31: Took this warning out for convenience, can likely put it back with only a little hassle

curve25519-dalek/src/backend/serial/u64/scalar.rs
```diff
@@ -118,211 +150,401 @@ impl Scalar52 {
...
+            proof {
+                lemma_add_loop_invariant(sum, carry, i, a, b, old_carry, mask, sum_loop_start);
+            }
+            proof {lemma_add_carry_and_sum_bounds(carry, mask);}
         }
 
+        assert(seq_u64_to_nat(a.limbs@.subrange(0, 5 as int)) + seq_u64_to_nat(b.limbs@.subrange(0, 5 as int)) ==
+               seq_u64_to_nat(sum.limbs@.subrange(0, 5 as int)) + (carry >> 52) * pow2((52 * (5) as nat)));
+
+        proof {lemma_add_sum_simplify(a, b, &sum, carry);}
```
TheodoreEhrenborg, line 246: A bunch of the proof has been refactored into lemmas like these, but you'll notice that `add` has gotten longer due to all the non-executable proof lines. Probably a few more could get refactored into lemmas, but not all of them

curve25519-dalek/src/backend/serial/u64/scalar.rs
```diff
@@ -118,211 +150,401 @@ impl Scalar52 {
...
         z
     }
 
+
+    // TODO Make this function more like the original?
     /// Compute `a^2`
     #[inline(always)]
-    #[rustfmt::skip] // keep alignment of return calculations
-    fn square_internal(a: &Scalar52) -> [u128; 9] {
-        let aa = [
```
TheodoreEhrenborg, line 223: Verus didn't like the `aa` shorthand. :man_shrugging: Hence the alterations

curve25519-dalek/src/backend/serial/u64/scalar.rs
```diff
@@ -118,211 +150,401 @@ impl Scalar52 {
...
     #[rustfmt::skip] // keep alignment of n* and r* calculations
-    pub (crate) fn montgomery_reduce(limbs: &[u128; 9]) -> Scalar52 {
+    #[verifier::external_body] // TODO Verify this function
+    pub (crate) fn montgomery_reduce(limbs: &[u128; 9]) -> (result: Scalar52)
+    ensures
+        (to_nat(&result.limbs) * montgomery_radix()) % group_order() == slice128_to_nat(limbs) % group_order(),
+        limbs_bounded(&result),
+    {
 
-        #[inline(always)]
```
TheodoreEhrenborg, line 248: Verus didn't like inline functions

curve25519-dalek/src/backend/serial/u64/scalar.rs
```diff
@@ -118,211 +150,401 @@ impl Scalar52 {
...
+    {
+        let w = (sum as u64) & ((1u64 << 52) - 1);
+        let carry = sum >> 52;
+        (carry, w)
     }
 
     /// Compute `a * b` (mod l)
     #[inline(never)]
-    pub fn mul(a: &Scalar52, b: &Scalar52) -> Scalar52 {
+    #[verifier::external_body] // TODO Verify this function
```
TheodoreEhrenborg, line 468: `external_body` means that verus ignores this function

.github/workflows/curve25519-dalek.yml
TheodoreEhrenborg: See https://github.com/Beneficial-AI-Foundation/dalek-lite/pull/59 for an example of the CI passing

rozbb:
Thank you so much for this! I might take a little bit to go thru it, but this is super exciting.

Stating an initial opinion: I'm happy to merge code changes (minor ones preferably), but it might be annoying to keep proof-related annotations in our code. I still want everything to be real Rust and parse in any contributor's IDE without modification. Would it be possible to keep all Verus-related stuff in a separate folder entirely? I'm not sure how strict the layout is.

TheodoreEhrenborg:
(Sorry for the delay in responding; BAIF has entered crunch time for another project)

> parse in any contributor's IDE without modification

This somewhat works, since rust-analyzer expands the verus macro and can't see the verus annotations. I'm not a VSCode user, but I just installed it and the rust-analyzer extension. Rust-analyzer seems to highlight the types correctly if you hover over a variable:
20250912vscode2

It's not perfect:
- inlay type hints don't display inside the verus macro
- if you use rust-analyzer to rename a variable, it only renames the variable in the Rust code, not in the annotations. So that would break a proof. (Although the bigger problem is that the proof of any function will likely break at least a little bit if the function itself changes. So ideally we could find a function that is guaranteed not to change for a while, which might not be a possible thing to guarantee)

> Would it be possible to keep all Verus-related stuff in a separate folder entirely?

In Verus, the proof annotations (the ensures/requires in the signature) have to appear on the executable function, and usually there have to be a few calls to lemmas too. There are other proof approaches (e.g. https://github.com/cryspen/hax) where you first run an extraction tool to translate the Rust code to Lean/F*/etc, and then you write a proof in Lean, all without modifying the original Rust code (unless there's a feature that isn't supported, e.g. tuple structs for Verus above). Hax also doesn't solve the problem of proofs breaking when the function changes. I hope that BAIF can find a way to teach an LLM to be a good enough proof engineer to repair proofs (easier than writing proofs from scratch), but no promises yet

PR #816, opened 2025-09-03
Add `expose-field` to expose `FieldElement`

### LINKED ISSUE #389: Exposing the field element API

kayabaNerve:
The underlying field elements are unsafe for public consumption as they have undefined arithmetic after a certain amount of uses. This trait solves the problem as following:

- Defining a `ff::Field` wrapper which reduces after _every operation_
- Defining a bespoke `LazyField` trait which tracks consumed capacity using `typenum`
- Definining a wrapper so any existing `ff::Field` may satisfy `LazyField` (actually allowing the `LazyField` trait to be considered for usage)
- Defining a marker trait for any lazy field with a certain amount of capacity, so code generic to the field may still reduce how often they perform modular reductions
- Implementing `LazyField` for `FieldElement`

Supersedes #787 if and only if this path is preferred. Resolves #389. This new type also _could_ be used internally, ensuring there's no human-error in the existing uses of `FieldElement`, yet that would be a much more intensive PR best left to later.

### DIFF
--- a/curve25519-dalek/src/hazmat/lazy_field25519.rs
+++ b/curve25519-dalek/src/hazmat/lazy_field25519.rs
@@ -65,20 +65,25 @@ impl<CapacityUsed: Unsigned> LazyField<CapacityUsed> for FieldElement<CapacityUs
 #[cfg(test)]
 mod tests {
     use crate::hazmat::lazy_field::{EagerField, LazyField, LazyFieldWithCapacity, Reducible};
-    use typenum::{B1, U2, type_operators::IsLessOrEqual};
+    use typenum::{B1, U2, U3, type_operators::IsLessOrEqual};
 
-    fn add_pair_then_mul<F: LazyFieldWithCapacity<U2>>(
+    fn add_triple_then_mul<F: LazyFieldWithCapacity<U3>>(
         a: F,
         b: F,
         c: F,
         d: F,
+        e: F,
+        f: F,
     ) -> <F as Reducible>::Output
     where
         U2: IsLessOrEqual<F::Capacity, Output = B1>,
+        U3: IsLessOrEqual<F::Capacity, Output = B1>,
     {
         let ab = a.add(&b);
-        let cd = c.add(&d);
-        ab.mul(&cd)
+        let abc = ab.add(&c);
+        let de = d.add(&e);
+        let def = de.add(&f);
+        abc.mul(&def)
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


kayabaNerve:
cc @tarcieri 

My next step would be to work on tests if this is eligible for consideration. In reality, the lazy traits should be upstreamed _somewhere_, but it's probably best to resolve their API first here (with context of usage).

Unfortunately, due to the 32-bit backends, we only get three additions guaranteed (not eight). That may be without issue, as curve formula generally do an add and then a multiply (so only requiring a single addition), but the gap between 32-bit and 64-bit platforms is still notable.

Help is requested regarding the implementation of `Reducible` for `FieldElement`. This PR pushes Rust's type system's bounds evaluation, and for some reason it can't handle, in this context:

```rs
type X = Y;
fn z() -> Y;
```

where `z` is inherited from a trait defined as `fn z() -> Self::X`. As it fails to realize `Self::X == Y`, I had to use a unsafe pointer dereference to coerce the type (as it's immediately obvious the types are the same and therefore this isn't actually a cast).

Also, I'm unsure why the CI is failing. It's expecting a `std` feature and not finding it, yet I didn't touch the features except to add `expose-field`?

tarcieri:
Maybe just call it `hazmat`? I know `expose-field` matches e.g. `k256` but I've considered renaming that. Perhaps it should also be placed under a `hazmat` module.

@kayabaNerve the only build error I'm seeing is:

```
  error: associated function `from_bytes_wide` is never used
     --> curve25519-dalek/src/field.rs:104:19
      |
  101 | impl FieldElement {
      | ----------------- associated function in this implementation
  ...
  104 |     pub(crate) fn from_bytes_wide(bytes: &[u8; 64]) -> Self {
      |                   ^^^^^^^^^^^^^^^
      |
      = note: `-D dead-code` implied by `-D warnings`
      = help: to override `-D warnings` add `#[allow(dead_code)]`
  
  error: could not compile `curve25519-dalek` (lib) due to 1 previous error
```

> As it fails to realize Self::X == Y, I had to use a unsafe pointer dereference 

Can't you just bound appropriately, e.g. `where Self: SomeTrait<X = Y>`?

kayabaNerve:
I didn't realize `expose-field` was in k256 :sweat_smile:  I named this myself. I'm fine using `hazmat`.

The library compiles for me, with `--all-features --all-targets`. The CI is failing re: no-std for some reason. If you look at the implementation of `Reducible`, you'll see an unsafe pointer dereference which is obviously safe, as it's just a `Copy` that's opaque to Rust. If you try to remove it, as it shouldn't be needed, then Rust will fail to compile the library as the type system borks out (hence why I used `unsafe` to perform 'type coercion'. Because Rust can't tell it's the same type). I attempted adding bounds to resolve it, yet the type system seems to refuse to acknowledge an immediately-prior declared type is in fact the immediately-prior declared type in whatever archaic context I constructed.

I'll take a look at the `from_bytes_wide` error though.

kayabaNerve:
And somehow, CI passes now. Amazing. I have no idea what's different.

Renamed to `hazmat`, added test which verifies these constraints do actually produce a usable result.

kayabaNerve:
Fun. The latest commit edited the README and caused a spurious test failure... I'll revisit the bounds on 32-bit platforms. We may only get a single addition :/

curve25519-dalek/src/hazmat/lazy_field25519.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        assert_eq!(
+            LazyField::add(a, &b)
+                .add(&c)
+                .mul(&LazyField::add(d, &e).add(&f))
+                .0,
+            expected
+        );
+        assert_eq!(add_triple_then_mul(a, b, c, d, e, f).0, expected);
+    }
```
tarcieri: If you used `proptest`s for these, we could get the information needed to reproduce CI failures like: https://github.com/dalek-cryptography/curve25519-dalek/actions/runs/17444578081/job/49535734542?pr=816#step:8:205

kayabaNerve:
The docstring says the values are allowed to grow by up to `2^1.75`, which should mean the current bound of `3` is sufficient to prevent an overflow. I'm unsure why a panic was observed accordingly. The bounds stated in the u32 backend may be incorrect. I've decreased the bound to `2`, or just a single addition, for 32-bit backends and increased the runs per unit tests.

curve25519-dalek/src/hazmat/lazy_field25519.rs
kayabaNerve: Will look to set that up next.

kayabaNerve:
Found it. I assumed `from_bytes_wide` would return a reduced result, as `from_bytes` does. It does not. This caused field elements to start at `2`, where of course, any `2` plus any `2` would be `4` (`>` the bound of `3`).

PR #819, opened 2025-09-03
Add `EdwardsPoint::from_affine` (`const`)

### LINKED ISSUE #817: Support uncompressed point deserialization
I recently worked on a project which used a large Bulletproof, leading to a reference string of ~200,000 points. In order to not generate these at runtime, I baked them into my binary with a build script. Unfortunately, at runtime, I must still load them into memory which is only possible via `CompressedEdwardsY::decompress` (or technically `MontgomeryPoint::to_edwards`). The point decompression is actually about as expensive as simply generating them outright (~2x faster, so it's still worth it, but this isn't optimal).

At a minimum, I would ask for `EdwardsPoint::from_x_y(x: [u8; 32], y: [u8; 32]) -> CtOption<Self>`, which could still check the values are on-curve (ensuring the result is well-defined) yet avoids the cost of a square root. This would accelerate my loading from 2x to 40x (as checking the curve formula alone is ~20x faster than a square root). It'd may be preferable to merge #816 and then define `EdwardsPoint::from_x_y(x: FieldElement, y: FieldElement) -> CtOption<Self>`.

w.r.t. a `const fn`, this isn't feasible unless we add a curve formula check _within a const context_. That curve formula check would have to be a lowest-common-denominator backend (64-bit non-fiat arithmetic, accepting the penalty on 32-bit platforms?) or all backends would have to support `const` `add`, `mul`. The fiat backends already do so this wouldn't be impossible, just a bit more invasive if it does require modifying the field implementations.

kayabaNerve:
Builds on #816. Resolves #817.


curve25519-dalek/src/edwards.rs
```diff
@@ -763,6 +763,19 @@ impl EdwardsPoint {
...
+    ///
+    /// Returns `None` if the point is off-curve.
+    ///
+    /// This function runs in variable time.
+    #[cfg(feature = "hazmat")]
+    pub const fn from_affine_coordinates(x: FieldElement, y: FieldElement) -> Option<Self> {
+        let Some(point) = AffinePoint::from_coordinates(x, y) else {
+            return None;
+        };
+        Some(point.to_edwards())
```
tarcieri, line 777:
```suggestion
        AffinePoint::from_coordinates(x, y)?.to_edwards()
```

kayabaNerve: `?` isn't allowed within `const fn`s :/

tarcieri: Oh right, because it needs `From`/`Into`

PR #825, opened 2025-09-05
Optionally implement zerocopy traits on relevant curve25519_dalek and x25519_dalek types

### LINKED ISSUE #821: FR: add (optional) support for zerocopy?
TL;DR: zerocopy is neat, and I think some of the key types in the curve25519-dalek expanded universe should optionally derive its traits. I'm willing to send the patches to make it happen, WDYT?

## Background

The [zerocopy](https://docs.rs/zerocopy/latest/zerocopy/) crate ([github](https://github.com/google/zerocopy)) provides safe zero-copy transmutation facilities. It effectively lets you transmute values to byte sequences and back, with compile-time enforcement that the conversion is safe (e.g. all bit patterns are valid if you derive `FromBytes`, sizes and alignments are correct, ...).

This crate is used a lot in [Fuchsia](https://fuchsia.googlesource.com/), but it's not tied to it, and is quite handy for things like pulling apart network packets. For example, here's a WireGuard handshake initiation message:

```rust
#[repr(C, packed)]
#[derive(TryFromBytes, IntoBytes, Immutable, KnownLayout)]
struct HandshakeInitiation {
    msg_type: MessageType, // a repr(u8) enum
    _reserved: [u8; 3],
    sender_id: zerocopy::byteorder::little_endian::U32,
    ephemeral_pub: [u8; 32],
    static_pub_sealed: [u8; 32 + 16],
    timestamp_sealed: [u8; 12 + 16],
    mac1: [u8; 16],
    mac2: [u8; 16],
}
```

With those derives, I can do things like:

```rust
let mut pkt: Vec<u8> = obtain_packet();
// fails if slice is the wrong size/alignment or msg_type is invalid, otherwise transmutes without copying
let mut pkt = HandshakeInitiation::try_mut_from_bytes(&mut pkt)?;

// Twiddle pkt here...

// Transmute back to &[u8], again with no copying
write_to_network(pkt.as_bytes())
```

The derive macros know which primitive types can be safely transmuted in this way, and has a comprehensive ruleset for proving that composite types meet the trait requirements (e.g. a struct containing an enum field can derive `TryFromBytes` but not `FromBytes`, since not all tag bit patterns are valid).

## Proposal

For the code I'm writing, it would be quite handy for some of the curve25519-dalek key types to implement zerocopy traits. In the example struct above, `ephemeral_pub` should ideally be an `x25519_dalek::PublicKey`, so that I can remove the the extra noise of `PublicKey::from()` and `key.to_bytes()` when touching that field.

So, I propose adding a `zerocopy` crate feature (off by default), and when enabled derive the appropriate zerocopy traits on types that currently support serde serialization. The serde traits already pinpoint those types that are "acceptable to store", so to speak, and zerocopy enables approximately the same thing. So, that seems like a reasonable set of types to make zerocopy-able too.

Taking `x25519_dalek` as an example, that would mean deriving the zerocopy traits on `PublicKey` and `StaticSecret` only. In both cases, I believe (but the derive macros will verify!) that those types bottom out as `[u8; 32]`s with no invalid bit patterns, so they should be able to derive `FromBytes`, `TryFromBytes` and `IntoBytes` to get a complete set of safe byte conversions.

If this sounds like something you'd be okay with, I can do the writing of code and sending of PRs, but I wanted to first see if this is something you'd even want. I believe the maintenance burden would be small since it's not creating new feature surface, but ofc that's easy for me to say as not-the-maintainer :). I could also write up a draft PR to see what the actual change would look like, if that would help decide (and obviously with no hard feelings if the outcome is "no thanks" and throwing the draft away!).

danderson:
Fixes #821

## Overview

The PR has 3 self-contained commits:
 - The main change that does the Cargo.toml paperwork and adds the trait derivations to types which can/should be safely transmutable to byte sequences.
 - Changelog entries for both crates. I kept this one separate because I don't know the project convention: do you add changelog entries when actually doing a release, or do you accumulate changes incrementally into vNext as they show up? I kept the change separate so it's easy to drop if unwanted.
 - Feature table in x25519_dalek's readme. Not strictly related to this change, but I missed having it when I started using the crate, and so figured while I'm in the area... I tried to keep the style in line with the features table in curve25519_dalek, and like in that crate didn't document features whose only purpose is to propagate features into dependencies (e.g. precomputed tables).

## Testing

Tested both modified crates with:

```
cargo test --no-features
cargo test --all-features
cargo test --features=zerocopy
cargo test --no-default-features --features=zerocopy
cargo test --no-default-features --features=zerocopy,alloc
cargo clippy --all-features
```

I didn't add any new tests, since the change is merely deriving new traits that come with their own compile-time correctness checking. If you like, I could add a test that does nothing more than verify that the trait methods are available when the feature's enabled, to avoid a regression where the derives are mistakenly removed?

### DIFF
--- a/x25519-dalek/src/x25519.rs
+++ b/x25519-dalek/src/x25519.rs
@@ -30,6 +30,15 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 /// (in this crate) does *not* automatically happen, but either must be derived
 /// for Drop or explicitly called.
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    feature = "zerocopy",
+    derive(
+        zerocopy::FromBytes,
+        zerocopy::IntoBytes,
+        zerocopy::Immutable,
+        zerocopy::KnownLayout
+    )
+)]
 #[derive(PartialEq, Eq, Hash, Copy, Clone, Debug)]
 pub struct PublicKey(pub(crate) MontgomeryPoint);
 
@@ -198,6 +207,15 @@ impl ZeroizeOnDrop for ReusableSecret {}
 /// implications for many protocols.
 #[cfg(feature = "static_secrets")]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    feature = "zerocopy",
+    derive(
+        zerocopy::FromBytes,
+        zerocopy::IntoBytes,
+        zerocopy::Immutable,
+        zerocopy::KnownLayout
+    )
+)]
 #[derive(Clone)]
 pub struct StaticSecret([u8; 32]);
 
--- a/curve25519-dalek/CHANGELOG.md
+++ b/curve25519-dalek/CHANGELOG.md
@@ -5,6 +5,12 @@ major series.
 
 ## 5.x series
 
+## 5.0.0-pre.2
+
+* Add optional integration with [`zerocopy`][zerocopy], controlled by the `zerocopy` feature.
+
+[zerocopy]: https://docs.rs/zerocopy/latest/zerocopy/
+
 ## 5.0.0-pre.1
 
 * Rename `Scalar::batch_invert` -> `Scalar::invert_batch` for consistency. Also make it no-alloc.
--- a/curve25519-dalek/Cargo.toml
+++ b/curve25519-dalek/Cargo.toml
@@ -34,6 +34,7 @@ features = [
     "digest",
     "legacy_compatibility",
     "group-bits",
+    "zerocopy",
 ]
 
 [dev-dependencies]
@@ -68,6 +69,9 @@ serde = { version = "1.0", default-features = false, optional = true, features =
     "derive",
 ] }
 zeroize = { version = "1", default-features = false, optional = true }
+zerocopy = { version = "0.8", default-features = false, optional = true, features = [
+    "derive"
+] }
 
 [target.'cfg(target_arch = "x86_64")'.dependencies]
 cpufeatures = "0.2.17"
@@ -77,12 +81,13 @@ fiat-crypto = { version = "0.3.0", default-features = false }
 
 [features]
 default = ["alloc", "precomputed-tables", "zeroize"]
-alloc = ["zeroize?/alloc"]
+alloc = ["zerocopy?/alloc", "zeroize?/alloc"]
 precomputed-tables = []
 legacy_compatibility = []
 group = ["dep:group", "rand_core"]
 group-bits = ["group", "ff/bits"]
 digest = ["dep:digest"]
+zerocopy = ["dep:zerocopy"]
 
 [target.'cfg(all(not(curve25519_dalek_backend = "fiat"), not(curve25519_dalek_backend = "serial"), target_arch = "x86_64"))'.dependencies]
 curve25519-dalek-derive = "0.1"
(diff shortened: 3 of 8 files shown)


x25519-dalek/src/x25519.rs
```diff
@@ -198,6 +207,15 @@ impl ZeroizeOnDrop for ReusableSecret {}
...
+    feature = "zerocopy",
+    derive(
+        zerocopy::FromBytes,
+        zerocopy::IntoBytes,
+        zerocopy::Immutable,
+        zerocopy::KnownLayout
+    )
+)]
 #[derive(Clone)]
 pub struct StaticSecret([u8; 32]);
```
tarcieri, line 220: Impls on the public key types are fine, but impls on secrets are slightly worrisome.

One scenario is someone may unintentionally pass the wrong type and could accidentally expose key material because both the public and secret types impl the same trait.

danderson: The same concern applies to the serde traits, no? By definition, StaticSecret is transformable into a form that can be stored, that's why it exists as a separate (riskier) type than ReusableSecret. You could also as_bytes() the key and pass it to anything that handles bytes and the type system won't stop you.

I can see an argument for only allowing serialization via explicit, non-trait methods in order to make it harder to accidentally trigger serialization, but in that case the same argument should apply to the serde derivations?

tarcieri: Yes, FWIW the @RustCrypto crates don't implement the `serde` traits on secrets, though I guess that happened in `curve25519-dalek`

danderson: For my purposes, zerocopy is most useful for serializing data into network packets, and indeed in that case it would be quite bad to serialize a cleartext secret key :)

I'm okay with either outcome, either consistency with the serde traits or treating the serde traits as a historical mistake (I assume we can't remove them without a breaking semver release). In the latter case I can remove the derivations from StaticSecret.

... although tbh that means this entire change, with a new cargo feature and all the plumbing, is effectively to support zerocopy on a single type (x25519 PublicKey), and maybe we're back to the discussion of "is this even worth it" :)

rozbb: Agreed that the serde traits are historical, and prob not something we wanna replicate. And also agreed that it's a tougher pitch if there's just one type that implements zerocopy traits, since it brings in a new feature flag, dependency, and documentation.

I'm leaning towards not pulling zerocopy in until we have more types that'd benefit from it. Wdyt @tarcieri ?

tarcieri: It seems like in `curve25519-dalek` there could be impls on `CompressedEdwardsY` and `CompressedRistretto`

rozbb: Yeah I could get behind that

carloskiki: I would also benefit from this! 

I currently use `[u8; 32]` in types instead of `CompressedEdwardsY` because I need the `zerocopy` traits for serialization. I also don't need these implemented on secret types.

pinkforest: IMO serialization here has more risk than benefits especially given it's just keymaterial that should be dealt with care over implicit serilization as pointed earlier. It also complicates testing and adds more surface area to test for unknown concrete benefit over 32 bytes. I don't think it's worthwhile on the balance to add yet another dependency for something like key that needs to be handled with care over implicit easy to get wrong serialization (or use really and our APIs are mis-use resistant..)

danderson:
Coming back to this: it looks like the sentiment on this is either neutral-ish, or leaning negative for a couple different reasons (crate feature creep; added maintainer burden; making serialization of secret material too transparent/convenient is an antifeature).

So I think that amounts to a "thank you, but no" on this PR and #821 ? Shall I close them?

(To be crystal clear, because text is a lossy medium for emotion: I'm totally fine with this outcome! I was hesitant about this change myself, and if all that comes out of it is a closed bug y'all can point to in future, that's a win in my book :) )

tarcieri:
I'm personally fine with these impls except on secret values, particularly ones intended not to be exfiltrated like `StaticSecret`

rozbb:
I think we'll close for now, since this seems like a large extra dependency/maintenance burden just to expose a few 32-byte fields. Thank you for the PR regardless, I hope to receive more from you in the future!

PR #826, opened 2025-09-06
Implement Lizard encoding/decoding

rozbb:
The Lizard encoding/decoding algorithms specify a map from 16-byte strings to the Ristretto group, and back. This is useful for ElGamal encryption, where plaintexts are group elements.

Signal relies on Lizard, and has been vendoring `curve25519-dalek` for years now (bc we don't export `FieldElement`). I figured upstreaming would be a valuable addition to dalek, so long as it is extensively documented so as to be maintainable. It also doesn't add any new dependencies.

Some highlights of this PR:
* Implements and exposes the Lizard algorithms, as well as some lower-level `map_to_curve` algorithms necessary for downstream. Most of this is from the [Signal repo](https://github.com/signalapp/curve25519-dalek/blob/main/curve25519-dalek/src/lizard/)
* Includes a detailed writeup of the Lizard algorithm, how it works, and how the pseudocode maps to the Rust code
* Upgrades our Sage reference file to Python3 (wow)

This was prepared jointly with @rolfe-signal and with input from @jrose-signal. Thank you all so much for your work on this!


curve25519-dalek/src/lizard/lizard_ristretto.rs
```diff
@@ -0,0 +1,369 @@
...
+
+    /// Interprets the given bytestring as a positive field element and computes the Ristretto
+    /// Elligator map. Note this clears the bottom bit and top two bits of `bytes`.
+    ///
+    /// # Warning
+    ///
+    /// This function does not produce cryptographically random-looking Ristretto points. Use
+    /// [`Self::hash_from_bytes`] for that. DO NOT USE THIS FUNCTION unless you really know what
+    /// you're doing.
+    pub fn map_to_curve(mut bytes: [u8; 32]) -> RistrettoPoint {
```
rozbb: @jrose-signal I've modified this function a little bit from [`from_uniform_bytes_single_elligator`](https://github.com/signalapp/curve25519-dalek/blob/7c6d34756355a3566a704da84dce7b1c039a6572/curve25519-dalek/src/lizard/lizard_ristretto.rs#L26). It does the masking here rather than requiring [the caller](https://github.com/signalapp/libsignal/blob/43a23efa1118ac32a1434ab317025adfa2b91e4a/rust/zkgroup/src/crypto/profile_key_struct.rs#L30) to do it. I think that should simplify your code but let me know if it doesn't

curve25519-dalek/src/lizard/lizard_ristretto.rs
```diff
@@ -0,0 +1,369 @@
...
+        bytes[0] &= 0b11111110;
+        bytes[31] &= 0b00111111;
+
+        let fe = FieldElement::from_bytes(&bytes);
+        RistrettoPoint::elligator_ristretto_flavor(&fe)
+    }
+
+    /// Computes the possible bytestrings that could have produced this point via
+    /// [`Self::map_to_curve`].
+    pub fn map_to_curve_inverse(&self) -> [CtOption<[u8; 32]>; 8] {
```
rozbb: @jrose-signal another API change: I've renamed [`decode_253_bits`](https://github.com/signalapp/curve25519-dalek/blob/7c6d34756355a3566a704da84dce7b1c039a6572/curve25519-dalek/src/lizard/lizard_ristretto.rs#L87C12-L87C27) to this.

curve25519-dalek/src/lizard/lizard_ristretto.rs
jrose-signal: Alas, we have one place where we are not masking, where we are *not* using this as a bidirectional Elligator map but merely as a performance hack for a faster `from_uniform_bytes` (and changing that would be incompatible with existing clients). I wish we didn't, but we do.

curve25519-dalek/src/lizard/lizard_ristretto.rs
jrose-signal: Very reasonable, and definitely a nicer signature. Rolfe and I talked about going even further and returning `[[u8; 32]; 8]` by duplicating present values over absent ones, since there can already be duplicates; however, it's definitely extra work over the basic API.

curve25519-dalek/src/lizard/lizard_ristretto.rs
jrose-signal: While here, I also suggest the doc comment be *very* specific about what "bottom bit" and "top two bits" means; since the argument is just "bytes", a client wouldn't *immediately* think it's going to be used as a little-endian Curve25519 scalar value, and a bytestring doesn't usually have a "top". Though maybe that's at odds with "we (Signal) still need a non-masking version".

rozbb: > Alas, we have one place where we are not masking, where we are _not_ using this as a bidirectional Elligator map but merely as a performance hack for a faster `from_uniform_bytes` (and changing that would be incompatible with existing clients). I wish we didn't, but we do.

Oof ok I think I can make something work. Can you share that code by any chance?

rozbb: > While here, I also suggest the doc comment be _very_ specific about what "bottom bit" and "top two bits" means; since the argument is just "bytes", a client wouldn't _immediately_ think it's going to be used as a little-endian Curve25519 scalar value, and a bytestring doesn't usually have a "top". Though maybe that's at odds with "we (Signal) still need a non-masking version".

Great catch. Will fix

curve25519-dalek/src/lizard/lizard_ristretto.rs
rozbb: I see. I think the explicitness here is nice fwiw

curve25519-dalek/src/lizard/lizard_ristretto.rs
jrose-signal: It is here https://github.com/signalapp/libsignal/blob/43a23efa1118ac32a1434ab317025adfa2b91e4a/rust/zkgroup/src/common/sho.rs#L38

with exactly one caller here https://github.com/signalapp/libsignal/blob/main/rust/zkgroup/src/crypto/profile_key_struct.rs#L47

which needs to be fast because it's in the "try all 64 possibilities for recovering a full 32 bytes loop" here https://github.com/signalapp/libsignal/blob/main/rust/zkgroup/src/crypto/profile_key_encryption.rs#L100

(My non-cryptographer self suspects once we already gave up on hitting the entire curve we could have used a simpler function to produce points from bytes, but it's too late now.)

JobDoesburg:
Any chance on this being merged/released soon?

rozbb:
Yes apologies. I reckon we can merge this week. I need to address Jordan’s points above

curve25519-dalek/src/lizard/lizard_ristretto.rs
rozbb: @jrose-signal Sorry for the delay. I'm trying to come up with a name for the function that doesn't do this masking. Could you help me out? What precisely does it mean to compute the Elligator map on a non-positive field element? [This code](https://github.com/signalapp/libsignal/blob/eb616f63ed053af83e577f36169f5cb5889bb904/rust/zkgroup/src/crypto/profile_key_encryption.rs#L91-L99) sets the bottom bit (and also top bits but let's ignore that for now). But the map, per [our doc](https://github.com/dalek-cryptography/curve25519-dalek/blob/b8d7f99f41f72febe3be82c27018ecd9806e581f/curve25519-dalek/src/lizard/README.md#the-components-of-elligator2) is only defined for positive field elements. Any context on what this function is doing would be helpful. I'm also new to this :)

curve25519-dalek/src/lizard/lizard_ristretto.rs
```diff
@@ -93,11 +91,15 @@ impl RistrettoPoint {
...
         let jcs = self.to_jacobi_quartic_ristretto();
 
-        // Compute the inverse of the every point and its dual
-        let invs = jcs.iter().flat_map(|jc| [jc.e_inv(), jc.dual().e_inv()]);
+        // Compute the positive solution to e⁻¹ on every point and its dual
+        let pos_invs = jcs
+            .iter()
+            .flat_map(|jc| [jc.e_inv_positive(), jc.dual().e_inv_positive()]);
+        // Compute the other solutions to e⁻¹, ie the negatives of the above solutions
+        let neg_invs = pos_invs.clone().map(|mx| mx.map(|x| -&x));
```
jrose-signal: This will repeat work, since Rust iterators are lazy. Better to collect it in an Array first and then do the negation in a second round.

curve25519-dalek/src/lizard/lizard_ristretto.rs
```diff
@@ -178,33 +180,43 @@ impl RistrettoPoint {
...
     /// This function does not produce cryptographically random-looking Ristretto points. Use
     /// [`Self::hash_from_bytes`] for that. DO NOT USE THIS FUNCTION unless you really know what
     /// you're doing.
     pub fn map_to_curve(mut bytes: [u8; 32]) -> RistrettoPoint {
-        // We only have a meaningful inverse if we give Elligator a point in its domain, ie a
-        // positive (meaning low bit 0) field element. Mask off the top two bits to ensure it's less
-        // than the modulus, and the bottom bit for evenness.
-        bytes[0] &= 0b11111110;
-        bytes[31] &= 0b00111111;
+        bytes[31] &= 0b01111111;
```
jrose-signal: This now ought to be redundant with all current implementations of `FieldElement::from_bytes`, but I can see that it might be safer not to depend on that.

curve25519-dalek/src/lizard/lizard_ristretto.rs
rozbb: Alright @jrose-signal I split `map_to_curve` into `map_to_curve` (ie the actual MAP function from the spec), and `map_pos_felem_to_curve`, which does more bit clearing. I also made the inverse return 16 elements, the first 8 of which are positive. 

Is this okay for your use case? I think one issue might be that [`from_uniform_bytes_single_elligator`](https://github.com/signalapp/curve25519-dalek/blob/7c6d34756355a3566a704da84dce7b1c039a6572/curve25519-dalek/src/lizard/lizard_ristretto.rs#L26C12-L26C47) _doesn't even clear the top bit_. In other words, it's not using the MAP function, it's just doing its own thing. Is this load-bearing?

jrose-signal: All implementations of `FieldElement::from_bytes` in practice also clear the top bit, so we're not actually relying on it. (Which is a relief!)

Naming-wise, I do think `pos_felem` is very jargony compared to the other API in the crate, and not even fully accurate since it masks the second-highest bit as well. I'm inclined to think that the masking isn't pulling its weight now that you have the inverse returning 16 elements.

Another variation (take it or leave it): `map_to_curve_invertible`, and rather than *clearing* the top two bits it merely *asserts* that they are clear. More for the caller to write, but also harder for them to forget, and they can choose whether or not to clear the bottom bit as well if they only want to work with positive inverses.

curve25519-dalek/src/lizard/lizard_ristretto.rs
```diff
@@ -0,0 +1,403 @@
...
+    ///     [RFC 9496](https://www.rfc-editor.org/rfc/rfc9496.html#section-4.3.4-4).
+    /// Note this clears the top bit (`bytes[31] & 0x80`).
+    ///
+    /// # Warning
+    ///
+    /// This function does not produce cryptographically random-looking Ristretto points. Use
+    /// [`Self::hash_from_bytes`] for that. DO NOT USE THIS FUNCTION unless you really know what
+    /// you're doing.
+    pub fn map_to_curve(mut bytes: [u8; 32]) -> RistrettoPoint {
+        bytes[31] &= 0b01111111;
```
jrose-signal: This ought to be redundant with all current implementations of `FieldElement::from_bytes`, but I can see that it's cleaner/safer not to rely on that!

tarcieri:
I think the build failures are due to changes between `signature` v3.0.0-rc.3 and v3.0.0-rc.4.

As a general recommendation, I think we should check in Cargo.lock to keep the build deterministic. Otherwise changes in prerelease crates like that will cause random breakages like this in unrelated PRs.

Otherwise I can prepare a PR to do the upgrade.

tarcieri:
Oh wait, `signature` was updated in #828. So this probably just needs a rebase/merge.

curve25519-dalek/src/lizard/lizard_ristretto.rs
rozbb: Ah good catch

curve25519-dalek/src/lizard/lizard_ristretto.rs
rozbb: I agree `pos_felem` is not good. `_invertible` isn't quite right because the only difference is it's 8-to-1 instead of 16-to-1.

How about `map_to_curve_restricted`? Avoiding jargon it says: you give me `input` with these bits cleared, and I ensure you `input` will appear verbatim in the first 8 elements of `map_to_curve_inverse`. Just pushed this change. Lmk what you think

jrose-signal: The `_invertible` part was about the top bits only; I was thinking the bottom-bit-clear guarantee was separate (and not worth checking, just something mentioned in the docs for `map_to_curve_inverse`). But in any case your `map_to_curve_restricted` works for me!

rozbb: Wonderful! Okay let's run with this then. We can update the README to reflect this modification. Thank you for the patience

rozbb:
@jrose-signal Before we merge this (and once the build is working again), do you wanna do a cargo patch for libsignal, do the necessary edits, and see that all tests pass? That way we can keep all the fixes in this branch

curve25519-dalek/Cargo.toml
```diff
@@ -83,6 +83,7 @@ legacy_compatibility = []
 group = ["dep:group", "rand_core"]
 group-bits = ["group", "ff/bits"]
 digest = ["dep:digest"]
+lizard = ["digest"]
```
jrose-signal, line 86: Nitpick: `lizard` doesn't have to depend on `digest`, since the plain `map_to_curve[_inverse]` work on bytestrings. (Arguably, those operations are merely "Elligator 2" and not "Lizard" anyway.)

jrose-signal:
The good news: adoption was straightforward and ran into no unexpected trouble: https://github.com/signalapp/libsignal/pull/636

The bad news:
```
% cargo bench -p zkgroup -- decrypt
Benchmarking decrypt_uuid: Collecting 100 samples in estimated 5.5576 s
decrypt_uuid            time:   [157.25 µs 157.59 µs 157.97 µs]
                        change: [+21.553% +22.019% +22.447%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking decrypt_profile_key: Collecting 100 samples in estimated 5.9268 s
decrypt_profile_key     time:   [599.99 µs 600.79 µs 601.72 µs]
                        change: [+6.9417% +7.1157% +7.2925%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

I'm going to do a bit of profiling and see if there's an obvious culprit—maybe the negations *are* noticeable in the profile, maybe it's something else.

jrose-signal:
Aha, I think that's almost entirely due to the repeated `e_inv_positive()` calls we already discussed. If you push the fix for that I'll run the benchmarks again.

curve25519-dalek/Cargo.toml
rozbb: Yeah, but `lizard_encode` does a hash and embeds it in the field element

```rust
    pub fn lizard_encode<D>(data: &[u8; 16]) -> RistrettoPoint
    where
        D: Digest<OutputSize = U32> + HashMarker,
    {
        let mut fe_bytes: [u8; 32] = Default::default();

        let digest = D::digest(data);
        fe_bytes[0..32].copy_from_slice(digest.as_slice());
        fe_bytes[8..24].copy_from_slice(data);
```

rozbb:
Alright @jrose-signal, I fixed the unnecessary recalculation and rebased the branch on main. Lmk if the benchmarks are improved.

rozbb:
@tarcieri one last organizational thing: what do you think about making `ristretto.rs` into a module and shoving all the elligator stuff in `ristretto/elligator.rs`? We can also expose `map_to_curve` and `map_to_curve_inverse` under `ristretto` instead of `lizard`. Yes they have footguns, but they come with big warnings anyway.

jrose-signal:
We are down to a 2% regression for `decrypt_profile_key` and no change for `decrypt_uuid`, which is a little curious because they both only call `map_to_curve_inverse` once; the one that gets called more in the profile key case is `map_to_curve`, and that one seems a lot closer to the implementation we're currently using. I'll poke at the profiles a little more to see if I spot anything, but I don't think that needs to block the changes here—we're into the point of confounding factors like "it's possible sha2 0.11 is optimizing slightly worse than sha2 0.10 on this system, which wouldn't be ideal but such is life sometimes".

tarcieri:
> what do you think about making ristretto.rs into a module and shoving all the elligator stuff in ristretto/elligator.rs? 

@rozbb sounds fine to me.

> we're into the point of confounding factors like "it's possible sha2 0.11 is optimizing slightly worse than sha2 0.10 on this system, which wouldn't be ideal but such is life sometimes".

@jrose-signal 2% is definitely in the range where very subtle things unrelated to the particular change can be happening, like link order changes, and unless there's an obvious culprit it probably isn't worth worrying about

curve25519-dalek/src/lizard/README.md
```diff
@@ -478,12 +478,12 @@ Hence we see that $(s,t)$ has a preimage if and only if $y$, as defined above, e
...
 
 $$
-r_0 =  (\alpha + \mathsf{sgn}(s)s^2) y
+r_0 =  \pm(\alpha + \mathsf{sgn}(s)s^2) y
 $$
 
 ### Relation to the Formulas in Decaf
 
-These computations can also be seen in [Decaf][decaf], which also defines the map $e: \mathbb{F}_{>0} \rightarrow \mathcal{J}$, but we need to clarify an important difference in notation. In that work, they use variables $a$ and $d$ as parameters for a different, but isogenous, twisted Edwards curve. If we let $\hat{a}$ and $\hat{d}$ denote the Edwards curve parameters from Decaf we have the relations
+These computations can also be seen in [Decaf][decaf], which also defines the map $e: \mathbb{F} \rightarrow \mathcal{J}$, but we need to clarify an important difference in notation. In that work, they use variables $a$ and $d$ as parameters for a different, but isogenous, twisted Edwards curve. If we let $\hat{a}$ and $\hat{d}$ denote the Edwards curve parameters from Decaf we have the relations
```
rozbb, line 486: @rolfe-signal does this commit look good to you? One thing I wasn't sure about: I removed mention of $\mathbb{F}_{>0}$ entirely. I wasn't sure if this line was fully true tho. Where in Decaf does it define $e$, and does it really restrict to the positive field elems? Also its definition of positive is different, so it's a lie no matter what :)

curve25519-dalek/src/lizard/README.md
```diff
@@ -144,7 +144,7 @@ where $\mathrm{i} = \sqrt{-1}$ is the imaginary unit in $\mathbb{F}$.
 
 $E_2$ is built from four functions:
 
-1. A map from positive field elements to a Jacobi quartic: $e: \mathbb{F}_{>0} \rightarrow \mathcal{J}$.
+1. A map from positive field elements to a Jacobi quartic: $e: \mathbb{F} \rightarrow \mathcal{J}$.
```
jrose-signal: "from [positive] field elements"

curve25519-dalek/src/lizard/README.md
```diff
@@ -162,7 +162,7 @@ is an isomorphism.
 
 The full map is $E_2 = q_\mathcal{E} \circ \hat{\theta} \circ q_\mathcal{J} \circ e$. Our goal is to compute the preimage set $E_2^{-1}(P)$ for a given Ristretto point $P \in \mathcal{R}$.
 
-Since $q_\mathcal{E}$ is 2-to-1, $\hat{\theta}$ is 1-to-1, and $q_\mathcal{J}$ is 4-to-1, we expect to find $4 \times 2 = 8$ preimages on the Jacobi quartic $\mathcal{J}$. Each of these 8 points on $\mathcal{J}$ will then correspond to a unique field element via the inverse of $e$. This section explains how to find these 8 field elements.
+Since $q_\mathcal{E}$ is 2-to-1, $\hat{\theta}$ is 1-to-1, and $q_\mathcal{J}$ is 4-to-1, we expect to find $4 \times 2 = 8$ preimages on the Jacobi quartic $\mathcal{J}$. Each of these 8 points on $\mathcal{J}$ will then correspond to two field elements, one postive and one negative, via the inverse of $e$ (thus there is a unique corresponding positive field element via $e^{-1}$). This section explains how to find these 8 field elements.
```
jrose-signal: "16 field elements"

curve25519-dalek/src/lizard/README.md
rolfe-signal: Removal of $\mathbb{F}_{>0}$ everywhere along with describing $e$ as a 2-to-1 map is consistent with the literature and is fine now. The whole reason we restricted to positive field elements in the first place was to try to make this README closely match the code that was there. Now that the code has changed there is no need for it.

JobDoesburg:
Any plans on when this will also be released?

JobDoesburg:
@rozbb any plans on when this will be released?

rozbb:
Should be available now as `5.0.0-pre.2`

Issue #846, opened 2025-11-10
Add `from_pkcs8_pem` or `from_pkcs8_pem` for X25519 static secret


imlk0:
Hi there! 👋

I'm currently working on loading an X25519 private key from a PEM file in PKCS#8 format (e.g., generated by `openssl genpkey -algorithm X25519`), and I noticed that while `ed25519-dalek` provides convenient methods like:

```rust
SigningKey::from_pkcs8_pem(pem_str)?
```

the x25519-dalek crate does not offer similar functionality. This makes interoperability with standard key formats (like those used in TLS, OpenSSL, or configuration files) more cumbersome.

It would be incredibly helpful if `x25519-dalek` could provide built-in support for parsing PKCS#8 PEM/DER, just like `ed25519-dalek` does, e.g.:

```rust
let secret = x25519_dalek::StaticSecret::from_pkcs8_pem(pem_str)?;
// or
let secret = x25519_dalek::StaticSecret::from_pkcs8_der(der_bytes)?;
```

This would greatly improve usability and reduce boilerplate code for users integrating with standard tooling (OpenSSL, Let's Encrypt-style keys, etc.).

Would you be open to adding this feature? I’d be happy to help contribute if needed!

tarcieri:
We can add support for this. It should follow RFC8410:

https://datatracker.ietf.org/doc/html/rfc8410#section-7

You can reuse the implementation from the `ed22519` crate. It's the same encoding. I guess just copy and paste it for now:

https://github.com/RustCrypto/signatures/blob/master/ed25519/src/pkcs8.rs

Note the `AlgorithmIdentifier` OIDs will be different. See Section 9.

PR #847, opened 2025-11-11
feat(x25519-dalek): add PKCS#8 and PEM private key support

imlk0:
Introduce support for encoding and decoding X25519 private keys using the PKCS#8 standard (both DER and PEM formats). This enables interoperability with OpenSSL, TLS stacks, and other crypto systems.

Changes:
- Add `pkcs8` crate as optional dependency
- Implement `EncodePrivateKey` and `TryFrom<PrivateKeyInfoRef>`
- Handle nested OCTET STRING structure required by RFC
- Add tests for DER/PEM encoding round-trips
- Wire up `pem` feature behind feature gate

### DIFF
--- a/x25519-dalek/src/x25519.rs
+++ b/x25519-dalek/src/x25519.rs
@@ -450,7 +450,7 @@ fn to_pkcs8_der(
     Ok(result)
 }
 
-#[cfg(all(feature = "pkcs8"))]
+#[cfg(feature = "pkcs8")]
 impl TryFrom<PrivateKeyInfoRef<'_>> for EphemeralSecret {
     type Error = pkcs8::Error;
 


imlk0:
Hi @tarcieri. This is a implementation for #846. I've written tests, however it seems the current CI setup doesn't seem to test all feature combinations (especially `pkcs8` and `pem`). As a result, these new tests may not be triggered in the CI pipeline. Do you have any advice on how to proceed? Any advice would be welcome!

imlk0:
Just pushed a fix to keep Clippy happy. Thanks for catching that! @tarcieri

tarcieri (approved):

rozbb:
Thank you! Just two things:

1. Can you explain how you generated the PEM/DER test vectors and put them in a comment?
2. Maybe relatedly, why precisely is the `public_key` field in a `PrivateKeyRef` struct encoded as an ASN.1 `BitStringRef`? Why not an octet string, for example? Is this a standard somewhere?

tarcieri:
It’s standard PKCS#8, more specifically https://datatracker.ietf.org/doc/html/rfc5958

rozbb:
Ah, I see the answer to question 2 is that it's specified in [RFC 8410 §7](https://www.rfc-editor.org/rfc/rfc8410.html#section-7). Could you write that in a comment somewhere?

It looks like the PEM is not from the RFC, though. Is that right? Could you change the PEM to the PEM found in the RFC? Also make the DER equal to the translated PEM (and write the command you used for the translation)?

tarcieri:
@imlk0 can you fix the conflicts?

x25519-dalek/src/x25519.rs
```diff
@@ -392,3 +396,98 @@ pub fn x25519(k: [u8; 32], u: [u8; 32]) -> [u8; 32] {
...
+/// X25519 Algorithm Identifier.
+#[cfg(feature = "pkcs8")]
+pub const ALGORITHM_ID: pkcs8::AlgorithmIdentifierRef<'static> = pkcs8::AlgorithmIdentifierRef {
+    oid: ALGORITHM_OID,
+    parameters: None,
+};
+
+#[cfg(all(feature = "alloc", feature = "pkcs8"))]
+impl EncodePrivateKey for EphemeralSecret {
+    fn to_pkcs8_der(&self) -> Result<SecretDocument, pkcs8::Error> {
```
pinkforest, line 416: This should not be exported out as it violates perfect forward secrecy.

We made a point not to add any impls that allows serialization out.

x25519-dalek/src/x25519.rs
```diff
@@ -392,3 +396,98 @@ pub fn x25519(k: [u8; 32], u: [u8; 32]) -> [u8; 32] {
...
+    private_key.zeroize();
+
+    Ok(result)
+}
+
+#[cfg(feature = "pkcs8")]
+impl TryFrom<PrivateKeyInfoRef<'_>> for EphemeralSecret {
+    type Error = pkcs8::Error;
+
+    fn try_from(private_key: PrivateKeyInfoRef<'_>) -> Result<Self, pkcs8::Error> {
```
pinkforest, line 457: `EphemeralSecret` is supposed to be random generated always.

This should not be extended to it.

pinkforest:
`EphemeralSecret` purpose is to ensure no serialization out to guarantee perfect forward secrecy.

Adding any type of serialization to it will break this property and also breaks the property of ensuring randomized keys for it.

It's fine for `StaticSecret` but it shouldn't be for Ephemeral.

See documentation for `EphemeralSecret`:
> This type is identical to the StaticSecret type, except that the [EphemeralSecret::diffie_hellman](https://docs.rs/x25519-dalek/latest/x25519_dalek/struct.EphemeralSecret.html#method.diffie_hellman) method consumes and then wipes the secret key, **and there are no serialization methods defined**. This means that [EphemeralSecret](https://docs.rs/x25519-dalek/latest/x25519_dalek/struct.EphemeralSecret.html)s **can only be generated from fresh randomness where the compiler statically checks that the resulting secret is used at most once**.

PR #848, opened 2025-11-17
perf: use width-8 NAF for static scalars to match NafLookupTable8

GarmashAlex:
The precomputed Straus implementations for both the vector and serial backends build NafLookupTable8 for static points, which stores odd multiples up to 127A, but they decomposed static scalars with a width-5 NAF, restricting lookups to 1..15 and leaving most of the table unused. This change switches static scalar decomposition to non_adjacent_form(8), fully utilizing the 64-entry odd table. It reduces the number of static additions due to the lower nonzero digit density of width-8 NAF while keeping the number of doublings and the memory footprint unchanged, since the larger table was already computed. The change is safe because Scalar::non_adjacent_form supports w<=8 and NafLookupTable8::select requires odd digits less than 128, and it matches the existing pattern in vartime double-base multiplication where width 8 is used when a large odd-multiples table is available. No API or behavioral changes beyond this decomposition width are introduced.

### DIFF
--- a/curve25519-dalek/src/backend/vector/scalar_mul/precomputed_straus.rs
+++ b/curve25519-dalek/src/backend/vector/scalar_mul/precomputed_straus.rs
@@ -80,7 +80,7 @@ pub mod spec {
         {
             let static_nafs = static_scalars
                 .into_iter()
-                .map(|c| c.borrow().non_adjacent_form(5))
+                .map(|c| c.borrow().non_adjacent_form(8))
                 .collect::<Vec<_>>();
             let dynamic_nafs: Vec<_> = dynamic_scalars
                 .into_iter()


rozbb:
Great catch, ty! This looks fine to me but I'm not the most familiar with how NAF is used in our backends. @tarcieri wanna be another set of eyes?

rozbb:
Ok, I was wondering how it's even possible that using a width-5 NAF representation even works with a `NafLookup8`. The answer is that the `NafLookup8` stores strictly more information than a `NafLookup5`, and the overlap is in exactly the same indices. Therefore, the width-5 decomposition of the static scalars still produces indices that are compatible with the widht-8 table `select()` procedure.

Also, the benchmarks here show huge improvements. 30% improvement far large mixed-base multiscalar multiplications. And the binary size is no different, as expected.

tarcieri (approved):

PR #851, opened 2025-11-25
[WIP] curve: enable the `unreachable_pub` lint

tarcieri:
In #850 I noticed there are a lot of places that internal types are currently marked `pub` which is a bit confusing when trying to reason about whether APIs allow for violating invariants.

This enables the `unreachable_pub` lint which can automatically check for such situations. Most of them were corrected automatically using `cargo fix`


tarcieri:
Note: I only ran `cargo fix` on the 64-bit soft backend. There's so much conditional compilation it might be a bit difficult to track down all the places this lint is being violated.

tarcieri:
Per #855 I guess this shouldn't apply to anything in the `backend` module since that contains documentation which is published to docs.rs (or at least, we should wait for that to be merged so we can test docs.rs rendering)

PR #858, opened 2025-12-10
[FEAT] implement three scalar multiplication

zz-sol:
This is the first PR for enabling hEEA verification algorithm. See the discussions here  #856 

This PR has three commits for the ease of reviewing. 

The first commit is purely adding constants. The last commit is for benchmark and testing.
a0a6b03eb2da1c6b30a0ec44d5850152f22a97d3 is the main algorithm that is been implementated.


curve25519-dalek/Cargo.toml
```diff
@@ -61,6 +61,10 @@ serde = { version = "1.0", default-features = false, optional = true, features =
     "derive",
 ] }
 zeroize = { version = "1", default-features = false, optional = true }
+# temp use ethnum for u256/i256 which is more stable
+# a hand optimized i256 implementation (zz/impl_ches25_without_ethnum) is a little faster for heea computation
+# happy to merge those code if desired
+ethnum = "1.5"
```
tarcieri: Would definitely be good to get rid of this dependency

tarcieri:
Regarding tests, a `proptest` that `vartime_triple_scalar_mul_basepoint` is equivalent to either the raw operations on points/scalars or `Vartime(Precomputed)MultiscalarMul` sounds useful

curve25519-dalek/Cargo.toml
zz-sol: https://github.com/zz-sol/curve25519-dalek/pull/2/files

here is the PR.

tarcieri: Can you go ahead and merge it in so it can be reviewed in-band with this PR?

It looks like it also has performance benefits along with eliminating a dependency.

zz-sol: merged https://github.com/dalek-cryptography/curve25519-dalek/pull/859/commits/65b6374685a9d61e4d1a0c9c6ca084b5f679df11

zz-sol:
Hi gently ping to see if there is any updates on this PR?

tarcieri:
I still need to read the paper

tarcieri:
@zz-sol with `curve25519-dalek` v5.0.0 released we're probably in a better position to start looking at this, though I still have a ton of things on my plate.

I've now skimmed [the paper](https://tches.iacr.org/index.php/TCHES/article/view/12225) and done something superficially/structurally similar in `k256` which leverages the GLV endomorphism to decompose into two half-width scalars prior to wNAF/Straus multiscalar multiplication. hEEA seems like it has many similarities to that approach, although not leveraging an endomorphism.

I'll also note we have #323 open which leverages Pornin's lattice reduction using Lagrange's Algorithm, which is referred to in the hEEA paper. As I noted in #323 itself I think this work obsoletes that PR while also being smaller and therefore easier to review.

This PR actually ends up being smaller than it looks at first glance since a bulk of the additions are to precomputed lookup tables, so that's nice. Perhaps @rozbb can do another pass too.

I guess my immediate concern is testing: it doesn't seem like there are tests? If so I glossed over them. Perhaps you were expecting #859 to act as an integration test? It would be really nice to have some kind of unit tests as well.

In testing things like multiscalar multiplications in @RustCrypto crates I tend to use `proptest` and compare to equivalent single-scalar operations. I know quite a bit of formal verification went into the `curve25519-dalek` v5 release and it sure would be nice if that work could be extended to this PR as well, though I don't see that as a blocker for landing.

Notably it would be nice to see proptests of each backend that compare them to the equivalent set of single-scalar operations.

zz-sol:
> I guess my immediate concern is testing: it doesn't seem like there are tests? If so I glossed over them. Perhaps you were expecting https://github.com/dalek-cryptography/curve25519-dalek/pull/859 to act as an integration test? It would be really nice to have some kind of unit tests as well.


there were dedicated commitment for the testing code in this PR
- [unit test](https://github.com/dalek-cryptography/curve25519-dalek/pull/858/changes/8da112bca5ef2eedd8fd8d99027e787c49573814)
- [porp test](https://github.com/dalek-cryptography/curve25519-dalek/pull/858/changes/36819893d3b596197d6ecbd417622e798ead3e03)

the `porptest` is only available for serial backend though. I will add the rest backends.

This PR is for the MSM only. So the integration rest etc. are in follow up PRs. see https://github.com/dalek-cryptography/curve25519-dalek/pull/859

tarcieri:
Oh sorry, seems I glossed over that. Looks good for the serial backend, so if you add similar tests for the other backends I think we can do a more detailed review and get this merged.

zz-sol:
> Oh sorry, seems I glossed over that. Looks good for the serial backend, so if you add similar tests for the other backends I think we can do a more detailed review and get this merged.

Done.

curve25519-dalek/src/backend/vector/scalar_mul/vartime_triple_base.rs
```diff
@@ -0,0 +1,238 @@
...
+// Copyright (c) 2016-2019 Henry de Valence
+// See LICENSE for licensing information.
+
+#![allow(non_snake_case)]
+
+#[curve25519_dalek_derive::unsafe_target_feature_specialize(
+    "avx2",
+    conditional(
+        "avx512ifma,avx512vl",
+        all(curve25519_dalek_backend = "unstable_avx512", nightly)
```
tarcieri: We stabilized this in #913 


```suggestion
        curve25519_dalek_backend = "avx512"
```

tarcieri:
@zz-sol I tried to address the merge conflicts but some more work is needed

zz-sol:
> @zz-sol I tried to address the merge conflicts but some more work is needed

Oh yeah the RNG api has changed since the upgrade. It is now fixed and my CI is green

zz-sol:
gently ping on this

tarcieri:
Sorry, I'm way behind on reviews, will look next week

tarcieri (approved):
Would appreciate some additional review from @rozbb but otherwise this seems like a promising start to me

curve25519-dalek/src/backend/serial/scalar_mul/vartime_triple_base.rs
```diff
@@ -0,0 +1,323 @@
...
+pub fn mul_128_128_256(
+    a1: &Scalar,
+    A1: &EdwardsPoint,
+    a2: &Scalar,
+    A2: &EdwardsPoint,
+    b: &Scalar,
+) -> EdwardsPoint {
+    // assert that a1 and a2 are less than 2^128
+    debug_assert!(a1.as_bytes()[16..32].iter().all(|&b| b == 0));
+    debug_assert!(a2.as_bytes()[16..32].iter().all(|&b| b == 0));
```
tarcieri: One thing I guess I'm a little unclear on: the comment makes it sound like passing an unreduced scalar is an optimization problem, but it will cause a panic in debug mode due to the `debug_assert!`s. What happens when a non-canonical scalar is passed in a release build?

rozbb: Agree “this function is optimized for” is not strong enough. Should have a “panics” section

curve25519-dalek/src/backend/serial/scalar_mul/vartime_triple_base.rs
```diff
@@ -0,0 +1,323 @@
...
+            Ordering::Equal => {}
+        }
+
+        // Add contributions from b_hi*B' where B' = B * 2^128
+        match b_hi_naf[i].cmp(&0) {
+            Ordering::Greater => t = &t.as_extended() + &table_B_128.select(b_hi_naf[i] as usize),
+            Ordering::Less => t = &t.as_extended() - &table_B_128.select(-b_hi_naf[i] as usize),
+            Ordering::Equal => {}
+        }
+
```
rozbb, line 100: I think all the above repetition should be consolidated somewhere. A function/lambda if possible, but I’d be okay w a macro too

curve25519-dalek/src/backend/serial/scalar_mul/vartime_triple_base.rs
zz-sol: Good catch — with `debug_assert!` right now a non-canonical `a1` and `a2` would silently truncate to the low 128 bits and produce a wrong point.

I think we can do the following

- Either swap `debug_assert!` for `assert!` -- this makes the code a bit slower.
- Or make this function `pub(crate)`, since the only caller is the `b_lo` and `b_hi` decomposition internally, which already guarantees both halves are < 2^128 by construction.

We can also do both. My preference is just do the `pub(crate)`.

curve25519-dalek/src/backend/vector/scalar_mul/vartime_triple_base.rs
```diff
@@ -0,0 +1,235 @@
...
+            match b_hi_naf[i].cmp(&0) {
+                Ordering::Greater => {
+                    Q = &Q + &table_B_128.select(b_hi_naf[i] as usize);
+                }
+                Ordering::Less => {
+                    Q = &Q - &table_B_128.select(-b_hi_naf[i] as usize);
+                }
+                Ordering::Equal => {}
+            }
+
```
rozbb, line 125: Ditto here, could you reduce the repetition if possible?

curve25519-dalek/src/backend/serial/scalar_mul/vartime_triple_base.rs
rozbb: Actually, perhaps an even stronger property: why not expose a new type: `HalfWidthScalar`? It wouldn't have to implement the usual `Scalar` methods. Just a simple constructor that ensures that it's within bounds. This lets us remove the assert and the `panics` condition

rozbb (commented):
Looks good! Some small things here and there. Happy to have this land soon

rozbb:
The nightly errors are unrelated to this PR. Need to update `cpufeatures` to 0.3.1 once that drops.

I'll merge this now though. Thank you very muhc @zz-sol !

I see a 29% speedup between the two new benchmarks

zz-sol:
> The nightly errors are unrelated to this PR. Need to update cpufeatures to 0.3.1 once that drops.
> 
> I'll merge this now though. Thank you very muhc @zz-sol !
> 
> I see a 29% speedup between the two new benchmarks

Thanks! This is great to see. 

next step I will clean up #859 on top of this

PR #860, opened 2025-12-27
Bump `rand_core` to v0.10.0-rc-3

tarcieri:
Most of the changes in this PR are actually from the associated `rand` crate updates which are happening in rust-random/rand#1697, notably `OsRng` has been renamed to `SysRng` (and is now provided by the `getrandom` crate).

We do use some `rand_core` APIs in a few places though, like the STROBE implementation, where this migrates from `rand_core::le` to `rand_core::utils`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,3 +14,7 @@ opt-level = 2
 curve25519-dalek-derive = { path = "./curve25519-dalek-derive" }
 curve25519-dalek = { path = "./curve25519-dalek" }
 x25519-dalek = { path = "./x25519-dalek" }
+
+# https://github.com/rust-random/rand/pull/1697
+rand = { git = "https://github.com/rust-random/rand", rev = "99ae06fb348e5c5fee0fe41d2201bd4cbb107635" }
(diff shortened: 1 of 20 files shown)


Cargo.toml
```diff
@@ -14,3 +14,7 @@ opt-level = 2
 curve25519-dalek-derive = { path = "./curve25519-dalek-derive" }
 curve25519-dalek = { path = "./curve25519-dalek" }
 x25519-dalek = { path = "./x25519-dalek" }
+
+# https://github.com/rust-random/rand/pull/1697
+rand = { git = "https://github.com/rust-random/rand", rev = "99ae06fb348e5c5fee0fe41d2201bd4cbb107635" }
+getrandom = { git = "https://github.com/rust-random/getrandom.git", rev = "f67c70110f706f80d7dd6fa52e38d0920754fe2c" }
```
tarcieri, line 20: Could go either way on waiting for these to be merged. Right now the build is broken on `main`, so I feel like getting it fixed would probably be for the best, even if we temporarily have some patches to do it.

rozbb: Agree

rozbb (approved):
Thank you for catching this!

pinkforest:
Needs a release still, getting some conflicts (through aead's) between pinned rc's.

PR #863, opened 2026-01-03
curve,ed,x: Bump prerelease version to fix build

rozbb:


curve25519-dalek/CHANGELOG.md
```diff
@@ -5,10 +5,16 @@ major series.
...
+
+* Add allocation-free `EdwardsPoint::compress_batch()`
+* Perf: Use maximum available NAF window size in `VartimePrecomputedStraus`
+* Upgrade `rand_core` dependency to v0.10.0-rc-3
+
 ## 5.0.0-pre.3
 
 * Add Lizard bytes-to-point injection for Ristretto. Gated under `lizard` feature.
-* Upgrade `rand_core` to v0.10.0-rc-2
+* Upgrade `rand_core` dependency to v0.10.0-rc-2
```
tarcieri, line 17: Hmm, guess we can leave notes for each prerelease like this but I think for a final release we should boil them all down to just the v5.0.0 release

tarcieri (approved):

curve25519-dalek/CHANGELOG.md
rozbb: Agreed, that was my plan

Issue #758, opened 2025-05-26
AVX-512 support on `stable`


tarcieri:
Per #757, `avx512_target_feature` will soon be stable on Rust 1.89. See also FCP for AVX-512 support: https://github.com/rust-lang/rust/issues/111137#issuecomment-2888077126

Given that, it will soon be possible to support AVX-512 on `stable` rather than as a `nightly`-only feature. With the target feature stable, any unstable intrinsics can be worked around by using inline ASM, which can be implemented in the form of "polyfills" which look identical to the unstable intrinsics. Though hopefully when that FCP completes, the relevant intrinsics will be stable soon after.

Issue #853, opened 2025-11-30
Snapshot cfg permutations from simd backends


pinkforest:
Currently we assume through Github CI that all the various cfg-permutations get touched and tested at some level.

This despite that github does not guarantee either what target arch / cpu features are available.

Wondering first if it would be feasible to check the codegen from these intrinistics so it stays as the expected / previously recorded / tested.

Secondly I would generally scan all the cfg-gates and maybe run some type of sanity / validation that the stuff under gets covered in the CI test-matrix.

This would catch any toolchain / build configurations where e.g. some bug deep down gets triggered only through some build gating and where the said code was not covered / tested etc.

ref: https://github.com/cryspen/libcrux/issues/1255 (there was a bug hidden through a specific target_feature)

Issue #744, opened 2025-04-24
derive: get rid of #[unsafe_target_feature]?


tarcieri:
Coming from #743 

It seems the main use of `unsafe_target_feature` is to support applying the `target_feature` attribute to safe functions.

This is now supported by stable rustc as of Rust 1.86, so is `unsafe_target_feature` even still needed?

PR #866, opened 2026-01-09
support avx512 without nightly

Slixe:
Remove the requirements of nightly for AVX512 in `build.rs` & `#[cfg]`

Solve #758

### DIFF
--- a/.github/workflows/curve25519-dalek.yml
+++ b/.github/workflows/curve25519-dalek.yml
@@ -130,7 +130,7 @@ jobs:
       run: cargo test --no-default-features --features alloc,precomputed-tables,zeroize,group-bits --target x86_64-unknown-linux-gnu
 
   msrv:
-    name: Current MSRV is 1.85.0
+    name: Current MSRV is 1.89.0
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
@@ -139,7 +139,7 @@ jobs:
     - run: cargo update -Z minimal-versions
     # Now check that `cargo build` works with respect to the oldest possible
     # deps and the stated MSRV
-    - uses: dtolnay/rust-toolchain@1.85.0
+    - uses: dtolnay/rust-toolchain@1.89.0
     - run: cargo build --no-default-features --features serde
     # Also make sure the AVX2 build works
     - run: cargo build --target x86_64-unknown-linux-gnu
--- a/.github/workflows/curve25519-dalek.yml
+++ b/.github/workflows/curve25519-dalek.yml
@@ -130,7 +130,7 @@ jobs:
       run: cargo test --no-default-features --features alloc,precomputed-tables,zeroize,group-bits --target x86_64-unknown-linux-gnu
 
   msrv:
-    name: Current MSRV is 1.89.0
+    name: Current MSRV is 1.85.0
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
@@ -139,7 +139,7 @@ jobs:
     - run: cargo update -Z minimal-versions
     # Now check that `cargo build` works with respect to the oldest possible
     # deps and the stated MSRV
-    - uses: dtolnay/rust-toolchain@1.89.0
+    - uses: dtolnay/rust-toolchain@1.85.0
     - run: cargo build --no-default-features --features serde
     # Also make sure the AVX2 build works
     - run: cargo build --target x86_64-unknown-linux-gnu
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -85,7 +85,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: dtolnay/rust-toolchain@1.89.0
+    - uses: dtolnay/rust-toolchain@1.87.0
       with:
         components: clippy
     - run: cargo clippy --target x86_64-unknown-linux-gnu --all-features
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -85,7 +85,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: dtolnay/rust-toolchain@1.87.0
+    - uses: dtolnay/rust-toolchain@1.89.0
       with:
         components: clippy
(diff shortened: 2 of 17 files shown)


Slixe:
this would actually require to set 1.89 as MSRV.

tarcieri:
Ideally it'd be nice to get a release out with a 1.85 MSRV so it could get packaged on Debian stable.

@Slixe perhaps it could still require an explicit `cfg` opt-in, to avoid the MSRV bump?

We could then later bump MSRV in a point release and remove it.

Slixe:
@tarcieri could it work if we simply enable the auto detection for avx512 >= 1.89 only from `build.rs` ?

tarcieri:
Yeah, you could temporarily add rust version detection until we bump MSRV

tarcieri:
@Slixe seems like you can remove draft?

Slixe:
I'm still having an issue, I'm not sure how the `conditional` works.

```log
Compiling criterion v0.5.1        
warning: unexpected `cfg` condition value: `avx512ifma,avx512vl`      
  --> curve25519-dalek/src/backend/vector/scalar_mul/pippenger.rs:12:1
   |                                                                                             
12 | / #[curve25519_dalek_derive::unsafe_target_feature_specialize(
13 | |     "avx2",                
14 | |     conditional("avx512ifma,avx512vl", curve25519_dalek_backend = "avx512")
   | |_____________________________________^
   |                                                                                             
   = note: expected values for `target_feature` are: `10e60`, `2e3`, `32s`, `3e3r1`, `3e3r2`, `3e3r3`, `3e7`, `7e10`, `a`, `aclass`, `adx`, `aes`, `altivec`, `alu32`, `amx-avx512`, `amx-bf16`, `
amx-complex`, `amx-fp16`, `amx-fp8`, `amx-int8`, `amx-movrs`, `amx-tf32`, `amx-tile`, `amx-transpose`, `apxf`, `atomics`, `avx`, `avx10.1`, `avx10.2`, `avx2`, `avx512bf16`, `avx512bitalg`, `avx5
12bw`, `avx512cd`, and `avx512dq` and 429 more
   = note: using a cfg inside a attribute macro will use the cfgs from the destination crate and not the ones from the defining crate
   = help: try referring to `curve25519_dalek_derive::unsafe_target_feature_specialize` crate for guidance on how handle this unexpected cfg
   = help: the attribute macro `curve25519_dalek_derive::unsafe_target_feature_specialize` may come from an old version of the `curve25519_dalek_derive` crate, try updating your dependency with 
`cargo update -p curve25519_dalek_derive`   
   = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
   = note: requested on the command line with `-W unexpected-cfgs`
   = note: this warning originates in the attribute macro `curve25519_dalek_derive::unsafe_target_feature_specialize` (in Nightly builds, run with -Z macro-backtrace for more info)
                                                
```

rustc 1.91.1

EDIT: it seems `target_feature` support one by one only, and the macro don't split it correctly?

tarcieri:
I think we could probably completely get rid of `curve25519-dalek-derive`, although it might require a bump to Rust 1.86: https://github.com/dalek-cryptography/curve25519-dalek/issues/744

Slixe:
I believe this new commit should be good enough?
avx512 backend is only enabled if both features are available anyway.
and this prevent the bump of MSRV as you asked

Slixe:
@tarcieri the avx512 backend may be available & detected but won't work if the target features aren't included for compilation.

which option would you prefer?
- enables it in `build.rs` only when available


- in the code, add the requirement of target features along the actual avx512 backend check like here:
```rust
    #[cfg(curve25519_dalek_backend = "avx512")]
    Avx512ifma(
```

tarcieri:
@Slixe I'm not sure I quite understand the problem. If all the AVX-512 code is feature-gated with `#[cfg(curve25519_dalek_backend = "avx512")]`, then it shouldn't be an issue to declare `#[(unsafe_)target_feature("avx512*")]` immediately after that.

If the code isn't compiled with `#[cfg(curve25519_dalek_backend = "avx512")]`, then conditional compilation should ensure that none of that code is included.

`build.rs` can detect when the rustc version and target architecture are compatible and set `curve25519_dalek_backend = "avx512"` accordingly.

Then `cpufeatures` is already configured to use CPUID to detect if the features are enabled on the target CPU at runtime, I believe.

pinkforest:
There is no way to test this in CI. How / where does this get tested and who ensures it gets done if not in CI long run ?

Especially if this is default-on backend that should be tested in CI... ideally..

Even if the backend doesn't change the stuff around it just might and might be hard to catch errors there.

Or even how gating errors can lead to trouble e.g.:
- https://github.com/cryspen/libcrux/issues/1220

tarcieri:
I believe some of the GitHub Actions runners have AVX-512 support. If CPU feature detection is enabled, it should at least opportunistically test when the runner has support. Not a great answer, but it is what it is:

https://github.com/actions/runner/issues/1069

Otherwise they suggest setting up a self-hosted runner. Intel SDE is also an option, I believe.

pinkforest:
> I believe some of the GitHub Actions runners have AVX-512 support.

They used to have that with intel - which we relied on earlier - to but not anymore and they have no AMD AVX512 either.

As I suggested in libcrux, should probably maybe prehaps validate that we test all the gating gi matrix as well
- https://github.com/dalek-cryptography/curve25519-dalek/issues/853

Anyways I would not turn it on by default - not because the backend hasn't changed but more that the gating is quite involved easy to mess up and it's not the first time it has caused trouble elsewhere and we can't test it.

tarcieri:
This PR to add AVX-512 support to `chacha20` includes a CI config that uses Intel SDE as the runner: https://github.com/RustCrypto/stream-ciphers/pull/477/files#diff-4406b424fd59f8f997267f1955fac9159be3d9ba9cde8f6b6b8a53c528edcba9

nazar-pc:
@Slixe will you be able to finish this any time soon? I bet you can generate necessary CI testing configuration for AVX512 using an example with LLMs quite easily.

tarcieri:
FWIW @RustCrypto has an install action for Intel SDE here: http://github.com/RustCrypto/actions/blob/master/intel-sde-install/action.yaml

nazar-pc:
I opened https://github.com/dalek-cryptography/curve25519-dalek/pull/913 with a few more fixes and AVX512 CI testing added.

PR #873, opened 2026-01-24
Bump `rand_core` to v0.10.0-rc-6

tarcieri:
This also rips out `rand`: it's just acting as a facade for what we're using that complicates these sorts of upgrades.

### DIFF
--- a/ed25519-dalek/src/lib.rs
+++ b/ed25519-dalek/src/lib.rs
@@ -22,12 +22,12 @@
 #![cfg_attr(not(feature = "rand_core"), doc = "```ignore")]
 //! # fn main() {
 //! // $ cargo add ed25519_dalek --features rand_core
-//! use rand::rngs::SysRng;
-//! use rand_core::TryRngCore;
+//! use getrandom::{SysRng, rand_core::UnwrapErr};
+//! use rand_core::TryRng;
 //! use ed25519_dalek::SigningKey;
 //! use ed25519_dalek::Signature;
 //!
-//! let mut csprng = SysRng.unwrap_err();
+//! let mut csprng = UnwrapErr(SysRng);
 //! let signing_key: SigningKey = SigningKey::generate(&mut csprng);
 //! # }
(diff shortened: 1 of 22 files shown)


ed25519-dalek/src/lib.rs
```diff
@@ -22,12 +22,12 @@
...
-//! use rand::rngs::SysRng;
-//! use rand_core::TryRngCore;
+//! use getrandom::{SysRng, rand_core::UnwrapErr};
+//! use rand_core::TryRng;
 //! use ed25519_dalek::SigningKey;
 //! use ed25519_dalek::Signature;
 //!
-//! let mut csprng = SysRng.unwrap_err();
+//! let mut csprng = UnwrapErr(SysRng);
 //! let signing_key: SigningKey = SigningKey::generate(&mut csprng);
```
tarcieri, line 31: FYI @rozbb, the reason we might consider supporting `TryCryptoRng` is it's the trait that `SysRng` impls, and if we had first-class support for it there wouldn't be any need for this annoying `UnwrapErr`.

Another solution to the same problem is what we did in @RustCrypto which was to give the APIs that use an `*Rng` parameter names like `*_with_rng` and for the short name, make a `getrandom`-based API that passes `SysRng` for you and doesn't have any parameters at all, like `SigningKey::generate()`

rozbb: I like the second one a lot! It makes the common case ("generate this, idc how") fast and easy, and it makes alternate cases ("use this RNG") still possible. And if someone really wants to use a specific fallible RNG, they are free to seed an infallible RNG with the fallible one.

rozbb:
Looks good! To prevent #871 from happening again, do you think it makes sense to check in Cargo.lock? Or maybe just add the `=` prefix to rc deps for now?

tarcieri:
I'm a big fan of checking in Cargo.lock in general although the only way to fix #871 is merge this and get another release out

PR #882, opened 2026-02-10
perf(lizard): skip negative preimages in lizard_decode

GarmashAlex:
lizard_decode iterated over all 16 Elligator2 preimages (8 positive + 8 negative), but lizard_encode always produces a positive field element (LSB cleared). Negative preimages have LSB=1 and can never pass the tag check — the comparison is guaranteed to fail. The Go reference implementation (go-ristretto) correctly iterates only 8 times.
Limited the loop to the first 8 (positive) preimages, removing 8 wasted SHA-256 computations per decode call.


nzengi:
This PR's key observation (negative preimages can never pass TagCheck) is formally justified in PR #915, which adds a proof sketch to `lizard_decode` explaining why the bound is exactly **7 · 2^-125 < 2^-122** in the ROM.

The proof shows: any other positive preimage `fe[j]` with the same payload bytes[8..24] as the original UUID would be forced to equal `fe_uuid` by the tag check — contradiction. So only preimages with *distinct* payloads can reach the SHA256 comparison, and each independently fails with probability 1 - 2^-125. Union bound over the 7 remaining positive preimages gives the tight bound.

PR #915 also documents the algebraic identity `s0 · s1 = -1` in GF(p) for the first Jacobi quartic pair (derivable from the `to_jacobi_quartic_ristretto` source), which may be of independent interest.

rozbb:
Thank you for this, and sorry for taking so long. Just added some docs here and there

PR #884, opened 2026-02-14
curve: Impl `WnafGroup` for `EdwardsPoint`

rozbb:
This is a hint function. I don't believe the hint has to be very good. I think it's more important that it just be implemented. What do you think @tarcieri @mmaker ?

### DIFF
--- a/curve25519-dalek/src/edwards.rs
+++ b/curve25519-dalek/src/edwards.rs
@@ -2282,6 +2282,41 @@ mod test {
         assert!(P1.compress().to_bytes() == P2.compress().to_bytes());
     }
 
+    /// Test that the group WNAF implementation works for random scalars.
+    #[cfg(all(feature = "group", feature = "alloc"))]
+    #[test]
+    fn wnaf_group() {
+        let mut rng = UnwrapErr(SysRng);
+        let num_scalars = 100;
+
+        // Run our tests on a random basepoint and the canonical basepoint
+        let basepoints = {
+            let s = Scalar::random(&mut rng);
+            [
+                constants::ED25519_BASEPOINT_POINT,
+                s * constants::ED25519_BASEPOINT_POINT,
+            ]
+        };
+
+        for basepoint in basepoints {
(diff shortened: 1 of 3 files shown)


mmaker:
thanks! looks good to me

tarcieri:
Have you tried actually using the wNAF implementation? Is it computing correct results?

rozbb:
Just wrote a test comparing `group::Wnaf` results to our multiplication. Passes no problem. I’ll commit it tomorrow.

curve25519-dalek/src/edwards.rs
```diff
@@ -1510,6 +1510,17 @@ impl GroupEncoding for EdwardsPoint {
...
 }
 
+#[cfg(all(feature = "group", feature = "alloc"))]
+impl group::WnafGroup for EdwardsPoint {
+    /// A hint for how large the window should be to do multiplication with a given number of
+    /// scalars
+    fn recommended_wnaf_for_num_scalars(_num_scalars: usize) -> usize {
+        // Just return 5 for now. This is a hint function, and the hint isn't very good. We don't
+        // have empirical benchmarks to pick anything more intelligently
+        5
```
tarcieri, line 1520: Chosen by fair dice roll? 😉

mmaker: It's at least not too little (2) to be useful and not too much (10) for memory. Is this blocking? Happy to open a separate issue for it.

rozbb: Just a [joke](https://xkcd.com/221/) :)

rozbb:
Alright, committed aforementioned test

tarcieri:
This should be fine aside from the merge conflicts, but I would expect the existing Strauss/Pippenger variable-time multiscalar multiplication implementations in `curve25519-dalek` would outperform using the generic one in `group`  (which is just Strauss I believe) so I'm not sure what the particular use case for this is other than being able to write generic code (in which case perhaps `group` should support APIs for variable-time multiscalar multiplications)

PR #903, opened 2026-05-03
ed25519-dalek: bump `ed25519` crate dependency to v3.0.0

tarcieri:
This has two test failures:

    serialisation::serialize_deserialize_signature_json
    serialisation::serialize_signature_size

These are due to a last minute change to use `serdect` for implementing `serde` support: RustCrypto/signatures#1324.

The signature size regression from `64` to `72` is a deliberate tradeoff we've accepted in `serdect`. First note that this test is for now the unmaintained `bincode` crate.

The core problem is that `serde` does not actually provide fixed-size arrays as a type within its data model (serde-rs/serde#1937) and you're instead left faking it using `SerializeTuple` and serializing it a byte-at-a-time as a tuple. While this gives optimal-sized results on `bincode`, it gives less-than-optimal results on e.g. `rmp-serde` where the bytes each end up tagged with a type prefix.

`serde` does provide portable APIs for using optimal format-specific byte encodings, but they carry an additional length prefix, which is unnecessary and suboptimal for something fixed-width like an Ed25519 signature, but alas as noted earlier `serde` does not actually have fixed-width arrays in its data model.

The second test failure occurs specifically because `serdect` introspects the format and uses `base16ct` to perform hex serialization for human readable formats. While this is a readability improvement for these formats (and also makes them easier to implement in constant-time, though it's not relevant here), the drawback is it currently requires the `alloc` feature for such formats, which is a regression:

    Error("serializer is human readable, which requires the `alloc` crate feature", line: 0, column: 0)

Note that regardless these are both breaking changes to how `serde` serialization is handled.

Edit: added reference below to `serde_bytes` which was also removed, but could potentially be added back to address some of the issues outlined here. I would be curious to know the use cases where it helps though.

See also:
- RustCrypto/formats#1111
- RustCrypto/formats#1112
- dalek-cryptography/ed25519-dalek#140
- dalek-cryptography/ed25519-dalek#146

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1116,9 +1116,9 @@ dependencies = [
 
 [[package]]
 name = "serdect"
-version = "0.4.2"
+version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9af4a3e75ebd5599b30d4de5768e00b5095d518a79fefc3ecbaf77e665d1ec06"
+checksum = "66cf8fedced2fcf12406bcb34223dffb92eaf34908ede12fed414c82b7f00b3e"
 dependencies = [
  "base16ct",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,6 +47,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
+name = "base16ct"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


tarcieri:
For starters, I would suggest removing tests based on `bincode` since it's unmaintained. I can do that in a separate PR. In the @RustCrypto repos we have migrated to using `postcard` for similar purposes, though `wincode` seems to be a popular ecosystem option too.

I can potentially revert the `ed25519` changes even though we've already shipped v3.0.0 which would be breaking, though we are trying to standardize on `serdect` across-the-board for all @RustCrypto projects.

tarcieri:
Alright, well after [e9df154](https://github.com/dalek-cryptography/curve25519-dalek/pull/903/commits/e9df1548bb65d10d83491e0b7cce8872ccb37658) and a bump to `serdect` v0.4.3 which fixed `no_alloc` serialization, the build is now green 🎉

PR #906, opened 2026-05-23
Prep v5.0-rc.0

rozbb:
Per discussion, I've commented out all the parts of the crate that rely on `ff` and `groups`. Both those deps have been at a release-candidate stage for a while, and we want to do a proper dalek release. So we decided for v5.0 we would just remove the code that relies on those deps, and add the code back in v5.1 once the deps got a release.

I'll be adding the corresponding updates to the remaining crates soon.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -305,7 +305,7 @@ dependencies = [
 
 [[package]]
 name = "curve25519-dalek"
-version = "5.0.0"
+version = "5.0.0-rc.0"
 dependencies = [
  "cfg-if",
  "cpufeatures",
@@ -370,7 +370,7 @@ dependencies = [
 
 [[package]]
 name = "ed25519-dalek"
-version = "3.0.0"
+version = "3.0.0-rc.0"
 dependencies = [
  "blake2",
  "chacha20",
@@ -1466,7 +1466,7 @@ dependencies = [
 
 [[package]]
 name = "x25519-dalek"
-version = "3.0.0"
+version = "3.0.0-rc.0"
 dependencies = [
  "criterion",
  "curve25519-dalek",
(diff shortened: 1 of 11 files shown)


tarcieri:
I'd suggest just removing the code for `ff`/`group` in its own commit/PR. That way you can `git revert` it when you're ready to bring it back.

Also we're doing a `5.0.0-rc` first, right?

rozbb:
ah, right. Fixed both and rebased

tarcieri:
@rozbb if you squash and merge, it won't keep the removal of `ff`/`group` in its own commit, FYI.

It might be cleaner to make that its own PR, and then it can be linked to in the CHANGELOG

rozbb:
Good point, moved to https://github.com/dalek-cryptography/curve25519-dalek/pull/907 . I'll rebase this PR once that one is merged

.github/workflows/curve25519-dalek.yml
```diff
@@ -115,20 +115,6 @@ jobs:
...
-    steps:
-    - uses: actions/checkout@v3
-    - uses: dtolnay/rust-toolchain@stable
-    - env:
-        # This will run AVX2-specific tests and run all of the normal tests
-        # with the AVX2 backend, even if the runner supports AVX512.
-        # This should automatically pick up the simd backend in a x86_64 runner
-        # It should pick AVX2 due to stable toolchain used since AVX512 requires nigthly
-        RUSTFLAGS: '-C target_feature=+avx2'
-      run: cargo test --no-default-features --features alloc,precomputed-tables,zeroize,group-bits --target x86_64-unknown-linux-gnu
```
tarcieri: Just a reminder about this so it's included after rebasing

rozbb:
Alright should be ready to go

tarcieri (approved):

PR #907, opened 2026-05-25
curve: Remove ff/group features for now

rozbb:
Per discussion, I've removed all the parts of `curve25519-dalek` that rely on ff and groups. Both those deps have been at a release-candidate stage for a while, and we want to do a proper dalek release. So we decided for v5.0 we would just remove the code that relies on those deps, and add the code back in v5.1 once the deps got a release.


.github/workflows/curve25519-dalek.yml
```diff
@@ -115,20 +115,6 @@ jobs:
...
-    steps:
-    - uses: actions/checkout@v3
-    - uses: dtolnay/rust-toolchain@stable
-    - env:
-        # This will run AVX2-specific tests and run all of the normal tests
-        # with the AVX2 backend, even if the runner supports AVX512.
-        # This should automatically pick up the simd backend in a x86_64 runner
-        # It should pick AVX2 due to stable toolchain used since AVX512 requires nigthly
-        RUSTFLAGS: '-C target_feature=+avx2'
-      run: cargo test --no-default-features --features alloc,precomputed-tables,zeroize,group-bits --target x86_64-unknown-linux-gnu
```
tarcieri, line 130: Hmm, is this broken or something?

tarcieri (approved):
Looks fine aside from the CI config change which otherwise seems irrelevant?

rozbb:
Good catch, I didn't realize I could just remoev the `group-bits` feature but keep the rest of the test.

tarcieri:
Aah cool. Sidebar: we might consider not adding that feature back. `bitvec` hasn't been receiving regular maintenance and I've recently discovered it isn't sound under either SB or TB.

Otherwise looks good now! I already hit approve.

Issue #912, opened 2026-06-21
Proc macro tests fail on machine that supports AVX2


nazar-pc:
I have the following in `~/.cargo/config.toml`:
```toml
[target.x86_64-unknown-linux-gnu]
rustflags = [
    "-C", "target-cpu=znver4",
    "-C", "link-arg=-fuse-ld=mold"
]
```

With that two tests fail:
```
---- inner_spec_avx2::test_specialized stdout ----

thread 'inner_spec_avx2::test_specialized' (3937277) panicked at curve25519-dalek-derive/tests/tests.rs:109:9:
assertion failed: !IS_AVX2
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- inner_spec_avx2::tests::test_specialized_inner stdout ----

thread 'inner_spec_avx2::tests::test_specialized_inner' (3937278) panicked at curve25519-dalek-derive/tests/tests.rs:116:13:
assertion failed: !super::IS_AVX2


failures:
    inner_spec_avx2::test_specialized
    inner_spec_avx2::tests::test_specialized_inner
```

PR #913, opened 2026-06-21
Support AVX512 on stable Rust

### LINKED ISSUE #758: AVX-512 support on `stable`

nazar-pc:
I took https://github.com/dalek-cryptography/curve25519-dalek/pull/866 as is, rebased it on `main` and fixed it up in the last two commits. The last two commits are the only changes compared to https://github.com/dalek-cryptography/curve25519-dalek/pull/866.

https://github.com/RustCrypto/actions/blob/master/intel-sde-install/action.yaml turned out to be useless due to `813591` hardcoded in it, so I followed example from https://github.com/RustCrypto/stream-ciphers/blob/2aa613ea7007e4756a31c2951dfd1023266ec960/.github/workflows/chacha20.yml#L80-L118

I have not changed anything that wasn't required to get things running, but I had to fix a few bugs and can confirm it now passes existing tests on my Zen 4 CPU with `-C target-cpu=znver4`.

I did hit https://github.com/dalek-cryptography/curve25519-dalek/issues/912 in the process though, but it doesn't seem to be tested in CI, so :shrug:.

I'd love to see runtime detection for AVX512 in addition to compile-time check, however it is beyond simply making things work on stable.

Fixes https://github.com/dalek-cryptography/curve25519-dalek/issues/758, closes https://github.com/dalek-cryptography/curve25519-dalek/pull/866

### DIFF
--- a/.github/workflows/curve25519-dalek.yml
+++ b/.github/workflows/curve25519-dalek.yml
@@ -65,7 +65,7 @@ jobs:
         run: cargo build -p ${{ matrix.crate }} --target thumbv7em-none-eabi --release --no-default-features
       - name: no_std fiat / cargo hack ${{ matrix.crate }}
         env:
-          RUSTFLAGS: '--cfg curve25519_dalek_backend="fiat"'      
+          RUSTFLAGS: '--cfg curve25519_dalek_backend="fiat"'
         run: cargo hack build -p ${{ matrix.crate }} --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std,getrandom
 
   test-serial:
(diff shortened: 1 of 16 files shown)


tarcieri:
FYI, I *just* fixed https://github.com/RustCrypto/actions/tree/master/intel-sde-install

All it needs is:

```yaml
- uses: RustCrypto/actions/intel-sde-install@master
```

tarcieri:
@rozbb IMO this is worth getting into the v5.0.0 release

nazar-pc:
Done, all tests pass now. Let me know if further changes are needed, I should be able to make them. Or feel free to push on top, you should have permissions to do so.

.github/workflows/curve25519-dalek.yml
```diff
@@ -19,6 +19,9 @@ defaults:
 env:
   CARGO_TERM_COLOR: always
   RUSTFLAGS: '-D warnings'
+  # NOTE: The mirror number changes with each version so keep these in sync
+  SDE_FULL_VERSION_MIRROR: '915934'
+  SDE_FULL_VERSION: '10.8.0-2026-03-15'
```
tarcieri: These aren't needed any more

tarcieri (approved):

tarcieri:
Hmm strange, saw this build failure on nightly after merging: https://github.com/dalek-cryptography/curve25519-dalek/actions/runs/28380522785/job/84081560067

cc @nazar-pc @rozbb 

```
error[E0433]: cannot find `spec_avx512ifma_avx512vl` in `precomputed_straus`
   --> curve25519-dalek/src/backend.rs:105:49
    |
105 |         vector::scalar_mul::precomputed_straus::spec_avx512ifma_avx512vl::VartimePrecomputedStraus,
    |                                                 ^^^^^^^^^^^^^^^^^^^^^^^^ could not find `spec_avx512ifma_avx512vl` in `precomputed_straus`
    |
note: found an item that was configured out
   --> curve25519-dalek/src/backend/vector/scalar_mul/precomputed_straus.rs:18:9
    |
 14 | / #[curve25519_dalek_derive::unsafe_target_feature_specialize(
 15 | |     "avx2",
 16 | |     conditional("avx512ifma,avx512vl", curve25519_dalek_backend = "avx512")
    | |_____________________________________- the item is gated behind the `avx512ifma` feature
 17 |   )]
 18 |   pub mod spec {
    |           ^^^^
```

PR #915, opened 2026-06-30
doc(lizard): formal collision bound proof sketch and Jacobi pair s0·s1=-1 identity

nzengi:
## Summary

Two documentation improvements to `lizard_ristretto.rs`. No functional changes.

### 1. Formal collision bound proof sketch

Replaces the informal comment _"the likelihood that n_found == 2 is something like 2^-122"_ with a precise argument showing the tight bound is **7 · 2^-125 < 2^-122** in the random oracle model.

The core argument:

- There are at most **7** other positive preimage field elements beside `fe_uuid`.
- **Claim:** any `fe[j]` (j ≥ 1) that passes TagCheck must have `payload(fe[j]) ≠ uuid`. _Proof:_ if they match, the tag check forces `fe[j] == fe_uuid` as field elements — contradiction.
- Under the ROM, SHA256 of that distinct payload is independent uniform, constrained to **125 bits** (63 from bytes[0..8] with LSB always 0, plus 62 from bytes[24..32] with top two bits always 0). Each succeeds with probability 2^-125.
- Union bound over 7 preimages: **7 · 2^-125 < 2^-122**.

Also explains why the 8 negative preimage field elements (LSB = 1) can never pass TagCheck, the formal justification behind the optimization in PR #882.

### 2. Algebraic identity s0 · s1 = -1 in GF(p)

In `to_jacobi_quartic_ristretto`, the S-coordinates of the first Jacobi quartic pair satisfy:

```
s0 =  γ·Y²·(Z-Y)·X,  s1 = -γ·Y²·(Z+Y)·X,  γ² = 1/(Y⁴·X²·(Z²-Y²))
⟹  s0·s1 = -γ²·Y⁴·X²·(Z²-Y²) = -1  (exactly, in GF(p))
```

So JC₀ and JC₁ are algebraically dependent (S₁ = -1/S₀). The comment explains why this does not affect the collision bound: the tag check operates on serialized bytes, not S-coordinates, and GF(p) inversion scrambles the byte representation (validated empirically: Welch t < 2 over 2.5M samples).

## Relation to PR #882

This PR provides the formal justification for the key observation in #882 that negative preimages can be skipped. The proof sketch here explains exactly _why_ they can never contribute to the collision count.

## Files changed

`curve25519-dalek/src/lizard/lizard_ristretto.rs`: two comment blocks added, one existing comment replaced (31 lines added, 1 removed). No functional changes.

### DIFF
--- a/curve25519-dalek/src/lizard/lizard_ristretto.rs
+++ b/curve25519-dalek/src/lizard/lizard_ristretto.rs
@@ -66,7 +66,28 @@ impl RistrettoPoint {
             }
             n_found += ok.unwrap_u8();
         }
-        // Per the README, the likelihood that n_found == 2 is something like 2^-122
+        // The probability that n_found >= 2 is at most 7 * 2^-125 < 2^-122.
+        //
+        // Proof (ROM): there are at most 7 other positive preimage field elements
+        // fe[1]..fe[7] besides the encoding preimage fe[0] = fe_uuid.
+        //
+        // Claim: if fe[j] (j >= 1) passes TagCheck, then payload(fe[j]) != uuid.
+        // Proof of claim: if payload(fe[j]) == uuid, the tag check would require
+        // outer_bytes(fe[j]) == mask(SHA256(uuid)) == outer_bytes(fe_uuid), forcing
+        // fe[j] == fe_uuid as field elements — contradiction.
+        //
+        // Therefore any fe[j] that could pass TagCheck queries SHA256 at a fresh
+        // input distinct from uuid. Under the random oracle model that output is
+        // independent uniform, and TagCheck constrains 63 bits from bytes[0..8]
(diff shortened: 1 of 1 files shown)


rozbb:
I'm not sure I understand

For change 1, how does this differ from the collision probability argument presented in the README?

For change 2, what precisely is the thing being claimed? Why would one think that the relation between s0 and s1 would affect the collision probability argument?

nzengi:
Fair points, thanks. On (1): you're right, this is essentially a restatement of the derivation already in src/lizard/README.md (Westerbaan's argument), just moved next to the code not a new result. On (2): "formal proof sketch" oversold it; the decorrelation claim rests on an empirical Welch's t-test, not a proof, and I'd need to double check a sign discrepancy between the README's s0/s1 formulas and the code before standing behind the identity as stated. Not worth carrying forward as-is closing this out. Happy to revisit if there's actual interest in documenting the negative-preimage exclusion argument from #882 more formally.

PR #919, opened 2026-07-25
refactor(curve25519-dalek-derive): bump `syn` to v3

AudaciousAxiom:
Following the [release of a new major version of `syn`](https://github.com/dtolnay/syn/releases/tag/3.0.0), this bumps the version used by `curve25519-dalek-derive`, with the goal of eventually removing the duplicated dependency in binaries (and therefore reducing compile times).

`syn` v3 advertises an [MSRV of 1.71](https://crates.io/crates/syn/3.0.0/code/Cargo.toml#L14), so this shouldn't affect the MSRV of this crate.

I've gone through the changelog and did the necessary refactors; I didn't find any other relevant breaking changes (but there are quite a few of them so I could have missed some). I've run `cargo test` but didn't do any other kind of testing.


curve25519-dalek-derive/src/lib.rs
```diff
@@ -311,7 +312,7 @@ fn process_function(
         })
         .unzip::<_, _, Vec<_>, Vec<_>>();
 
-    if function.sig.unsafety.is_some() {
+    if matches!(function.sig.safety, syn::Safety::Unsafe(_)) {
```
AudaciousAxiom, line 315: Function signatures now have [a 3-variant enum for encoding their safety](https://docs.rs/syn/latest/syn/enum.Safety.html), which a default variant for when there is no `safe` or `unsafe` keyword.

curve25519-dalek-derive/src/lib.rs
```diff
@@ -338,7 +339,13 @@ fn process_function(
         match arg {
             syn::FnArg::Receiver(receiver) => {
                 unsupported_if_some!(receiver.attrs.first());
-                unsupported_if_some!(receiver.colon_token);
+                unsupported_if_some!(
+                    if let syn::ReceiverKind::Typed(colon_token, _) = receiver.kind {
```
AudaciousAxiom: Receivers now [have a kind field](https://docs.rs/syn/latest/syn/enum.ReceiverKind.html#variant.Typed). This should preserve the existing span for the potential error as the colon token is still returned (but I haven't tested this). It may also be possible to return the whole receiver if preserving the span is not important.

Cargo.lock
AudaciousAxiom, line 782: A new version of `prettyplease` that depends on syn@v3 has been released. I think this is better updated in another PR, but I can also update the lockfile in this PR if you prefer.

AudaciousAxiom, line 1014: A new version of `serde_derive` that depends on syn@v3 has also been released.

AudaciousAxiom, line 1176: A new version of `thiserror-impl` that depends on syn@v3 has also been released.

rozbb:
Thank you! Could i have permission to push to your branch please? I had a tiny unrelated change I wanted to add while we're already fixing `derive` code

AudaciousAxiom:
> Could i have permission to push to your branch please?

@rozbb I'm not sure what you want me to do: "Allow edits by maintainers" is already enabled (feel free to use the permission). I can easily rebase on top of whichever commit you want to add however.

rozbb:
Ah, my bad, I was pushing to the wrong branch. Thank you for the PR!

rozbb (approved):

tarcieri (approved):

tarcieri:
Hmm, swear it was green but now some tests are failing?

rozbb:
Spurious. Nightly is failing bc of `cpufeatures`. Need to bump to 0.3.1 when it drops

tarcieri:
I'd say go ahead and merge it then

rozbb:
Checked that this doesn't modify any codegen. Pasting the script for posterity. Requires [cargo-expand](https://github.com/dtolnay/cargo-expand).


```bash
#!/bin/bash

set -euo pipefail

CMD="cargo expand -q --target x86_64-unknown-linux-gnu -p curve25519-dalek --lib"

echo -n "Serial expanded syntax hash:"
RUSTFLAGS='--cfg curve25519_dalek_backend="serial"' \
        $CMD | sha256sum

echo ""

echo -n "AVX2 expanded syntax hash:"
RUSTFLAGS='--cfg curve25519_dalek_backend="simd" -C target_feature=+avx2' \
        $CMD | sha256sum

echo ""

echo -n "AVX512 expanded syntax hash:"
RUSTFLAGS='--cfg curve25519_dalek_backend="avx512" -C target_feature=+avx512ifma,+avx512vl' \
        $CMD | sha256sum
```

nazar-pc:
I'd really appreciate a new release with this change

